PR #1, opened 2016-04-30
Two basic examples for elif checker, and new pylintrc configuration file added

david-yz-liu:


pyta.py
```diff
@@ -30,6 +30,8 @@ def check(module_name):
     linter = lint.PyLinter(reporter=reporter)
     linter.load_default_plugins()
     linter.load_plugin_modules(['checkers'])
+    linter.read_config_file()
+    linter.load_config_file()
```
nigef: Here is a comment. :)

PR #24, opened 2016-05-11
String Constant Checker Examples

Jasminewwwu:


examples/pylint/W1401_anomalous_backslash_in_string.py
```diff
@@ -0,0 +1,4 @@
+"""pylint: anomalous backslash in string
+"""
+import re
```
david-yz-liu: Don't import `re` here. Just show a string with the anomalous backslash.

david-yz-liu:
Thanks, @Jasminewwwu!

PR #27, opened 2016-05-15
IO functions checker

Jasminewwwu:


IO_Function_checker.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @check_messages("I/O-function-not-allowed")
+    def visit_call(self, node):
+        if isinstance(node.func, astroid.Name):
+            name = node.func.name
+            # ignore the name if it's not a builtin (i.e. not defined in the
+            # locals nor globals scope)
+            if not (name in node.frame() or name in node.root()):
+                # if name in self.config.io_functions:
+                if name == "print" or name == "input" or name == "open":
```
david-yz-liu: This check can be made a little cleaner by putting the forbidden names in a global variable. Try looking at what Hayley did if you aren't sure what I mean.

IO_Function_checker.py
```diff
@@ -0,0 +1,43 @@
...
+                      'That doesn\'t mean you can not use it !'),
+           }
+
+    @check_messages("I/O-function-not-allowed")
+    def visit_call(self, node):
+        if isinstance(node.func, astroid.Name):
+            name = node.func.name
+            # ignore the name if it's not a builtin (i.e. not defined in the
+            # locals nor globals scope)
+            if not (name in node.frame() or name in node.root()):
```
david-yz-liu: This is an **excellent** check. @hayleyZHL, can you add the same to yours?

david-yz-liu:
@Jasminewwwu well done! One general comment is to use `IO` instead of `I/O` in your names for the checkers.

IO_Function_checker.py
hayleyZHL: Yes, I'm not very sure which part should I add? because we used almost the same approach.

david-yz-liu: This exact check. (Comments always appear beneath the line they are referring to.)

pyta.py
```diff
@@ -30,6 +30,7 @@ def check(module_name):
     linter = lint.PyLinter(reporter=reporter)
     linter.load_default_plugins()
     linter.load_plugin_modules(['checkers'])
+    linter.load_plugin_modules(['IO_Function_checker'])
```
david-yz-liu: I think you can add this module to the list in the previous line

IO_Function_checker.py
```diff
@@ -0,0 +1,51 @@
...
+                # if name in self.config.io_functions:
+                if name in self.config.functions_not_allowed:
+                    args = "{} on line {}".format(name, node.lineno)
+                    # add the message
+                    self.add_message('IO-function-not-allowed', node=node, args=args)
+
+
+def register(linter):
+    """
```
david-yz-liu: Remove this docstring

david-yz-liu:
@Jasminewwwu this pull request contains too many new features. Remember that every new task you work on should be on a separate branch. Here, you have included two new checkers, when only one is asked for.

You should create two new branches, each of which has your changes for exactly one of the new checkers.

PR #29, opened 2016-05-16
Disable E1303, W1306, W1307

Jasminewwwu:
disable error:
W1306: missing-format-attribute
W1307: invalid-format-index
E1303: format-needs-mapping


david-yz-liu:
Thanks, @Jasminewwwu!

PR #31, opened 2016-05-16
Global variables checker

shweta97:
Created and tested a custom checker that forbids the use of the global and nonlocal statements

### DIFF
--- a/examples/global_variables_test.py
+++ b/examples/global_variables_test.py
@@ -2,7 +2,7 @@
 """
 
 
-def method():
+def function():
     """ A test for the global variables checker
     @rtype: None
     """
@@ -12,16 +12,16 @@ def method():
     VALUE = 100
 
 VALUE = 0
-method()
+function()
 print(VALUE)
 
 
-def method1():
+def function1():
     """ A test for the global variables checker
     @rtype: None
     """
 
-    def method2():
+    def function2():
         """ A test for the global variables checker
          @rtype: None
         """
@@ -31,7 +31,7 @@ def method1():
 
     # Set local.
     value = 10
-    method2()
+    function2()
 
     # Local variable reflects nonlocal change.
     print(value)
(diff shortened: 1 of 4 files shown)


examples/global_variables_test.py
```diff
@@ -0,0 +1,37 @@
+"""Test file for global variables checker
+"""
+
+
+def method():
```
david-yz-liu: Use the term "function", not "method". Methods are functions which are defined in a class, bound to an object.

david-yz-liu:
@shweta97 this is an excellent start, and I will merge it in. However, we also want to check for all top-level variables which are not constants. This is harder to do, and is what I want you to work on next.

shweta97:
I've been trying to do a check for top-level variables with no luck. I will continue to work on it.

pyta.py
```diff
@@ -30,6 +30,9 @@ def check(module_name):
     linter = lint.PyLinter(reporter=reporter)
     linter.load_default_plugins()
     linter.load_plugin_modules(['checkers'])
+    linter.load_plugin_modules(['global_variables_checker'])
```
david-yz-liu: I think you can add this module to the list in the previous line

pyta.py
```diff
@@ -30,6 +30,9 @@ def check(module_name):
...
+    linter.read_config_file()
```
david-yz-liu: Make sure to do a pull from master.

shweta97: I've tried that but there are some conflicts arising and I'm not sure where I can see what the conflicts are.

shweta97:
I'm still trying to differentiate the top-level constants from the variables using a bit of regex. But i've committed the rest of it what I have done this week.

david-yz-liu:
@shweta97 You have made the same mistake as @Jasminewwwu: adding too many changes to the same pull request.

You should now create two new branches, one for each new checker you added, and make two different pull requests.

PR #36, opened 2016-05-23
Comment extract

WChrisK:


WChrisK:
Note: Add docs

WChrisK:
Do you want me to put this into its own class? Or are top level functions okay? Are the names okay?

david-yz-liu:
@WChrisK I appreciate the work you did; I didn't realize the `inspect` module is as powerful as it is.

However, it seems that extracting other information provided by `inspect` really duplicates what astroid already provides. I looked into the documentation of the pylint source code; it appears at https://github.com/PyCQA/pylint/blob/master/pylint/checkers/base.py#L1398 to access a `doc` attribute of a node, which after some testing seems to be more or less exactly what we want. Briefly, it looks like at least both `Module` and `FunctionDef` node classes have this attribute.

So don't bother with creating new classes or anything; any documentation checker that we actually want to write can and should be based on accessing this attribute, which is good because that way the checkers can be of the same form as the other custom checkers our team is writing.

I recommend working on the next part of the issue (parsing the docstring text according to the 108 docstring format).

PR #37, opened 2016-05-23
Issue 4

nigef:
Removed docstrings (placing any points that would vanish into the markdown file), made minor adjustments to some examples.


david-yz-liu:
This is a nice improvement, thanks @nigef!

PR #39, opened 2016-05-23
Always_returning_in_a_loop_checker

Jasminewwwu:


Always_return_checker.py
```diff
@@ -0,0 +1,82 @@
...
+        first_branch = node.body[0]
+        # If the node is just a single statement, check whether it's a return
+        # statement.
+        if length == 1 and first_branch.is_statement:
+            if isinstance(first_branch, astroid.Return):
+                args = "{}".format(node.lineno)
+                self.add_message("always_returning_in_a_loop", node=node, args=args)
+        # If it's a sequence of statements, check whether one of them is a
+        # return statement
+        if length > 1 and all([child.is_statement for child in node.body]):
```
david-yz-liu: This case can be merged with the previous case (a single statement and a sequence of statements are handled the same way).

Always_return_checker.py
```diff
@@ -0,0 +1,82 @@
...
+            if any([isinstance(child, astroid.Return) for child in node.body]):
+                args = "{}".format(node.lineno)
+                self.add_message("always_returning_in_a_loop", node=node, args=args)
+        # If it's an if statement, check each branch separately, and only
+        # return True if every branch has an explicit return.
+        if length == 1 and isinstance(first_branch, astroid.If):
+            if any([isinstance(child, astroid.Return) for child in first_branch.body]):
```
david-yz-liu: I'm not convinced this works if each branch has *more* than a single statement, e.g.

`
if x > 5:
    print(x)
    return 5
else:
    y = x + 5
    return 10
```

**Comment by david-yz-liu** [comment]
@Jasminewwwu You have made good progress; however, your approach is limited for a few different reasons.

You need to take what you've done, and put the bulk of the work in a **recursive helper function**, whose different cases correspond to the different types of nodes. By making it recursive, you'll be able to do things like handle the case where an `if` branch has multiple statements, simply by making a recursive call on the body as a whole.

**Comment by Jasminewwwu** [comment]
When I run use pyta to check the example
The result is correct which is 
```
E9996: (always_returning_in_a_loop)
    Always returning in loop on line 18
```

File: Always_return_checker.py [hunk 3]
```diff
@@ -0,0 +1,80 @@
...
+    @check_messages("always_returning_in_a_loop")
+    def helper_fuc(self, node):
+        # Check if the node is a return statement
+        if isinstance(node, astroid.Return):
+            args = "{}".format(node.lineno)
+            self.add_message("always_returning_in_a_loop", node=node, args=args)
+        # check if the current branch is astroid.If
+        if isinstance(node, astroid.If):
+            flag = True
+            if not any([isinstance(child, astroid.Return) for child in (
```
david-yz-liu: Rather than checking if the child is an `astroid.Return`, you should make *additional* recursive calls on the children. This will allow you to handle, for example, a nested `if`.

PR #40, opened 2016-05-24
Global variables checker

shweta97:
Reverted the branch back to the state in the old pull request and separated the work between different branches.


shweta97:
I'm sorry for my mistake. I was under the assumption that all my tasks under a specific issue were supposed to be on the same pull request. But now I understand it better. Sorry once again. This will not repeat.

david-yz-liu:
Thanks, @shweta97!

PR #41, opened 2016-05-24
Moving custom checker into subdirectory

hayleyZHL:
separated from dynamic-execution-checker branch.


examples/dynamic_execution_checker_example.py
```diff
@@ -2,6 +2,7 @@
 
 
 def dynamic_execution(name):
+    """Append in front of string 'name'"""
```
david-yz-liu: This change doesn't belong here.

david-yz-liu:
@hayleyZHL This is a much better pull request. Why don't you do a merge from master (resolving any conflicts which appear - ask for help on Slack), and move all of the new checkers into the subfolder as well? After that, I will merge this branch in and give you a new task.

david-yz-liu:
@hayleyZHL make sure to closed old pull requests if you replace them with a new one.

PR #43, opened 2016-05-24
Top level global variables

shweta97:
Modified the global variables checker so that it throws an error when top-level variables are declared and used.

### DIFF
--- a/global_variables_checker.py
+++ b/global_variables_checker.py
@@ -11,8 +11,8 @@ class GlobalVariablesChecker(BaseChecker):
     __implements__ = IAstroidChecker
 
     name = 'global_variables'
-    msgs = {'E9997': ('Global variables should not be used in CSC108/CSC148 - %s', 'forbidden-global-variables',
-                      '')}
+    msgs = {'E9997': ('Global variables should not be used in CSC108/CSC148 - '
+                      '%s', 'forbidden-global-variables', '')}
 
     options = ()
     # this is important so that your checker is executed before others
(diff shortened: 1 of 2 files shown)


david-yz-liu:
@shweta97 you should pull from master again, as I have now merged in your previous branch.

You can look into the regular expression defined in pylint itself (https://github.com/PyCQA/pylint/blob/master/pylint/checkers/base.py#L58) for detecting a naming convention for constant expressions. This is a good next step to work on.

shweta97:
@david-yz-liu I have programmed the checker to raise an error when a top-level variable is used like `x = 5` but not for just `x`. If the error should be raised even for just `x`, then I would have to add another method. Please let me know your thoughts on this.

david-yz-liu:
@shweta97 sorry for the delay in replying - only checking top-level assignment statements is okay.

global_variables_checker.py
```diff
@@ -24,6 +26,16 @@ def visit_nonlocal(self, node):
         args = "you used the keyword 'nonlocal' on line {}".format(node.lineno)
         self.add_message('forbidden-global-variables', node=node, args=args)
 
+    def visit_assign(self, node):
+        if isinstance(node.frame(), astroid.scoped_nodes.Module):
+            regex = str(node.targets[0])
+            s = re.findall('\((.*?)\)', regex)[0]
+            a = re.match('(([A-Z_][A-Z0-9_]*)|(__.*__))$', s)
```
david-yz-liu: I'd like you see you actually import and use the one defined in pylint directly, rather than making a copy.

global_variables_checker.py
```diff
@@ -24,6 +26,16 @@ def visit_nonlocal(self, node):
...
+            # Raise an error only if it's not a constant
+            if a is None:
+                args = "you declared global variables on line {}".format(node.lineno)
```
david-yz-liu: Can you please add the *name* of the variable to this error message?

global_variables_checker.py
shweta97: I have tried importing base.py from pylint.checkers where it is originally defined but I'm not able to use the constant.

david-yz-liu: You likely weren't importing the correct thing. Try doing `from pylint.checkers.base import CONST_NAME_RGX`.

shweta97: Yes I had imported the wrong thing. It works now.

david-yz-liu:
Thanks, @shweta97!

PR #44, opened 2016-05-25
Merged from master and put Custom Checkers into the Subdirectory `checkers`.

hayleyZHL:


examples/dynamic_execution_example.py
```diff
@@ -2,6 +2,7 @@
```
nigef: we dont use the docstrings at the top of examples anymore

david-yz-liu:
@hayleyZHL thanks - I updated the branch according to @nigef's comment and merged this in.

PR #46, opened 2016-05-29
remove the docstring of E9998 example

Jasminewwwu:


david-yz-liu:
@Jasminewwwu Nigel was actually talking about the module docstring, not the function docstring.

PR #48, opened 2016-05-29
Robustness of module loading

hayleyZHL:
1. If the extension `.py` is added, remove it but load the correct
module without error.
2. when a module is not found, and report an error message to the user.
3. If the module name is a recognizable type by python, but not the type string, report an error message
to the user.


pyta.py
```diff
@@ -26,7 +26,19 @@ def check(module_name):
     The name of the module should be passed in as a string,
     without a file extension (.py).
     """
+
+    # Check if `module_name` is not the type str, raise error.
+    if not isinstance(module_name, str):
+        raise NameError("The Module '{}' has an invalid name. Module "
```
david-yz-liu: This shouldn't raise an error either.

pyta.py
```diff
@@ -36,7 +48,13 @@ def check(module_name):
                                 'checkers/IO_Function_checker'])
     linter.read_config_file()
     linter.load_config_file()
-    linter.check([spec.origin])
+
+    # When module is not found, raise exception.
+    try:
+        linter.check([spec.origin])
+    except AttributeError:
+        print ("The Module '{}' could not be found in the given "
```
nigef: you can bind this error to a variable,

and use `sys.exc_info()` from the `sys` package to get the object that caused the error, and the line number stored in the `tb_lineno` property of the traceback object (the third element ing the tuple returned by `sys.exc_info()`)

david-yz-liu: I'm not sure what @nigef is suggesting the message should actually be.

@hayleyZHL the phrase "given location" is unclear. For now you can simply end the sentence after `found`.

nigef: Purpose is to mention a way to add more information to the message, i.e. line number, etc..

hayleyZHL:
@david-yz-liu  Thanks and they are now fixed.

pyta.py
hayleyZHL: @nigef  Since this is the `module not found` issue, there is no line number. In this case, the file they want to check does not exist.. so it's not their code problem..

pyta.py
```diff
@@ -26,7 +26,18 @@ def check(module_name):
     The name of the module should be passed in as a string,
     without a file extension (.py).
     """
+
+    # Check if `module_name` is not the type str, raise error.
+    if not isinstance(module_name, str):
+        print ("The Module '{}' has an invalid name. Module "
+                        "name must be the type str.".format(module_name))
```
david-yz-liu: You should `return` after this line (the rest of the code shouldn't execute).

pyta.py
```diff
@@ -36,7 +47,12 @@ def check(module_name):
                                 'checkers/IO_Function_checker'])
     linter.read_config_file()
     linter.load_config_file()
-    linter.check([spec.origin])
+
+    # When module is not found, raise exception.
+    try:
+        linter.check([spec.origin])
+    except AttributeError:
+        print ("The Module '{}' could not be found.".format(module_name))
```
david-yz-liu: Similarly, return after this line (nothing else should be printed).

PR #49, opened 2016-05-30
Custom checker 

shweta97:
Examples for messages in custom checker.


examples/pylint/C0301_line_too_long.py
```diff
@@ -0,0 +1,5 @@
+"""pylint: Line too long
```
nigef: we are not including module docstrings in the examples, based on our new format

examples/pylint/C0321_multiple_statements.py
```diff
@@ -0,0 +1,11 @@
+"""pylint: Multiple statements
+
+"""
+
+
+def pos(temp):
+    """Print 'positive' if <temp> is greater than 0. ELse print 'negative'.
+     """
```
nigef: i think theres an extra space here

examples/pylint/C0321_multiple_statements.py
```diff
@@ -0,0 +1,11 @@
...
+    if temp > 0: print("positive")
```
nigef: i think you meant to consistently use single quotes

examples/pylint/C0326_bad_whitespace.py
```diff
@@ -0,0 +1,12 @@
+""" pylint: Bad whitespace
+
+"""
+
+
+def func( temp):
+    """Prints if a function is positive or negative
```
nigef: "Prints if a function"?

.pylintrc
```diff
@@ -4,6 +4,11 @@
 # there seems to be an off by one issue in how max-nested-blocks is used.
 max-nested-blocks = 2
 
+[FORMAT]
```
nigef: i don't think this format part should be in the commit, right?

shweta97: What do you mean exactly? I just followed David's initial pull request to change default values.

nigef: i wasn't sure if it was added by mistake. not sure what David's initial pull request is.

examples/pylint/C0321_multiple_statements.py
```diff
@@ -0,0 +1,6 @@
+def pos(temp):
+    """Print "positive" if <temp> is greater than 0. ELse print "negative".
```
david-yz-liu: In general, please don't use `print` in examples. Just return something instead.

examples/pylint/C0321_multiple_statements.py
```diff
@@ -0,0 +1,8 @@
+def pos(temp):
+    """Return "positive" if <temp> is greater than 0. ELse return "negative".
```
david-yz-liu: There is a typo with `ELse`

examples/pylint/C0321_multiple_statements.py
```diff
@@ -0,0 +1,8 @@
...
+    @type temp: int
+    @rtype: str
+    """
+    if temp > 0: return "positive"
```
david-yz-liu: Please put this on a new line.

Also, please use single-quotes rather than double-quotes throughout.

examples/pylint/C0326_bad_whitespace.py
```diff
@@ -0,0 +1,9 @@
+def func( temp):
+    """Return "positive" if <temp> is greater than 0. ELse return "negative".
```
david-yz-liu: This docstring doesn't match what the function actually does.

examples/pylint/C0301_line_too_long.py
```diff
@@ -0,0 +1 @@
+TEMP = "I am in the file C0301_line_too_long.py. I raise the error for a very long line."
```
nigef: instead of,
I raise the error for a very long line.

maybe say like,
This example raises an error for exceeding the acceptable line length.

i'm not sure if we should use "I"

david-yz-liu: I agree with @nigef. Also don't use the term "raise" for making pylint emit errors, since this is different from the program raising an exception.

david-yz-liu:
@shweta97 a reminder that there are outstanding comments on this pull request

examples/pylint/C0321_multiple_statements.py
shweta97: This is an example for the message multiple statements which is only emitted when I do this.

examples/pylint/C0321_multiple_statements.py
```diff
@@ -0,0 +1,8 @@
+def pos(temp):
+    '''Return 'positive' if <temp> is greater than 0. Else return 'negative'.
+    @type temp: int
+    @rtype: str
+    '''
+    if temp > 0: return 'positive'
```
david-yz-liu: Thanks for the clarification, @shweta97!

PR #51, opened 2016-05-31
Fixes#32 always returning checker

Jasminewwwu:
Now the error message will show the for/while node lineno.

However, do I need to add the specific type of loop in the error message?


Jasminewwwu:
I did not see the comment until now... I will change the code.

david-yz-liu:
Saying the loop type would be good. Also, I made a previous comment about putting more recursive calls in your previous pull request. The problem with closing pull requests is you don't see the comments any more, so you and @hayleyZHL should stop doing that. Look through your old one to find the comments.

Jasminewwwu:
Won't happen again. Actually, I added the recursive part as you suggested in that pull request. But I forgot to mention that.  It is from` line 32` - `line 34` in `always_returning_checker.py`

checkers/always_returning_checker.py
```diff
@@ -0,0 +1,89 @@
...
+            flag = True
+            if not any([isinstance(child, astroid.Return) for child in (
+                    node.body + node.orelse)]):
+                flag = False
+            # set first_branch
+            if node.orelse and isinstance(node.orelse[0], astroid.If):
+                first_branch = node.orelse[0]
+            elif node.orelse:
+                # Return False if there is no return in else branch
+                if not any([isinstance(child, astroid.Return) for child in (
```
david-yz-liu: @Jasminewwwu You should be making a recurisve call here, as well (and in the other places similar to this one).

examples/always_returning_example.py
```diff
@@ -0,0 +1,21 @@
+def loop():
+    for j in range(0, 1):
```
nigef: does the checker report an error for this loop on line 2? should it?

checkers/always_returning_checker.py
```diff
@@ -0,0 +1,103 @@
...
+                      'always_returning_in_a_loop',
+                      'Used when you always return an object or none in a loop, '
+                      'this may cause the loop only runs once.'),
+           }
+
+    # this is important so that your checker is executed before others
+    priority = -1
+
+    @check_messages("always_returning_in_a_loop")
+    def helper_func(self, node, args):
```
nigef: what should this function be named?

checkers/always_returning_checker.py
```diff
@@ -0,0 +1,103 @@
...
+                else:
+                    first_branch = "end"
+            # check the flag and decide whether there is error
+            if flag:
+                self.add_message("always_returning_in_a_loop", node=node,
+                                 args=args)
+        # If it's a loop, check the body of the loop.
+        if isinstance(node, astroid.For) or isinstance(node, astroid.While):
+            for child in node.body:
+                self.helper_func(child, args=node.lineno)
```
nigef: it would be visually nice if there was some more blank lines used between some code blocks in this function.

checkers/always_returning_checker.py
```diff
@@ -0,0 +1,103 @@
...
+           }
+
+    # this is important so that your checker is executed before others
+    priority = -1
+
+    @check_messages("always_returning_in_a_loop")
+    def helper_func(self, node, args):
+        # Check if the node is a return statement
+        if isinstance(node, astroid.Return):
+            self.add_message("always_returning_in_a_loop", node=node, args=args)
```
nigef: the node and args assignment is not necessary in the arguments

checkers/always_returning_checker.py
```diff
@@ -0,0 +1,103 @@
...
+                                                      astroid.If):
+                    # deal with the condition of nested if
+                    for child in first_branch.orelse.body:
+                        if isinstance(child, astroid.If):
+                            return self.helper_func(child, args)
+                    first_branch = first_branch.orelse[0]
+                # Return False if there is no return in else branch
+                elif first_branch.orelse:
+                    for child in first_branch.orelse:
+                        if isinstance(child, astroid.If):
```
nigef: i see repeated code on line 33, 59, 70, 76, which is a bit of a red flag regarding the conciseness of this function.
please consider refactoring.

checkers/always_returning_checker.py
```diff
@@ -0,0 +1,103 @@
...
+        # If it's a loop, check the body of the loop.
+        if isinstance(node, astroid.For) or isinstance(node, astroid.While):
+            for child in node.body:
+                self.helper_func(child, args=node.lineno)
+
+    def visit_for(self, node):
+        args = "{}".format(node.lineno)
+        return self.helper_func(node, args)
+
+    def visit_while(self, node):
```
nigef: the contents of visit_for and visit_while are identical, so why do we need two of them?

checkers/always_returning_checker.py
david-yz-liu: @nigef how do you know this?

checkers/always_returning_checker.py
david-yz-liu: @nigef I do not require more style changes for this function before mergin this in. You are welcome to work with @Jasminewwwu further for a second pass after this is complete.

examples/always_returning_example.py
david-yz-liu: @Jasminewwwu increase the loop index. The current range does indeed distract from the check.

checkers/always_returning_checker.py
nigef: followup: my mistake. now i understand the overloading requires it.

hayleyZHL:
@david-yz-liu  thanks and they are now fixed.

examples/always_returning_example.py
Jasminewwwu: Emmm, do you think that may be a little bit confusing for students? If they see two messages indicating the same error. 
Currently, the checker is reporting specifically the location of the error(the closest loop).
What do you think?

david-yz-liu: @Jasminewwwu I don't understand what you mean.

Jasminewwwu: for this example:
```
def loop():
    for j in range(0, 1):
        if j < 2:
            j += 1
            for k in range(0, 2):
                if k > 2:
                    return 1
                else:
                    return 1
        elif j > 1:
            return 1
        else:
            j += 1
            return 1
    i = 0
    while i < 10:
        if i > 2:
            return 2
        else:
            print(1)
        return 4
```
Now, the message indicates there is an error in the loop on line 5. I wonder do u want me to change the lineno to line 2 or to line 2 & 5.

david-yz-liu: There should be an error message in both.

But it appears you didn't understand what I asked you to do: change `range(0, 1)` to something like `range(0, 10)`. `range(0, 1)` is an error in itself that we certainly do not want to model, and perhaps we should check for.

Jasminewwwu: Oh!!! Sorry I did not notice that mistake. Maybe we could do a checker for that(Especially beneficial for careless students like me.)

david-yz-liu: Sure, why don't you write up an issue for that?

PR #54, opened 2016-06-05
Fixes#3 check python version

Jasminewwwu:
some warning messages will show up if wrong python version is detected


pyta.py
```diff
@@ -26,6 +27,9 @@ def check(module_name):
     The name of the module should be passed in as a string,
     without a file extension (.py).
     """
+    if sys.version_info<(3,3,0):
```
david-yz-liu: Please put the necessary whitespace

pyta.py
```diff
@@ -26,6 +27,9 @@ def check(module_name):
...
+        sys.stderr.write("You need python 3.3 or later to run this script\n")
```
david-yz-liu: Please use single-quotes; also, only allow Python version 3.4+. And also, please capitalize "Python."

pyta.py
```diff
@@ -14,11 +14,17 @@
...
 import os
+import sys
 from urllib.request import pathname2url
 
 # Local version of website; will be updated later.
 HELP_URL = 'file:' + pathname2url(os.path.abspath('website/index.html'))
 
+# check the python version
+if sys.version_info > (3,4,0):
+    sys.stderr.write('You need Python 3.4 or later to run this script\n')
```
nigef: maybe say what this script is

pyta.py
```diff
@@ -14,11 +14,17 @@
...
 import webbrowser
 import os
+import sys
 from urllib.request import pathname2url
 
 # Local version of website; will be updated later.
 HELP_URL = 'file:' + pathname2url(os.path.abspath('website/index.html'))
 
+# check the python version
+if sys.version_info > (3,4,0):
```
david-yz-liu: @Jasminewwwu put spaces after commas as well.

pyta.py
david-yz-liu: Don't use stderr (print is fine), and the version check is not actually for running PyTA, but for the course itself.

Jasminewwwu: Then where should I put these lines?

david-yz-liu: As I said, use `print`

pyta.py
```diff
@@ -14,11 +14,17 @@
...
+import sys
 from urllib.request import pathname2url
 
 # Local version of website; will be updated later.
 HELP_URL = 'file:' + pathname2url(os.path.abspath('website/index.html'))
 
+# check the python version
+if sys.version_info < (3, 4, 0):
+    print('You need Python 3.4 or later to run this script\n')
+    exit(1)
```
david-yz-liu: @Jasminewwwu I realized that there shouldn't actually be an `exit` here. Just the warning message is enough.

david-yz-liu:
Thanks, @Jasminewwwu!

PR #55, opened 2016-06-05
Added comment extractor using astroid

WChrisK:
Example usage:

```python
>>> m = DocumentedModule('pyta.py')
```
```python
>>> m.classes
[<__main__.DocumentedClass object at 0x03456170>]
```
```python
>>> m.functions
[<__main__.DocumentedFunction object at 0x03456210>, <__main__.DocumentedFunction object at 0x03456250>]
```
```python
>>> m.classes[0]
<__main__.DocumentedClass object at 0x03456170>
```
```python
>>> m.classes[0].name
'PyTAReporter'
>>> m.classes[0].docstring
''
>>> m.classes[0].methods
[<__main__.DocumentedFunction object at 0x03457AF0>, <__main__.DocumentedFunction object at 0x03457B50>, <__main__.DocumentedFunction object at 0x03C0F790>]
```
```python
>>> for f in m.classes[0].methods:
	print(f.name, f.args, f.docstring)

	
__init__ ['self'] 
handle_message ['self', 'msg'] Handle a new message triggered on the current file.
print_message_ids ['self'] 
>>> print(m.functions[0].name, m.functions[0].args, m.functions[0].docstring)
check ['module_name'] Check a module for errors, printing a report.

```


david-yz-liu:
@WChrisK Your documentation in the PR description is excellent. Can you please add the same documentation to the module docstring?

comment_extractor.py
```diff
@@ -0,0 +1,49 @@
+from astroid.scoped_nodes import FunctionDef
+from astroid.scoped_nodes import ClassDef
+from astroid.manager import AstroidManager
+
+
+class DocumentedModule:
+    '''Encapsulates a module with all the classes and top level functions being
```
david-yz-liu: Please use double-quotes.

comment_extractor.py
```diff
@@ -0,0 +1,49 @@
...
+    documented.'''
+
+    def __init__(self, module_path):
+        '''Takes a Python file located at module_path, reads the AST and
+        generates the appropriate documentation objects.
+        '''
+        ast = AstroidManager().ast_from_file(module_path)
```
david-yz-liu: I think you can use `astroid.parse` here, which is the top-level API function (encouraged use).

david-yz-liu:
Oh, and just a quick reminder that you should *not* use regular expressions to attempt to parse the docstrings. I suggested a library to use in the original issue.

comment_extractor.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(self, node):
+        '''Takes the Astroid node and extracts the data. It should be a
+        ClassDef node.
+        '''
+        self.docstring = node.doc if node.doc else ''
+        self.name = node.name
+        self.methods = []
+        for n in node.body:
+            if isinstance(n, FunctionDef):
+                self.methods.append(DocumentedFunction(n))
```
nigef: why u no comment in the code

WChrisK: Do you think we need comments here? Is it trivial code, or no?

nigef: In general as you know, comments are helpful for future people to familiarize themselves with this code written by you. I understand it may seem trivial.

WChrisK: I just realized an even better way to do this based on a book for clean code that I'm reading, I'll submit this later tonight.
My goal as of late is to follow the "only comment when you have no choice, if you have to comment it means you're code is not readable and you have to redesign it", so I'm hoping to fulfill that tonight.

nigef: I assumed that's what you were doing -- sounds like a plan, as long as everyone, like @david-yz-liu is okay with it.. 👍

david-yz-liu: I'm not sure what prompted the original comment here. To me this seems like a pretty straightforward list building loop, which does not need a comment.

The place that really should have more comments is the class docstrings. These must document the attributes of the class.

nigef: Clearly there is not one comment in the entire file. I mistakenly commented on a specific line when I should have made a more general comment. Although it now seems like this maybe is not an issue.

nigef:
Is this where I comment on the file?

comment_extractor.py
david-yz-liu: @nigef since you edited your original comment, my comment makes less sense. Please don't do that.

nigef: Apologies, it was edited to reduce ambiguity. Next time I will leave the original comment, and append edits like this, `EDIT: ...`

david-yz-liu: You can also use ~~strikethrough~~

WChrisK:
I added unit tests now.
Requesting you to check it over and see if you're happy with the changes. Are you okay with the docstrings?

david-yz-liu:
@WChrisK I actually meant for you to include the "example usage" from the PR description in the module docstring.

But the class docstrings are alright. It would be better if you explicitly lists the class attributes, along with their types.

comment_extractor.py
```diff
@@ -0,0 +1,79 @@
...
+    nodes, and create all the DocumentedMethods and DocumentedFunctions, which
+    can be accessed through their respective fields."""
+
+    def __init__(self, module_path):
+        """Takes a Python file located at module_path, reads the AST and
+        generates the appropriate documentation objects.
+        module_path: Must be a valid path that can be read from, which is the
+        desired module to read.
+        """
+        ast = AstroidManager().ast_from_file(module_path)
```
WChrisK: TODO - Do the easier parse method (the comment was buried)

PR #58, opened 2016-06-06
Fixes#16 forbidden use of __import__

Jasminewwwu:


checkers/forbidden_import_call_checker.py
```diff
@@ -0,0 +1,36 @@
...
+    msgs = {'E9995': ('Call the function __import__ on %s',
+                      'func__import__not_allowed',
+                      'Used when you  call the function __import__ to import a '
+                      'module'),
+           }
+    # this is important so that your checker is executed before others
+    priority = -1
+
+    @check_messages("func__import__not_allowed")
+    def visit_call(self, node):
```
david-yz-liu: @Jasminewwwu I've given this one a little thought. Rather than create a new class, I would like you to add this method to the existing import checker class. You can use the same message as the existing checker, as well.

checkers/forbidden_import_checker.py
```diff
@@ -1,26 +1,42 @@
...
 from pylint.checkers import BaseChecker
+from pylint.checkers.utils import check_messages
 
 
 class ForbiddenImportChecker(BaseChecker):
     __implements__ = IAstroidChecker
 
     name = 'custom'
     msgs = {'E9999':
-                ('You may not import any modules - you imported \033[4;34m%s\033[0m on line %s.',
```
david-yz-liu: Don't change the message. You can use the same message for both checks.

examples/forbidden_import_example.py
```diff
@@ -1,4 +1,6 @@
 import re
+__import__(re)
```
david-yz-liu: Please pick a different module name. Also, the argument should be a string.

checkers/forbidden_import_checker.py
Jasminewwwu: Actually, I just changed the num of arguments in the message. In my checker, it simply detected the use of __import__ function.  And the message shows like this：
```
E9999: (forbidden-import)
    You may not import any modules - you imported re on line 1.
E9999: (forbidden-import)
    You may not import any modules - you use __import__ on line 2
```

david-yz-liu: I understand what you did. I'm saying not to do it like that, but instead use the same message for both checks.

Jasminewwwu:
@david-yz-liu  Now it shows like this:
```
E9999: (forbidden-import)
    You may not import any modules - you imported re on line 1.
E9999: (forbidden-import)
    You may not import any modules - you imported 'pyta' on line 2.
```

david-yz-liu:
Thanks, @Jasminewwwu!

PR #65, opened 2016-06-13
Fixes 64 check for invalid range

Jasminewwwu:


examples/invalid_range_index_example.py
```diff
@@ -0,0 +1,6 @@
+for i in range(0):
+    for j in range(0, 1, 3):
```
david-yz-liu: The example would be clearer if each of the loops were separate, not nested.

checkers/invalid_range_index_checker.py
```diff
@@ -0,0 +1,68 @@
...
+    priority = -1
+
+    @check_messages("invalid-range-index")
+    def visit_call(self, node):
+        if isinstance(node.func, astroid.Name):
+            name = node.func.name
+            # ignore the name if it's not a builtin (i.e. not defined in the
+            # locals nor globals scope)
+            if not (name in node.frame() or name in node.root()):
+                if name == 'range':
```
david-yz-liu: Merge this check with the previous line. Nesting level is too high.

checkers/invalid_range_index_checker.py
```diff
@@ -0,0 +1,68 @@
...
+                    # the arguments of 'range' call
+                    arg = node.args
+                    flag = True
```
david-yz-liu: Use a multi-line compound `if` condition rather than a flag. So something like

```python
if len(arg) == 0 ||
   not all([isinstance(literal_eval(c.as_string()), int) for c in arg]) ||
   ... :
    self.add_message...
```

pyta.py
```diff
@@ -53,7 +53,9 @@ def check(module_name):
                                 'checkers/global_variables_checker',
                                 'checkers/dynamic_execution_checker',
                                 'checkers/IO_Function_checker',
+                                'checkers/invalid_range_index_checker',
                                 'checkers/always_returning_checker'])
+
```
david-yz-liu: Again, please do not include extraneous whitespace changes.

examples/invalid_range_index_example.py
```diff
@@ -0,0 +1,14 @@
+for i in range(0):
+    i+= 1
```
david-yz-liu: You have a space missing here.

checkers/invalid_range_index_checker.py
```diff
@@ -0,0 +1,50 @@
...
+            # ignore the name if it's not a builtin (i.e. not defined in the
+            # locals nor globals scope)
+            if not (name in node.frame() or name in node.root()) and name == 'range':
+                # the arguments of 'range' call
+                arg = node.args
+                # check if there is no args in 'range' call
+                if len(arg) == 0 or \
+                        not all([isinstance(literal_eval(c.as_string()), int) for c in arg])\
+                    or (len(arg) == 1 and literal_eval(arg[0].as_string()) < 2)\
+                    or (len(arg) == 2 and literal_eval(arg[1].as_string()) < 2):
```
david-yz-liu: Seems like this will fail to give the error on `range(4, 5)`.

checkers/invalid_range_index_checker.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @check_messages("invalid-range-index")
+    def visit_call(self, node):
+        if isinstance(node.func, astroid.Name):
+            name = node.func.name
+            # ignore the name if it's not a builtin (i.e. not defined in the
+            # locals nor globals scope)
+            if not (name in node.frame() or name in node.root()) and name == 'range':
+                # the arguments of 'range' call
+                arg = node.args
```
david-yz-liu: Please use the `map` function to make the calls to `as_string` and `literal_eval` here, rather than separately later.

checkers/invalid_range_index_checker.py
```diff
@@ -0,0 +1,49 @@
...
+                lst = list(map(lambda z: literal_eval(z.as_string()), arg))
+                # check if there is no args in 'range' call
+                if len(arg) == 0 or \
+                        not all([isinstance(literal_eval(c.as_string()), int) for c in arg])\
```
david-yz-liu: You should be able to use `lst` in this line as well.

checkers/invalid_range_index_checker.py
```diff
@@ -0,0 +1,49 @@
...
+    @check_messages("invalid-range-index")
+    def visit_call(self, node):
+        if isinstance(node.func, astroid.Name):
+            name = node.func.name
+            # ignore the name if it's not a builtin (i.e. not defined in the
+            # locals nor globals scope)
+            if not (name in node.frame() or name in node.root()) and name == 'range':
+                # the arguments of 'range' call
+                arg = node.args
+                lst = list(map(lambda z: literal_eval(z.as_string()), arg))
```
david-yz-liu: Give a better name than `lst` here.

checkers/invalid_range_index_checker.py
```diff
@@ -0,0 +1,49 @@
...
+                # check if there is no args in 'range' call
+                if len(arg) == 0 or \
```
david-yz-liu: Change the formatting of the compound condition as follows:

```python
if (len(arg) == 0 or
    not all(...) or
    (len(arg) == 1 and lst[0] < 2) or
    ...):
```

Each condition separated by an `or` is on a new line, vertically aligned, with all the `or`s going at the ends of the lines.

checkers/invalid_range_index_checker.py
```diff
@@ -0,0 +1,51 @@
...
+    def visit_call(self, node):
+        if isinstance(node.func, astroid.Name):
+            name = node.func.name
+            # ignore the name if it's not a builtin (i.e. not defined in the
+            # locals nor globals scope)
+            if not ((name in node.frame() or name in node.root()) and
+                            name == 'range'):
+                # the arguments of 'range' call
+                arg = node.args
+                eval_parm = list(map(lambda z: literal_eval(z.as_string()), arg))
```
david-yz-liu: @Jasminewwwu I realized after I merged this in that this line will raise an error if the range input is not a constant. I have disabled this checker until this problem is fixed.

In general, we should never use `eval`, so you need to find a new approach.

PR #69, opened 2016-06-21
Group errors in Reporter

### LINKED ISSUE #67: Group errors together in reporter
Right now, I believe the errors detected by PyTA are sorted by their code before being output. However, we should make two modifications:

1. The error code and name should only appear once, and then underneath the individual messages (with line numbers). It should be obvious what message corresponds to the error code, so good use of whitespace there matters.
2. We probably should limit the number of messages shown for a particular error, but show a message saying something like "This error also occurs in XX more places."

@shweta97 I'm assigning you this task since you've been working on the reporter already.

shweta97:
Groups errors in reporter class by limiting the number of times a particular error appears. Fixes #67.


shweta97:
I created this branch off my other branch reporters which hasn't been merged yet. Hence, this pull request shows more modifications than the actual ones in this branch.

reporters/plain_reporter.py
```diff
@@ -12,8 +13,52 @@ def handle_message(self, msg):
...
+        # Sort the messages by their type.
+        self._messages.sort(key=lambda s: s[0])
+
+        i = 0
+        while i < len(self._messages):
+            current_id = self._messages[i].msg_id
+            count = 1
+            messages = ''
+            msg_count = 1
+            while i+1 < len(self._messages) and self._messages[i+1].msg_id == current_id:
```
david-yz-liu: Put spaces around each `+`

reporters/plain_reporter.py
```diff
@@ -12,8 +13,52 @@ def handle_message(self, msg):
...
+                count += 1
+                if msg_count <= 5:
+                    messages += '\n    ' + self._messages[i+1].msg
+                    msg_count += 1
+                self._messages.pop(i+1)
+
+            msg = self._messages[i].msg
+
+            if count > 1:
+                msg = 'This error occurs at '+ str(count) + ' places:\n    ' + self._messages[i].msg + messages
```
david-yz-liu: Check whitespace

reporters/plain_reporter.py
```diff
@@ -12,8 +13,52 @@ def handle_message(self, msg):
...
+            msgid = self._messages[i].msg_id
+            symbol = self._messages[i].symbol
+            confidence = self._messages[i].confidence
+            abspath = self._messages[i].abspath
+            path = self._messages[i].path
+            module = self._messages[i].module
+            obj = self._messages[i].obj
+            line = self._messages[i].line
+            column = self._messages[i].column
+            self._messages[i] = Message(msgid, symbol, (abspath, path,
```
david-yz-liu: You can use the `_replace` method for `namedtuple` here to just replace the value of one field. Note that this method returns a new namedtuple, so you'd need to do something like `self._messages[i] = self._message[i]._replace(...)`

reporters/plain_reporter.py
```diff
@@ -12,8 +13,52 @@ def handle_message(self, msg):
...
+                                                            module, obj, line,
+                                                            column), msg,
+                                            confidence)
+
+            #Message(msgid, symbol='forbidden-global-variables', msg="Global variables should not be used in CSC108/CSC148 - you used the keyword 'global' on line 7", C='E', category='error', confidence)
+
+            count = 0
```
david-yz-liu: This line is likely unnecessary

reporters/plain_reporter.py
david-yz-liu: As we discussed in the meeting, this string should not appear here as a message, but only when the messages are reported.

It is likely necessary to create a new attribute, or use an existing one, to store the counts for each message type so that they can be reported.

reporters/plain_reporter.py
```diff
@@ -12,8 +13,52 @@ def handle_message(self, msg):
...
+        i = 0
+        while i < len(self._messages):
+            current_id = self._messages[i].msg_id
+            count = 1
+            messages = ''
+            msg_count = 1
+            while i+1 < len(self._messages) and self._messages[i+1].msg_id == current_id:
+                count += 1
+                if msg_count <= 5:
+                    messages += '\n    ' + self._messages[i+1].msg
```
david-yz-liu: Make `messages` a list of strings rather than a string; this allows you to do a `len` to make the "max 5" check. You can call `'\n    '.join(messages)` later on.

david-yz-liu:
@shweta97 if you are confident your code works, then you should remove all the commented out code.

reporters/plain_reporter.py
```diff
@@ -11,9 +12,34 @@ def handle_message(self, msg):
...
         # Sort the messages by their type.
         self._messages.sort(key=lambda s: s[0])
 
-        for msg in self._messages:
-            code = msg.msg_id
-            print(code, '({})\n    {}'.format(msg.symbol, msg.msg))
+        i = 0
+        while i < len(self._messages):
+            current_id = self._messages[i].msg_id
+            count = 1
```
david-yz-liu: It appears this variable is no longer needed.

reporters/plain_reporter.py
```diff
@@ -11,9 +12,34 @@ def handle_message(self, msg):
...
+            messages = []
+            while i + 1 < len(self._messages) and self._messages[i + 1].msg_id == current_id:
+                count += 1
+                if len(messages) <= 5:
+                    messages.append(self._messages[i+1].msg)
```
david-yz-liu: Put spaces around `+`.

When I give you feedback about style, please check over all of your code changes so that I don't need to point out the same error at every location.

reporters/plain_reporter.py
```diff
@@ -11,9 +12,34 @@ def handle_message(self, msg):
...
+                self._messages.pop(i+1)
+
+            msg_new = self._messages[i].msg
+
+            if len(messages) > 0:
+                temp = '\n    ' + '\n    '.join(messages)
```
david-yz-liu: Don't use a temporary variable; I believe the update can be done with `+=`, and fit on one line.

reporters/plain_reporter.py
```diff
@@ -11,9 +12,34 @@ def handle_message(self, msg):
...
+                msg_new = self._messages[i].msg + temp
+
+            obj_new = 'This error occurs at ' + str(count) + ' places:'
+            self._messages[i] = self._messages[i]._replace(msg=msg_new)
```
david-yz-liu: You should be able to do this in just one call to `_replace`, rather than two.

reporters/plain_reporter.py
```diff
@@ -1,4 +1,5 @@
 from pylint.reporters import BaseReporter
+from pylint.utils import Message
```
david-yz-liu: Please check whether this import is still necessary.

reporters/plain_reporter.py
```diff
@@ -11,9 +12,34 @@ def handle_message(self, msg):
...
+
+            msg_new = self._messages[i].msg
+
+            if len(messages) > 0:
+                temp = '\n    ' + '\n    '.join(messages)
+                msg_new = self._messages[i].msg + temp
+
+            obj_new = 'This error occurs at ' + str(count) + ' places:'
+            self._messages[i] = self._messages[i]._replace(msg=msg_new)
+            self._messages[i] = self._messages[i]._replace(obj=obj_new)
```
david-yz-liu: This is an oddly-named attribute. What was its value before you replaced it?

reporters/plain_reporter.py
shweta97: I use this variable to add the number of times this error appears in the message.

reporters/plain_reporter.py
shweta97: It was just an empty string. That's why I thought of using it.

reporters/plain_reporter.py
david-yz-liu: Ah yes, thanks for point that out.

david-yz-liu:
Thanks, @shweta97!

PR #70, opened 2016-06-21
Checking constraints

hayleyZHL:


type_inference_visitor.py
```diff
@@ -0,0 +1,93 @@
...
+                visited = self._visit_generic(value)
+                setattr(node, field, visited)
+        return self._transform(node)
+
+    def _transform(self, node):
+        """Call matching transforms for the given node if any and return the
+        transformed node.
+        """
+        node_class = node.__class__
+        if node_class not in self.transforms:
```
david-yz-liu: I believe we already discussed this, but I don't think this block is necessary. The whole point of the `register_transform` is to tell the visitor what nodes to perform actions on.

type_inference_visitor.py
```diff
@@ -0,0 +1,93 @@
...
+
+def set_const(node):
+    """Populate type constraints for astroid nodes."""
+    if isinstance(node, astroid.Const):
+        # # astroid.Const represent a constant node like num/str/bool/None/bytes.
+        # if isinstance(node.parent, astroid.UnaryOp):
+        #     print(node.value)
+        #     Expr(value=astroid.UnaryOp(op=USub(), operand=Name(id='a', ctx=Load())))
+        node.type_constraints = [type(node.value)]
+    elif isinstance(node, astroid.Tuple):
```
nigef: I thought it was decided to make one function per node type.

hayleyZHL: @nigef  Yes, this is my work from last week not from this week, you have not seen my update yet, which I will update later.

examples/type_inference/type_inference_visitor_unittest.py
```diff
@@ -0,0 +1,243 @@
+"""Unittest for the type_inference_visitor."""
```
david-yz-liu: This file should be under `test/`. You can create a separate subdirectory called `type_inference`.

examples/type_inference/type_inference_visitor_unittest.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+class SetConstFunctionTest(unittest.TestCase):
+    """testers specifically for the function set_const using single nodes."""
+
+    def test_const_str(self):
+        """testing if the function set_const has the correct output for node
+        type str."""
+        expected = str
+        node = test_utils.extract_node("""a = __('sample_string')""")
```
david-yz-liu: You should be able to just have the expression be `'sample_string'`, i.e., you don't need an assignment for these nodes.

examples/type_inference/type_inference_visitor_unittest.py
```diff
@@ -0,0 +1,243 @@
...
+        set_const(node)
```
david-yz-liu: Also, you shouldn't have to call `set_const` directly here - you can use the same approach as in the other tests, i.e., creating a Visitor object

examples/type_inference/type_inference_visitor_unittest.py
```diff
@@ -0,0 +1,243 @@
...
+        Module(body=[
+        Expr(value=UnaryOp(op=USub(), operand=Name(id='a', ctx=Load()))),
+        ])
+        """
+        type_visitor = TypeVisitor()
+        type_visitor.register_transform(astroid.Const, set_const)
+        type_visitor.register_transform(astroid.BinOp, set_binop)
+        module = astroid.parse("""10 + 2""")    # int
+        transformed_module = type_visitor.visit(module)
+        for node in transformed_module.body:
```
david-yz-liu: Look up the `nodes_of_class` method. I believe you can use this.

type_inference_visitor.py
```diff
@@ -0,0 +1,126 @@
+from astroid.transforms import TransformVisitor
+import astroid
+import warnings
+from typing import Tuple, List, Dict
+
+
+class TypeVisitor(TransformVisitor):
+
+    def _visit(self, node):
```
david-yz-liu: I would guess that you don't need to override this method.

type_inference_visitor.py
```diff
@@ -0,0 +1,126 @@
...
+        if hasattr(node, '_astroid_fields'):
+            for field in node._astroid_fields:
+                value = getattr(node, field)
+                visited = self._visit_generic(value)
+                setattr(node, field, visited)
+        return self._transform(node)
+
+    def _visit_generic(self, node):
```
david-yz-liu: Or this one.

type_inference_visitor.py
```diff
@@ -0,0 +1,126 @@
...
+        if isinstance(node, list):
+            return [self._visit_generic(child) for child in node]
+        elif isinstance(node, tuple):
+            return tuple(self._visit_generic(child) for child in node)
+        else:
+            return self._visit(node)
+
+    # def _visit(self, node):
```
david-yz-liu: And this can be removed.

type_inference_visitor.py
```diff
@@ -0,0 +1,126 @@
...
+    #             value = getattr(node, field)
+    #             visited = self._visit_generic(value)
+    #             setattr(node, field, visited)
+    #     return self._transform(node)
+
+
+def set_const(node):
+    """Populate type constraints for astroid nodes."""
+    # astroid.Const represent a constant node like num/str/bool/None/bytes.
+    result = type(node.value)
```
david-yz-liu: Don't use a temporary variable.

type_inference_visitor.py
```diff
@@ -0,0 +1,126 @@
...
+    #     if hasattr(node, '_astroid_fields'):
+    #         for field in node._astroid_fields:
+    #             value = getattr(node, field)
+    #             visited = self._visit_generic(value)
+    #             setattr(node, field, visited)
+    #     return self._transform(node)
+
+
+def set_const(node):
+    """Populate type constraints for astroid nodes."""
```
david-yz-liu: It would be good to specify the precise node type (`Const`) that this method is for.

type_inference_visitor.py
```diff
@@ -0,0 +1,126 @@
...
+    #         return self._transform(node)
+    #     if hasattr(node, '_astroid_fields'):
+    #         for field in node._astroid_fields:
+    #             value = getattr(node, field)
+    #             visited = self._visit_generic(value)
+    #             setattr(node, field, visited)
+    #     return self._transform(node)
+
+
+def set_const(node):
```
david-yz-liu: Change the name to `set_const_type_constraints`, and similarly for the other methods below.

type_inference_visitor.py
```diff
@@ -0,0 +1,126 @@
...
+    # node_types contains types of elements inside tuple.
+    node_types = [node_child.type_constraints for node_child in node.elts]
+    # Since tuple has only 2 elements, node.type_constraints will return
+    # types of both elements.
+    node.type_constraints = Tuple[node_types[0], node_types[1]]
+
+
+def set_list(node):
+    # node_types contains types of elements inside list.
+    node_types = []
```
david-yz-liu: Look up Python `set`s, which provide a more elegant representation. https://docs.python.org/3.5/library/stdtypes.html?highlight=set#set

type_inference_visitor.py
```diff
@@ -0,0 +1,126 @@
...
+    for node_child in node.elts:
+        if node_child.type_constraints not in node_types:
+            node_types.append(node_child.type_constraints)
+
+    # if list has more than one types, just set node.type_constraints to
+    # list, if list has only 1 types, set the node.type_constraints to be
+    # List of that type.
+    if len(node_types) == 1:
+        node.type_constraints = List[node_types[0]]
+    elif len(node_types) == 0 or len(node_types) > 1:
```
david-yz-liu: You can do an `else` here.

type_inference_visitor.py
```diff
@@ -0,0 +1,126 @@
...
+        node.type_constraints = Dict
+
+
+def set_binop(node):
+    op = node.op
+    left_operand = node.left.type_constraints
+    right_operand = node.right.type_constraints
+
+    node.type_constraints = left_operand
+    if right_operand == int and node.type_constraints ==float:
```
david-yz-liu: There can be fewer conditions. Do something like:

```
if ((left_operand == float and right_operand == int) or
    (left_operand == int and right_operand == float)):
    node.type_constraints = float
elif ...:
    ... 
else:
    warnings.warn(...)
```

type_inference_visitor.py
```diff
@@ -0,0 +1,126 @@
...
+                      'might have a type error.' % node)
+
+
+def set_unaryop(node):
+    op = node.op
+    operand = node.operand.value
+    node.type_constraints = type(operand)
+
+
+if __name__ == '__main__':
```
david-yz-liu: This can be removed. The example usage will exist in our tests.

david-yz-liu:
@hayleyZHL I've made some comments. Your first priority should be cleaning up this pull request (including making the necessary changes to `Tuple` and `Dict` inference discussed in the meeting), so I can merge this in.

After, you can continue your work on extending your work to more complex combinations of these basic operations (e.g., what you were talking about with nested BinOps).

examples/type_inference/type_inference_visitor_unittest.py
hayleyZHL: Hi David, I added an assignment for the node 'sample_string' because that's the instruction in `test_utils.extract_node` function, [https://github.com/landscapeio/astroid/blob/master/test_utils.py](url) [line108-113]
`    Expressions:
     To extract arbitrary expressions, surround them with the fake
     function call __(...). After parsing, the surrounded expression
     will be returned and the whole AST (accessible via the returned
     node's parent attribute) will look like the function call was
     never there in the first place.`

hayleyZHL: I tried without an assignment, this is the error I got: 
`Error
  File "/Users/HL/Desktop/pyta/tests/type_inference/type_inference_visitor_unittest.py", line 19, in test_const_str
    node = test_utils.extract_node("""'sample_string'""")
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/astroid/test_utils.py", line 163, in extract_node
    extracted.append(tree.body[-1])
IndexError: list index out of range`

examples/type_inference/type_inference_visitor_unittest.py
hayleyZHL: thanks! fixed

examples/type_inference/type_inference_visitor_unittest.py
hayleyZHL: fixed.

type_inference_visitor.py
hayleyZHL: fixed

type_inference_visitor.py
hayleyZHL: fixed

type_inference_visitor.py
hayleyZHL: fixed

type_inference_visitor.py
hayleyZHL: I found that we don't need to override anything in TransfromVistor.. should I just use TransfromVisitor instead of TypeVisitor?

type_inference_visitor.py
hayleyZHL: fixed

type_inference_visitor.py
hayleyZHL: fixed

type_inference_visitor.py
hayleyZHL: fixed

examples/type_inference/type_inference_visitor_unittest.py
hayleyZHL: I changed the code to 
`visited_module = type_visitor.visit(module)
self.assertEqual(Dict, visited_module.body[0].value.type_constraints)` instead of the for loop. It gives desired result. But I'm not sure what `nodes_of_class` is for.. since I'm checking the type constraints, I called `node.type_constarints` and comparing it with the correct class type.

type_inference_visitor.py
david-yz-liu: Sure, I suppose that means all the behaviour is specified using `register_transform`

examples/type_inference/type_inference_visitor_unittest.py
david-yz-liu: @hayleyZHL I asked you to look into `nodes_of_class`. Saying that you don't know what it's for is not an acceptable response. I expect you to be able to learn on your own "what it is for."

examples/type_inference/type_inference_visitor_unittest.py
david-yz-liu: @hayleyZHL Okay, let me put it this way. I see no reason to use `extract_node` at all, instead of `astroid.parse` like the tests for all the other nodes.

type_inference_visitor.py
```diff
@@ -0,0 +1,93 @@
...
+    """Populate type constraints for astroid nodes num/str/bool/None/bytes."""
+    node.type_constraints = node.value.__class__
+
+
+def set_tuple_type_constraints(node):
+    # node_types contains types of elements inside tuple.
+    node_types = [node_child.type_constraints for node_child in node.elts]
+    # Since tuple has only 2 elements, node.type_constraints will return
+    # types of both elements.
+    node.type_constraints = Tuple[node_types[0], node_types[1]]
```
david-yz-liu: @hayleyZHL This is a reminder that this needs to be fixed.

type_inference_visitor.py
```diff
@@ -0,0 +1,93 @@
...
+        node.type_constraints = Dict
+
+
+def set_binop_type_constraints(node):
+    op = node.op
+    left_operand = node.left.type_constraints
+    right_operand = node.right.type_constraints
+
+    node.type_constraints = left_operand
+    if ((right_operand == int and node.type_constraints ==float) or
```
david-yz-liu: I believe I pointed out a whitespace error here that you did not fix, even though you changed this line.

Please be aware that every time you push a change to a line, GitHub won't show the previous line comment inline any more. This means that you need to make special effort to make all the required changes for a given line, not just one of them.

type_inference_visitor.py
```diff
@@ -0,0 +1,93 @@
...
+        (right_operand == float and node.type_constraints == int)):
+        node.type_constraints = float
+    elif right_operand==left_operand:
```
david-yz-liu: There is a whitespace issue here as well.

In general, please be careful with your whitespace before committing. I don't want to keep having to point these out.

david-yz-liu:
@hayleyZHL your changes look fine.

For the `Tuple` part, if I recall correctly we did something like `Tuple.__getitem__(tuple(*the_list))`.

Also, I noticed this is actually targeting the wrong branch. You shouldn't be making this pull request into master, but rather into the `typecheck` branch. Please remake this pull request.

PR #71, opened 2016-06-24
Standardized some examples

nigef:
made some more PEP 8 compliant, and edited some other issues.


david-yz-liu:
This is great, thanks @nigef.

PR #72, opened 2016-06-26
Fixes#66 Assigning to self checker

Jasminewwwu:


checkers/assigning_to_self_checker.py
```diff
@@ -0,0 +1,32 @@
...
+                      'Used when you assign a value to self'),
+           }
+
+    # this is important so that your checker is executed before others
+    priority = -1
+
+    @check_messages("assigning_to_self")
+    def visit_assign(self, node):
+        target = node.targets[0]
+        if isinstance(target, astroid.AssignName) and target.as_string() == "self":
```
david-yz-liu: Please check the attributes of `target` to see if you can directly access the name being assigned to, rather than converting the node into a string with `as_string`

PR #74, opened 2016-06-27
fixed example that was weird

nigef:
i don't know why this example is used, but it had an annoying mistake.


david-yz-liu:
@nigef don't use words like "annoying" or "weird" when describing people's work.

nigef:
Thanks for pointing that out. I completely forgot a person is 'attached' to the work, if that makes sense. 
Hope I will always remember this little thing..

PR #76, opened 2016-06-30
Travis

WChrisK:


WChrisK:
I need some input on doing the unit test calls rather than hardcoding. Using `nose` didn't work too well, though I think I need some more time to mess around with it and maybe I can get it working.

I could try to dynamically run the tests from a script (despite that being exactly what the other libraries are supposed to do...), however I think the reason they failed might be due to using ```__name__ == __main__``` code, since running them manually gave the desired effect.

I need to look into this more.

nigef:
👍 Travis

david-yz-liu:
@WChrisK I looked into this. It should be possible to ditch the `unittest` module and instead use nose's test generators: https://nose.readthedocs.io/en/latest/writing_tests.html#test-generators.

This should actually simplify the code a little. Why don't you work on converting your test suite to use `nose` first (and make a pull request, of course), and then we'll revisit Travis.

Issue #75, opened 2016-06-30
HTML Reporter


david-yz-liu:
We should create a third reporter class called `HTMLReporter`, which generates basically the same output as the other reporters, except in HTML form.

Note that the output is still just a string printed to the console, but now that string looks like `"<div><h1>E9999 ...</h1><p>Some message...</p></div>"`, i.e., contains the actual HTML tags.

We will use the [Jinja template library](http://jinja.pocoo.org/) to generate the HTML strings. Note that it's likely `jinja` will need to be pip-installed.

For now, focus only on generating the raw HTML, and don't worry about styling. It is possible to preview what the HTML looks like raw by pasting the output into a text file, and opening it in a web browser.

PR #78, opened 2016-07-04
HTML reporter

shweta97:
Addresses #75. Created a new class that outputs in html.


david-yz-liu:
@shweta97 you're getting too used to building branches off of previous pull requests.
In this case, there is no reason to do so; please submit a pull request which is built off of master and introduces the new reporter class.

PR #79, opened 2016-07-05
Categorize nodes

nigef:
Succinct examples of each AST Node relevant to python3


nodes/Assert.py
```diff
@@ -0,0 +1 @@
+assert (x)
```
david-yz-liu: Probably don't need parentheses around `x`, unless some checker complained?

nodes/AssignName.py
```diff
@@ -0,0 +1,2 @@
+for x in range(3):
```
david-yz-liu: This is unexpected. Is `AssignName` not just a child of `Assign`? (In which case, perhaps use the same example as `Assign` for clarity.)

nodes/Compare.py
```diff
@@ -0,0 +1,2 @@
+if (3 > 2):
```
david-yz-liu: No parens around condition

david-yz-liu: Oh, but you could also just do `3 > 2` as a single expression, rather than the `if`, I assume.

nodes/Const.py
```diff
@@ -0,0 +1 @@
+x = 1
```
david-yz-liu: Here I'd like to see just the expression `1` rather than an assignment statement.

nodes/Dict.py
```diff
@@ -0,0 +1 @@
+a = {"b": 1}
```
david-yz-liu: Please use single-quotes here.

nodes/GeneratorExp.py
```diff
@@ -0,0 +1 @@
+print(g for ip in [num for num in range(1, 9)])
```
david-yz-liu: You don't need the call to print - just the parentheses should be enough.

Also, you don't need the nested list comprehension: `(num for num in range(10))` should work just fine, yeah?

nodes/If.py
```diff
@@ -0,0 +1,2 @@
+if (3 > 2):
```
david-yz-liu: No parens around conditions

nodes/IfExp.py
```diff
@@ -0,0 +1,2 @@
+n = None
+x = 1 if n else 0
```
david-yz-liu: Let's not use a temporary variable: `1 if True else 0`

nodes/Name.py
```diff
@@ -0,0 +1 @@
+[l+l for l in 'str']
```
david-yz-liu: This is interesting. Is not just a single name like `my_var` enough to make the `Name` node?

nodes/Yield.py
```diff
@@ -0,0 +1,3 @@
+class What():
```
david-yz-liu: No parens after `What`

david-yz-liu: Actually, please don't use a class here.

david-yz-liu:
Thanks, @nigef!

PR #80, opened 2016-07-11
HTML reporter

shweta97:
Created a HTML Reporter that outputs in HTML format. Addresses issue #75.


reporters/html_reporter.py
```diff
@@ -0,0 +1,25 @@
...
+        super().__init__()
+
+    # Override this method
+    def print_message_ids(self):
+        # Sort the messages.
+        self.sort_messages()
+
+        #template = Environment(keep_trailing_newline=True, autoescape=False).from_string(HTML)
+        template = Environment().get_template('template.txt')
+        f = open('output.html', 'w')
```
david-yz-liu: Use `with open(...) as f: ...` (you may need to look up the purpose of this syntax)

reporters/html_reporter.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self):
+        super().__init__()
+
+    # Override this method
+    def print_message_ids(self):
+        # Sort the messages.
+        self.sort_messages()
+
+        #template = Environment(keep_trailing_newline=True, autoescape=False).from_string(HTML)
+        template = Environment().get_template('template.txt')
```
david-yz-liu: Good stuff. You need to check in `template.txt` as well then. For now, you can put it under `reporters/templates/template.txt`.

reporters/html_reporter.py
```diff
@@ -0,0 +1,25 @@
...
+        f = open('output.html', 'w')
+
+        for msg in self._messages:
+            code = msg.msg_id
+            msg_new = msg.msg.replace('\n', '<br/>')
```
david-yz-liu: I recommend first doing `replace('\r', '')` and then `replace('\n', '<br/>')` to make this compatible with multiple line ending types.

reporters/html_reporter.py
```diff
@@ -0,0 +1,25 @@
...
+            msg = msg._replace(msg=msg_new)
+            f = open('output.html', 'a')
```
david-yz-liu: In general, think carefully about when you are using files, which is why `with` is so powerful. Your code has the error that it opens `output.html` too many times.

reporters/html_reporter.py
```diff
@@ -0,0 +1,25 @@
...
+            f.write(template.render(code=code,symbol=msg.symbol,obj=msg.obj,msg=msg.msg))
+            print(template.render(code=code,symbol=msg.symbol,obj=msg.obj,msg=msg.msg))
```
david-yz-liu: Probably don't need to do both writing to a file and printing.

Why don't you add a parameter to the constructor, and an attribute, which is either a filename to output the HTML to, or `None`, in which case the output is printed to stdout.

shweta97: @david-yz-liu: I was thinking of creating a new issue for outputting to a file in general for all reporters. What do you think?

reporters/templates/template.txt
```diff
@@ -0,0 +1,9 @@
+<html>
+<head>
+</head>
+<body>
+    <p>{{ code }} ({{symbol}}) {{obj}}</p>
```
david-yz-liu: @shweta97 I asked you to look into use a `for` loop in here, to only repeat the `<p>` part for each message, not the entire template.

david-yz-liu: @shweta97 reminder about this one

david-yz-liu:
Thanks, @shweta97!

(Don't forget to update your other pull request to make the `number_of_messages` attribute for this class as well.

PR #83, opened 2016-07-13
Now works with nosetests

WChrisK:
~~Not ready yet, just committing. However you can run `nosetests` in the tests directory.~~
Now works with `nosetest --config=nose.cfg`


tests/test_examples.py
```diff
@@ -1,15 +1,14 @@
-import unittest
 import os
 import os.path
 import subprocess
 import re
 
 
-_EXAMPLES_PATH = 'examples/pylint/'
+_EXAMPLES_PATH = '../examples/pylint/'
 _EXAMPLE_PREFIX_REGEX = '[CEFRW]\d{4}'
```
WChrisK: I don't like relative paths, going to fix this with a nose config if possible.

tests/test_examples.py
```diff
@@ -21,42 +20,34 @@ def get_test_file_paths():
     return test_files
 
 
-def create_test_function(test_file, checker_name):
+def create_function(test_file, checker_name):
```
WChrisK: Is this function name too generic?

WChrisK:
Some notes for discussion on why some tests continue failing after digging around in the source code:

---

signature-differs (W0222):
 	Signature differs from %s %r method Used when a method signature is different than in the implemented interface or in an overridden method.

https://github.com/PyCQA/pylint/blob/ad43689aed84accf51223443acbd9ded1900576a/pylint/checkers/classes.py#L1123

        if _different_parameters(refmethod, method1):
            self.add_message('arguments-differ',
                             args=(class_type, method1.name),
                             node=method1)
        elif len(method1.args.defaults) < len(refmethod.args.defaults):
            self.add_message('signature-differs',
                             args=(class_type, method1.name),
                             node=method1)

---

no-init (W0232):
 	Class has no __init__ method Used when a class has no __init__ method, neither its parent classes.

https://github.com/PyCQA/pylint/blob/ad43689aed84accf51223443acbd9ded1900576a/pylint/checkers/classes.py#L454

        if node.type == 'class' and has_known_bases(node):
            try:
                node.local_attr('__init__')
            except astroid.NotFoundError:
                self.add_message('no-init', args=node, node=node)

examples/pylint/C0121_singleton_comparison.py
```diff
@@ -3,7 +3,7 @@ def is_true():
     @rtype: bool
     """
     temp = 5
-    if temp is None:
+    if temp == None:
```
nigef: oh, did `is` this not raise the intended message?

tests/test_examples.py
```diff
@@ -16,7 +25,9 @@ def get_file_paths():
     test_files = []
     for root, directories, files in os.walk(_EXAMPLES_PATH, topdown=True):
         for filename in files:
-            test_files.append(_EXAMPLES_PATH + filename)
+            if filename not in ignored_tests:
+                if filename == 'R0101_too_many_nested_blocks.py':
```
nigef: curious, why this file needs special treatment..

examples/pylint/C0121_singleton_comparison.py
david-yz-liu: I believe this check's purpose is to make sure you compare against singletons with `is` instead of `==`

examples/pylint/R0101_too_many_nested_blocks.py
```diff
@@ -1,9 +1,18 @@
-def my_f(num):
```
david-yz-liu: @WChrisK I'm not sure why you changed this example.

examples/pylint/W0231_super_init_not_called.py
```diff
@@ -5,4 +5,4 @@ def __init__(self):
 
 class Child(Parent):
     def __init__(self):
-        Parent.__init__(self)  # You must have this
+        pass #Parent.__init__(self)  # You must have this
```
david-yz-liu: You can put the comment on the line above - it's clearer than having it on the same line as the `pass`

examples/pylint/W0232_no_init.py
```diff
@@ -9,3 +9,5 @@ def does_more_stuff(self, n, m):
 class ClassWithNoInitTwo(ClassWithNoInit):
```
david-yz-liu: @WChrisK You can delete this check (please add it to the disable list, and remove it from the website documentation as well).

tests/test_examples.py
david-yz-liu: Hmm okay, that's fine.

tests/test_examples.py
david-yz-liu: Maybe `create_checker_test`. The important noun is "test," not "function."

tests/test_examples.py
```diff
@@ -1,62 +1,63 @@
...
-class TestExamples(unittest.TestCase):
-    """The test class to be populated dynamically."""
-    pass
-
-
-if __name__ == '__main__':
-    make_new_unit_tests()
-    unittest.main()
+        test_function = create_function(test_file, checker_name)
+        yield test_function
```
david-yz-liu: Even though this is exactly what I said, I'm surprised it was as simple as I thought. Yay!

david-yz-liu:
Thanks, @WChrisK. Note that nosetests looks in the `tests/` folder by default, so you don't need to specify that (plus the pathing got messed up). I fixed this in 04e926b.

PR #87, opened 2016-07-18
Max Number of Messages

shweta97:
Added a new argument to the constructor for reporters to set the maximum number of messages that appear. If it is 0, all the messages will be printed. The default is set to 5 messages.


reporters/color_reporter.py
```diff
@@ -6,8 +6,8 @@
 
 
 class ColorReporter(PlainReporter):
-    def __init__(self):
-        super().__init__()
+    def __init__(self, number):
+        super().__init__(number)
```
david-yz-liu: Please use a more descriptive attribute and parameter name.

david-yz-liu: @shweta97 reminder about this one.

david-yz-liu:
@shweta97 I just noticed you didn't fix this for the Color reporter.

It would be nice if you extracted the code into a helper method in the super class that you could use without change in the subclass.

shweta97:
@david-yz-liu I'm not sure if I understand exactly what you mean. The code that adds only a limited number of messages is in the method sort_messages in the super class and all the other reporters call it.

PR #88, opened 2016-07-20
Ending locations

nigef:
`end_lineno` seems correctly set for all nodes. Still working on end_col_offset, since there are some edge cases that fail with Astroid.

TODO: attach more transform functions, and add at least one test function for each one.


setendings.py
```diff
@@ -0,0 +1,624 @@
...
+    if node.tolineno is None:
+        # TODO: write new code to gracefully handle, if this is ever reached.
+        logging.error("⛔ lineno is None at node: {}".format(node))
+        sys.exit(1)
+    else:
+        node.end_lineno = node.tolineno
+
+def _set_end_col_offset_by_string_old(node):
+    """Set the end_col_offset property by the as_string attribute.
+    Uses the end column location of the last line, instead of the widest line.
```
nigef: this is actually incorrect. it uses the old way of max line width.

setendings.py
```diff
@@ -0,0 +1,624 @@
...
+################################################################################
+def set_general(node):
+    """General function called by many of the other transform functions.
+    Populate ending locations for astroid node."""
+    _set_end_lineno(node)
+    _set_end_col_offset_by_string(node)
+
+#######
+
+def set_arguments(node):
```
david-yz-liu: You don't need to define separate functions. You just need to register_transform with the `set_general` directly.

setendings.py
```diff
@@ -0,0 +1,624 @@
...
+
+
+
+
+def _set_end_col_offset_by_value(node):
+    """Set the end_col_offset property by the value attribute.
+    """
+    # Guard: should have .value property
+    if 'value' not in dir(node):
+        raise Exception('''⛔️ node must have the .value property.''')
```
david-yz-liu: Please stop using non-ASCII characters in your messages.

setendings.py
```diff
@@ -0,0 +1,624 @@
...
+# Set the log level (DEBUG, ERROR, ...), and message format.
+logging.basicConfig(format='', level=logging.DEBUG)
+
+# Helper functions to abstract common tasks. Global for now..
+# these functions are associated to types in the transforms dictionary in the
+# TransformVisitor class. The _transform method invokes the transform function
+# on the nodes.
+# TODO: Separate into Class? .. would need static methods.
+################################################################################
+def _set_end_lineno(node):
```
david-yz-liu: This is a duplicate function name. I'm assuming you should delete this version.

ending_location_visitor.py
```diff
@@ -1,32 +1,223 @@
...
+import unittest
+import logging
+from setendings import *
+from inspect_ast import visit_astroid
+
+# Set the log level (DEBUG, ERROR, ...), and message format.
+logging.basicConfig(format='', level=logging.DEBUG)
+
+
+class EasyTransformVisitor(TransformVisitor):
```
david-yz-liu: This class is not useful. Please remove it can just call `register_transform` inline.

ending_location_visitor.py
```diff
@@ -1,32 +1,223 @@
...
+        with open(file_location) as f:
+            content = f.read()
+        return self._get_string_as_module(content)
+
+    def _get_string_as_module(self, string):
+        """Parse the string with astroid, and return the module.
+        """
+        return astroid.parse(string)
+
+    def _are_equal(self, expected, props_assigned):
```
david-yz-liu: You don't need to define this. There's an `assertListEqual` method you can use.

ending_location_visitor.py
```diff
@@ -1,32 +1,223 @@
...
+
+    # TODO: Many more test functions here...
+
+
+
+
+
+if __name__ == '__main__':
+    # run tests
+    # unittest.main()
```
david-yz-liu: I'm not sure why `unittest.main()` isn't being used here.

setendings.py
```diff
@@ -0,0 +1,624 @@
...
+
+
+
+
+    
+
+
+    # Temporary:    
+    # print('props: ', dir(node))
+    last_child_end = node.last_child().end_col_offset
```
david-yz-liu: This is indeed the approach I suggested. You should remove the "as_string" stuff above.

Use this approach for setting the line number as well.

ending_location_visitor.py
nigef: assertListEqual seems to care about order of the items, which i would like to try to avoid.

setendings.py
nigef: ahh sorry, i like them. i will stop

setendings.py
```diff
@@ -0,0 +1,397 @@
...
+        raise Exception('''ERROR:️ lineno is None at node:''', node, 
+                        node.as_string())
+    else:
+        node.end_lineno = node.tolineno
+
+def _set_end_col_offset_by_string(node):
+    """Nodes without children can get end_col_offset by length of string.
+    Hopefully this works without problems.
+    """
+    NO_CHILDREN_TYPE = [astroid.Arguments, astroid.AssignName, astroid.Break,
```
david-yz-liu: You can move this into the global scope. I imagine eventually (but not now) you'll want to call `register_transform` on each of these. Although I do wonder about some of these, and would want to investigate further before doing so.

setendings.py
```diff
@@ -0,0 +1,397 @@
...
+import logging
+
+# Set the log level (DEBUG, ERROR, ...), and message format.
+logging.basicConfig(format='', level=logging.DEBUG)
+
+# Helper functions to abstract common tasks. Global for now..
+# these functions are associated to types in the transforms dictionary in the
+# TransformVisitor class. The _transform method invokes the transform function
+# on the nodes.
+# TODO: Separate into Class? .. would need static methods.
```
david-yz-liu: You don't need a class. You can remove this TODO.

setendings.py
```diff
@@ -0,0 +1,397 @@
...
+
+    # Compare with "None" since value could be "0", which is falsey.
+    # Returning a.s.a.p. reduces the runtime.
+    # Some nodes dont have the attribute set yet.
+    if hasattr(node, 'end_lineno') and hasattr(node, 'end_col_offset'):
+        if node.end_lineno is not None and node.end_col_offset is not None:
+            return  # since properties are already set.
+
+    # Doing this first makes it a postorder traversal.
+    for child in node.get_children():  # get_children() returns a generator.
```
david-yz-liu: Please remove this.

The postorder traversal happens "automatically" when the TransformVisitor does its thing.

Of course, it will only call the function on the nodes for which it has been registered. I hope I made clear today that the only nodes in the AST that I care about right now are `Const` and `BinOp`, which are the ones in your tests anyway (so they should still pass).

setendings.py
```diff
@@ -0,0 +1,397 @@
...
+        _set_end_col_offset(child)
+
+    # if possible, set the end_col_offset by that of the last child
+    if node.last_child():
+        _set_by_last_child(node, node.last_child())
+    # Arrgghh astroid.Arguments cannot rely on last_child() being set properly.
+    elif type(node) is astroid.Arguments:
```
david-yz-liu: This is a reminder to please move this into a separate helper function.

setendings.py
```diff
@@ -0,0 +1,397 @@
...
+        last_child = None
+        for last_child in node.get_children():  # get_children() returns a generator
+            pass  # skip to last
+        print('last item in Arguments:', last_child)
+        _set_by_last_child(node, last_child)
+    # No children. Postorder traversal sets these children first.
+    else:
```
david-yz-liu: This, too, can be split into a separate helper function.

In general, you can `register_transform` this function to the nodes which are guaranteed to have a `last_child` method (right now, just `BinOp`), and `register_transform` the `_set_end_col_offset_by_string` function to the ones that don't (right now, just `Const`).

setendings.py
```diff
@@ -0,0 +1,397 @@
...
+        _set_end_col_offset_by_string(node)
+
+def _set_by_last_child(node, last_child_node):
+    """Set end_col_offset by the value of the node's last child.
+    """
+    if hasattr(last_child_node, 'end_col_offset'):
+        if last_child_node.end_col_offset is None:  # this shoulnt happen
+            raise Exception("ERROR:️ last child's end_col_offset property is None.")
+        node.end_col_offset = last_child_node.end_col_offset
+    # Postorder sets children first!
```
david-yz-liu: This is a correct assertion. If you're trying to "fail fast" then the `else` block should also raise an exception.

setendings.py
```diff
@@ -0,0 +1,397 @@
+"""Helper module for ending_location_visitor.py
+"""
+from sys import exit
```
david-yz-liu: Please remove this import.

ending_location_visitor.py
```diff
@@ -1,32 +1,259 @@
...
+        """
+        try:
+            self.assertTrue(self._are_equal(expected, props))
+        except AssertionError as e:
+            logging.error('Compare:\n{}\nprops: {}'.format('-'*70, props))
+            logging.error('expected: {}'.format(expected))
+            logging.error('[lineno, end_lineno, col_offset, end_col_offset]')
+            raise e  # otherwise, test will always 'pass'
+
+    def _all_props_set(self):
```
david-yz-liu: Please remove this for now. Once more types are supported, we can add it back in.

ending_location_visitor.py
david-yz-liu: The order specified in the expected output should match the postorder traversal done by the Visitor.
So, order should matter to these tests.

However, it appears that `assertEqual` uses `assertListEqual` when given two lists, so you can just use `assertEqual`.

ending_location_visitor.py
```diff
@@ -1,32 +1,259 @@
...
+        Note we cannot use a set because this would hide the inequality from
+        repeated items.
+        TODO: check for equality without respect to order of items in top-level.
+        For example, the following lists are 'equal' for our purposes.
+        [[1, 1, 0, 6], [2, 2, 4, 6]]
+        [[2, 2, 4, 6], [1, 1, 0, 6]]
+        """
+        return expected == props_assigned
+
+    def assertSameness(self, expected, props):
```
david-yz-liu: If you change this to `assertEqual`, you will get both sides in the unittest output.

ending_location_visitor.py
```diff
@@ -1,32 +1,259 @@
...
+        self.ending_visitor = EndingVisitor()
+
+        # Register all `transform(node)` functions here...
+        # TODO: attach more nodes with their transform functions here.
+        self.ending_transformer.register_transform(astroid.Arguments, set_general)
+
+        self.ending_transformer.register_transform(astroid.Const, set_general)
+        self.ending_transformer.register_transform(astroid.BinOp, set_general)
+
+
```
david-yz-liu: Please remove this empty space.

ending_location_visitor.py
```diff
@@ -1,32 +1,259 @@
...
+        self.ending_transformer = TransformVisitor()
+
+        # A visitor to test the nodes property correctness.
+        self.ending_visitor = EndingVisitor()
+
+        # Register all `transform(node)` functions here...
+        # TODO: attach more nodes with their transform functions here.
+        self.ending_transformer.register_transform(astroid.Arguments, set_general)
+
+        self.ending_transformer.register_transform(astroid.Const, set_general)
```
david-yz-liu: Please create a separate `register_ending_setters` (or some other name) function in the `setendings.py`, and call it here on the `self.ending_transformer`.

In essence, the creation of the "ending transformer" should be wholly managed from the `setendings.py` module. The latter is in fact is a standalone module, not a helper for this one -- rather, this module is a test module for `setendings.py`.

In a related point, please move this module to inside the `test` folder.

ending_location_visitor.py
```diff
@@ -1,32 +1,259 @@
...
+        Register `transform(node)` function to be applied on the given
+        astroid's `node_class`. Internally, appends the function to a
+        dictionary in the TransformVisitor class. The keys are the node types,
+        each node must have been assigned/registered a transform function!!
+        The transform function may return a value which is then used to
+        substitute the original node in the tree.
+        params: node_class, transform function
+
+    We store the correct values as: 
+        [lineno, end_lineno, col_offset, end_col_offset]
```
david-yz-liu: Please change this data into a tuple rather than a list.

ending_location_visitor.py
```diff
@@ -1,32 +1,259 @@
...
+    def setUpClass(self):
+        """A class method called before tests in an individual class run.
+        setUpClass is called with the class as the only argument and must be
+        decorated as a classmethod():"""
+        
+        # A visitor to transform the nodes.
+        self.ending_transformer = TransformVisitor()
+
+        # A visitor to test the nodes property correctness.
+        self.ending_visitor = EndingVisitor()
```
david-yz-liu: You don't need a separate class here. A more elegant approach is to use the `nodes_of_class` Node method (https://github.com/PyCQA/astroid/blob/master/astroid/node_classes.py#L407). I believe @hayleyZHL did this, so you can ask her about it as well.

In general, it seems like trying to get a `TransformVisitor` to return some function of all the nodes is not idiomatic.
I did look for a bit to try to get a generator on all the nodes, and didn't find it in astroid, and it might be handy to create just that at a later point, using the approach in `nodes_of_class`.

setendings.py
nigef: Nice catch. The traversal in [line 83](https://github.com/PyCQA/astroid/blob/9d94babb2e968c5a343218a48684a9529cbab568/astroid/transforms.py#L83) eventually leads to the transform called on each node at [line 36](https://github.com/PyCQA/astroid/blob/9d94babb2e968c5a343218a48684a9529cbab568/astroid/transforms.py#L36). Noted for (my) future reference.

setendings.py
nigef: I'm not sure, since the postorder traversal -- sets the children first here.

setendings.py
```diff
@@ -0,0 +1,169 @@
...
+    if hasattr(last_child, 'end_col_offset'):  # Set by last child
+        assert last_child.end_col_offset is not None, '''ERROR:️ last child 
+            end_col_offset should not be None if used to set others. 
+            Node {}.'''.format(last_child)
+        node.end_col_offset = last_child.end_col_offset
+    # Set by string of node or last_child. May not have child.
+    else:
+        # Recall: children set first in postorder traversal.
+        # Should this always set by string, is there another way?
+        set_end_col_offset_by_string(node, last_child)
```
david-yz-liu: If the `last_child` has no `end_col_offset` attribute set, then this node has no business having its own `end_col_offset` set.

You can think of what you're doing as building up correct `end_col_offset` from the leaves upwards in an AST. No "holes" are permitted. Of course, this is why I had you start with `Const` and `BinOp`, because those form a fully self-contained recursive structure in which each node can be assured to have `end_col_offset` set.

So it is actually more appropriate for you to have an `assert` statement enforcing this condition, rather than calling `set_end_col_offset_by_string`.

setendings.py
```diff
@@ -0,0 +1,169 @@
...
+        return skip_to_last_child  # node, or None.
+
+# Define the functions to transform the nodes.
+# These functions are registered to their respective node type with the
+# register_transform function.
+# Mutates the nodes, adding the properties: `end_lineno`, `end_col_offset`.
+def set_general(node):
+    """General function called by many of the other transform functions.
+    Populate ending locations for astroid node."""
+    last_child = _get_last_child(node)
```
david-yz-liu: You can make it clearer what the real purpose of this function is by doing the following:

1. Change its name to `set_from_last_child`
2. Change the docstring to have as a precondition that `node` must have a last child accessible by `_get_last_child`
3. Put the assert statements here (e.g., `assert last_child is not None`, `assert hasattr(last_child, end_lineno)`, etc.

Similar comments apply to `set_without_children`.

setendings.py
```diff
@@ -0,0 +1,169 @@
...
+    set_end_lineno(node, last_child)
+    set_end_col_offset(node, last_child)
+
+def set_without_children(node):
+    """Populate ending locations for nodes that are guaranteed to never have
+    children. E.g. Const.
+    """
+    last_child = _get_last_child(node)
+    set_end_lineno(node, last_child)
+    set_end_col_offset_by_string(node, last_child)
```
david-yz-liu: Based on my comment above, you should change the signature of `set_end_col_offset_by_string` to only take in the node; last_child should always be `None` here.

tests/test_setendings.py
```diff
@@ -0,0 +1,171 @@
...
+        self.reset()
+
+    def reset(self):
+        """Reset between test methods.
+        """
+        # Keep a list rather than a set, because "too many" is wrong.
+        self._props_check = []
+
+    def check_endings(self, module, node_class):
+        for child_node in module.body:
```
david-yz-liu: Maybe I'm missing something here. Is it not possible to call `nodes_of_class` on the `module` already?

setendings.py
nigef: sometimes `set_end_col_offset_by_string` will be called here when the node is the leaf, so it will not have the `end_col_offset` property set yet.

setendings.py
nigef: I'm not sure how the leaf nodes would get set then?

nigef:
Transform functions are registered for all known nodes.
Some inconsistencies/etc worth looking into are mentioned for some node tests in the unit tests, with "Note: ..."

PR #89, opened 2016-07-25
Forbidden Import Checker Options

shweta97:
Modified the forbidden_import_checker to allow certain modules to be imported through setting options in the .pylintrc file


checkers/forbidden_import_checker.py
```diff
@@ -8,22 +8,35 @@
 class ForbiddenImportChecker(BaseChecker):
     __implements__ = IAstroidChecker
 
-    name = 'custom'
+    name = 'forbiddenimport'
```
david-yz-liu: As we discussed in the meeting, it would be good to put a space here.

david-yz-liu:
@shweta97 there are a couple of commented-out sections you need to remove.

Also, it would be good to add a *commented-out* section in the .pylintrc file which illustrates how to set this new option.

checkers/forbidden_import_checker.py
```diff
@@ -8,22 +8,30 @@
...
     priority = -1
 
     @check_messages("forbidden-import")
     def visit_import(self, node):
         """visit and Import node"""
-        self.add_message('forbidden-import',
-                         node=node,
-                         args=(', '.join(map(lambda x: x[0], node.names)),
-                               node.lineno))
+        temp = []
```
david-yz-liu: You can use a list comprehension to define this as well. (Remember that you can use `if`s in list comprehensions too.

david-yz-liu:
@shweta97 Nice work. I left one more comment.

But also in the .pylintrc file, you should have a commented out actual sample usage of the option.
I.e., a line that the user should be able to uncomment and have do something.
The message you left doesn't actually help the user know *how* to configure the checker.

.pylintrc
```diff
@@ -9,6 +9,11 @@ max-nested-blocks = 2
 # Set the maximum line length. The maximum line length in pep8 is 80 characters.
 max-line-length = 79
 
+[FORBIDDEN IMPORT]
+
+# Set the list of modules that are allowed to be imported
+# For example: allowed_import_modules=pyta
```
david-yz-liu: @shweta97 reminder to please use hyphens rather than underscores.

checkers/forbidden_import_checker.py
```diff
@@ -3,27 +3,36 @@
...
 from pylint.checkers import BaseChecker
 from pylint.checkers.utils import check_messages
+from colorama import Fore
 
 
 class ForbiddenImportChecker(BaseChecker):
     __implements__ = IAstroidChecker
 
-    name = 'custom'
+    # \033[4;34m%s\033[0m
```
david-yz-liu: You can delete this.

checkers/forbidden_import_checker.py
```diff
@@ -3,27 +3,36 @@
...
     def visit_import(self, node):
         """visit and Import node"""
-        self.add_message('forbidden-import',
-                         node=node,
-                         args=(', '.join(map(lambda x: x[0], node.names)),
-                               node.lineno))
+        temp = [name for name in node.names if name[0] not in self.config.allowed_import_modules]
+
+        if temp != []:
+            self.add_message('forbidden-import', node=node, args=(', '.join(map(lambda x: x[0], temp)), node.lineno))
```
david-yz-liu: This line is a little long; please split it up across multiple lines.

david-yz-liu:
@shweta97 FYI the option actually did work when hyphenated; the hyphens are automatically converted into underscores for the attribute name.

PR #98, opened 2016-08-27
Typecheck: Type Constraints

hayleyZHL:
Made sure it's on the right branch and keep the pull request clean (even though there are a lot of commits from my master branch, I reverted the previous commits so that only 2 necessary files are changed.)


david-yz-liu:
@hayleyZHL thanks. I think you'll need to do pull from `upstream typecheck` now that I've updated this branch.

PR #115, opened 2016-09-11
Ending locations

nigef:
Added logging. 
Stylistic adjustments in client. 
Expanded some tests. 
Found Module node.


python_ta/transforms/setendings_log.log
```diff
@@ -0,0 +1 @@
+2016-09-08 20:12:37 INFO fix_start_attributes Arguments,Comprehension,ExtSlice,Index,Keyword,Module,Slice
```
david-yz-liu: This file should not be committed.

python_ta/transforms/setendings.py
```diff
@@ -225,7 +264,16 @@ def set_arguments(node):
     """
     if _get_last_child(node):
         set_from_last_child(node)
-    else:
+    else:  # node does not have children.
+        # set from col offset of parent node, plus len of name, etc.
+        # Note: if there are no arguments, this node takes up no space
+        parent_node = node.parent
+        if isinstance(parent_node, astroid.FunctionDef):
```
david-yz-liu: These changes still suffer from whitespace issues. This means you should revert these changes, and the problems should be handled by examining the source code.

python_ta/transforms/setendings.py
```diff
@@ -65,13 +66,15 @@
     astroid.DelAttr,
     astroid.Delete,
     # TODO: missing right }
+    # [This one is tricky because there is no way to capture the last brace location]
```
david-yz-liu: This comment and the one below should be removed, now that we have adopted the approach of using the source code to fix them. In fact, all of the TODOs should be handled in this way, so putting in these comments in only two places is kind of misleading.

python_ta/transforms/setendings.py
```diff
@@ -109,6 +113,38 @@
 ]
 
 
+class NodeDataStore():
+    """Collect data and log at end of all tests."""
+    def __init__(self):
+        """Store data without dupes."""
+        self._storage = set()
+
+        # Question: do we want to move logger to its own module, in a utils directory?
```
david-yz-liu: I do not believe a class, nor a global variable, is necessary at all. It is fine to configure `logging` inside the main block of both the test file and the client.

python_ta/transforms/setendings.py
```diff
@@ -146,14 +181,31 @@ def init_register_ending_setters():
...
+
+
 def fix_start_attributes(node):
     """Some nodes don't always have the `col_offset` property set by Astroid:
-    astroid.Comprehension, astroid.ExtSlice, astroid.Index,
-    astroid.Keyword, astroid.Module, astroid.Slice
+    Comprehension, ExtSlice, Index, Keyword, Module, Slice.
 
     Question: is the 'fromlineno' attribute always set?
+        ==> preliminary answer is, yes.
```
david-yz-liu: Okay, that's good. You can remove lines 200 and 201.

sample_usage/print_nodes.py
```diff
@@ -21,40 +31,62 @@ def print_node(filename, node_class):
...
+
+            if first_line:  # Add a spacing after first_line
+                middle_lines = '\n' + middle_lines
+
+            out = [
+                Style.DIM + 
+                first_line[:node.col_offset] +
+                
+                # The key part:
+                MAIN_COL +
```
david-yz-liu: It is a good idea to have a helper function for the colouring like the one you created. I think you can use it here, and possibly other places too (with perhaps other modification).

sample_usage/print_nodes.py
```diff
@@ -21,40 +31,62 @@ def print_node(filename, node_class):
...
-            out = (
-                [Style.DIM +
-                 first_line[:node.col_offset] +
-                 Style.BRIGHT + Fore.LIGHTBLACK_EX + Back.BLACK +
-                 first_line[node.col_offset:]] +
+            first_line = source_lines[node.fromlineno - 1]  # string
+            middle_lines = source_lines[node.fromlineno: node.end_lineno - 1]  # list
+            last_line = source_lines[node.end_lineno - 1]  # string
+
+            if middle_lines:
```
david-yz-liu: Do not convert `middle_lines` into a single string here. It *is* a good idea to call map though, to get the colouring right:

```python
middle_lines = list(map(wrap_color, middle_lines))
```

python_ta/transforms/setendings.py
nigef: Should the changes be reverted in the git command sense -- to clean up commit history (challenging), or by simply deleting the changes manually?

david-yz-liu: You can do either, but I'd prefer you do the simpler one for now (since cleaning up the commit history isn't a high priority for me).

PR #116, opened 2016-09-21
Node docs last third

IskyN:
Documented the first 6 of my 20 astroid nodes to document. (The Dict node is to be documented by all 3 ROPes, but my version should not differ from the others'.)
Documented nodes:
- Dict (test node, documented by everyone)
- List
- ListComp
- Name
- Nonlocal
- Pass

Edit 2016-10-02: Now documented, with better template, until Slice.py.

### DIFF
--- a/nodes/List.py
+++ b/nodes/List.py
@@ -18,7 +18,7 @@ Example 2:
     - ctx   -> Load
 
 Example 3:
-    - elts  -> [Num(1), Num(2)]
+    - elts  -> [Name('x', Store()), Name('y', Store())]
     - ctx   -> Store
 """
 
--- a/nodes/List.py
+++ b/nodes/List.py
@@ -16,6 +16,10 @@ Example 1:
 Example 2:
     - elts  -> [Num(1), Num(2), Num(3)]
     - ctx   -> Load
+
+Example 3:
+    - elts  -> [Num(1), Num(2)]
+    - ctx   -> Store
 """
 
 # Example 1
@@ -23,3 +27,6 @@ Example 2:
 
 # Example 2
 [1, 2, 3]
+
+# Example 3
+[x, y] = 7, 8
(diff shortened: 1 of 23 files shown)


nodes/SetComp.py
```diff
@@ -1 +1,30 @@
...
+
+This node represents set comprehension expressions in Python, which take the
+form {element for increment in iterable}. This node also makes use of the
+Comprehension node.
+
+Attributes:
+    - elt         (Expr)
+        - The part to be evaluated to make every item in the set. This can
+          be any expression node, such as a BinOp or Str.
+    - generators  (List[])
```
david-yz-liu: In general for the `List` type, either you don't specify the type of the contents and just write `List`, or you *do* specify the type of the contents, e.g. `List[Comprehension]`.

nodes/Raise.py
```diff
@@ -1 +1,47 @@
-raise ZeroDivisionError()
+"""
+Raise astroid node
+
+This node represents the "raise" statement in Python, which is used to raise
+exceptions. This statement also handles exception chaining using the "from"
+clause, which is included in the Raise node.
+
+Attributes:
+    - exc    (node)
```
david-yz-liu: Make sure to capitalize `Node` when referring to it as a class name. Also, express the fact that this attribute could be none in its type: `Node | None`

nodes/List.py
```diff
@@ -1 +1,25 @@
+"""
+List astroid node
```
david-yz-liu: We talked about trying to get a "Store" context for lists. I wonder if this is achieved by the assignment statement

```python
[x, y] = [1, 2]
```

nodes/Raise.py
IskyN: Got it, thanks.

nodes/List.py
IskyN: I checked, and yes it is! Thank you for that :)

nodes/SetComp.py
IskyN: Oops, I just got distracted before finishing that one, thanks for pointing it out.

PR #117, opened 2016-09-21
Added documentations

yunyiwendy:
Documentations are added for the following nodes:
- Decorators.py
- DelAttr.py
- Delete.py
- DelName.py
- Delete.py
- Dict.py
- DictComp.py
- Ellipsis.py
- ExceptHandler.py
- Expr.py
- ExtSlice.py
- Lambda.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+<p align="center">
+    <img src="https://github.com/pyta-uoft/website/blob/master/images/pyta_logo_markdown.png?raw=true" alt="pyta logo" width="120">
+</p>
+
 # PyTA
 
 PyTA is a Python program which uses static code analysis to help students find


nodes/ExceptHandler.py
```diff
@@ -1,19 +1,35 @@
...
+Example 2:
+    - type  -> ValueError
+    - name  -> None
+    - body  -> [print("NOOOOOO!!!!!!!!!")]
 """
 
+#Example 1
 try:
     pass
 except:
```
nigef: it looks like line 33, 35 distracts from the essence of this example -- why not just one example with `except ValueError as e` on line 28?

nodes/Delete.py
```diff
@@ -1,9 +1,12 @@
...
 Delete astroid node
 
+Delete node represents a del statement.
+
 Attributes:
-    - targets  (list)  the targets to be deleted of type Name, Attributes, or
-                       Subscript
+    - targets  (List[Node])
+        - The targets to be deleted of type Name, Attributes, or Subscript
+          nodes.
```
nigef: In general, I'm a supporter of few words, like
> The target nodes to be deleted: Name, Attribute, or Subscript.

nodes/ExceptHandler.py
yunyiwendy: Hey Nigel thanks for commenting!
I wasn't sure if I can change the original example and we decided last week that we'll add another example to help beginners understand the purpose of this node. Btw this is not final draft (I committed these changes because I was practicing using git :P)

nodes/Decorators.py
```diff
@@ -1,3 +1,31 @@
+"""
+Decorators astroid node
```
david-yz-liu: Just a reminder that we discussed this node in depth at the meeting.

nodes/DelAttr.py
```diff
@@ -1,3 +1,21 @@
+"""
+DelAttr astroid node
+
+DelAttr deletes the named attribute, provided the object allows it. For example,
+delattr(x, 'foobar') is equivalent to del x.foobar.
+
+Attributes:
+    - object  (object)
```
david-yz-liu: I wonder if `object` here is really an instance of any Python class, or whether it's supposed to be a particular type of astroid Node.

nodes/DelName.py
```diff
@@ -1 +1,12 @@
+"""
+DelName astroid node
+
+Attributes:
+    - targets  (List[Node])
```
david-yz-liu: There might be the same kind of confusion here as with Decorators vs. FunctionDef.

It seems like you're saying both `Delete` and `DelName` nodes have an attribute called `target`, but this seems unlikely. (My intuition would be that `Delete` has the `target` attribute, which is a list of `DelName` and/or `DelAttr` nodes. I could be wrong, though.)

nodes/DictComp.py
```diff
@@ -1 +1,20 @@
+"""
+DictComp astroid node
+
+A DictComp node represents dictionary comprehensions in Python.
+
+Attributes:
+    - key         (node)
```
david-yz-liu: Make sure you capitalize `Node` when referring to it as a type.

nodes/DictComp.py
```diff
@@ -1 +1,20 @@
...
+        - The part that will be evaluated for each item.
```
david-yz-liu: This description shouldn't be the same as the one for "value" - there is a difference that should be documented..

nodes/DictComp.py
```diff
@@ -1 +1,20 @@
...
+    - value       (node)
+        - The part that will be evaluated for each item.
+    - generators  (List[Node])
```
david-yz-liu: If `generators` must be a list of Comprehension nodes, the proper type is `List[Comprehension]`

nodes/ExtSlice.py
```diff
@@ -1,6 +1,21 @@
+"""
+ExtSlice astroid node
+
+Extended slicing which most python object like list and string don't support.
+
+Attributes:
+    - dims  (List[Node])
```
david-yz-liu: Again, for the type to be consistent with the description, you'll probably want to write `List[Slice | Index]`

nodes/If.py
```diff
@@ -1,2 +1,24 @@
...
+
+An if statement.
+
+Attributes:
+    - test    (Node)
+        - Holds a single node such as Compare.
+    - Body    (List[Node])
+        - A list of nodes that will execute if the condition passes.
+    - orelse  (List[Node])
+        - The else clause.
```
david-yz-liu: This one is worth expanding on to incorporate an example with `elif`. Such clauses are represented in an interesting way in astroid.

nodes/Lambda.py
```diff
@@ -1 +1,17 @@
...
+"""
+Lambda astroid node
+
+Lambda is a minimal function definition that can be used inside an expression.
+
+Attributes:
+    - args  (list)
+        - The arguments for function lambda.
+    - body  (node)
+        - The body of function lambda.
```
david-yz-liu: It would be good to highlight here that the body is a single node, not a list of nodes; this is pretty interesting.

PR #119, opened 2016-09-22
Ast Annotations

rebeccakay:
First check to see if everything's ok
Arguments.py
Assert.py
Assign.py
Comprehension.py

### DIFF
--- a/nodes/Arguments.py
+++ b/nodes/Arguments.py
@@ -1,2 +1,25 @@
+"""
+Arguments astroid node
+
+Attributes:
+    - args         (list)  list of non-keyword argument names
+    - vararg       (arg)   variable-length argument or None
+    - kwonlyargs   (list)  list of keyword-only argument names
+    - kwarg        (arg)   single arg nodes and keyword only arguments
+    - defaults     (list)  n-tuple of the default values of the last n
+                           arguments, or None if no default arguments
+    - kw_defaults  (list)  list of default values for keyword-only arguments,
+                           but if one is None, the corresponding argument is
+                           required
+
+Example:
+    - args         -> arg
+    - vararg       -> None
+    - kwonlyargs   -> None
+    - kwarg        -> None
+    - defaults     -> None
+    - kw_defaults  -> None
+"""
+
 def fun(arg):
     pass
--- a/nodes/Arguments.py
+++ b/nodes/Arguments.py
@@ -11,7 +11,7 @@ Attributes:
           positionally.
     - kwonlyargs   (List[arg])
         - A list of keyword-only argument names.
-    - kw_defaults  (List[None|arg])
+    - kw_defaults  (List[None | arg])
         - A list of default values for keyword-only arguments.
     - vararg       (arg|None)
         - A variable-length argument.
@@ -27,6 +27,7 @@ Example:
     - kw_defaults  -> [None,Num(n=3)]
     - vararg       -> arg(arg='g', annotation=None)
     - kwarg        -> arg(arg='g', annotation=None)
+    - defaults     -> [Num(n=1),Num(n=2)]
 """
 
 def f(a: 'annotation', b=1, c=2, *d, e, f=3, **g):
--- a/nodes/Arguments.py
+++ b/nodes/Arguments.py
@@ -6,12 +6,12 @@ The arguments for a function.
 Attributes:
     - args         (List[arg])
         - A list of non-keyword argument names.
-    - defaults     (Tuple|None)
-        - N-tuple of the default values of the last n arguments, or None if no
-          default arguments.
+    - defaults     (List[Node])
+        - A list of default values for arguments that can be passed
+          positionally.
     - kwonlyargs   (List[arg])
         - A list of keyword-only argument names.
-    - kw_defaults  (List[arg|None])
+    - kw_defaults  (List[None|arg])
         - A list of default values for keyword-only arguments.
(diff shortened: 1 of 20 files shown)


nodes/Arguments.py
```diff
@@ -1,2 +1,25 @@
+"""
+Arguments astroid node
+
+Attributes:
+    - args         (list)  list of non-keyword argument names
```
david-yz-liu: As we discussed last week, use `List` instead of `list`, and in fact if you can be more specific with the node types, please do so. E.g., `List[Name]` (if that's the right node type).

nodes/Arguments.py
```diff
@@ -1,2 +1,25 @@
...
+    - vararg       (arg)   variable-length argument or None
+    - kwonlyargs   (list)  list of keyword-only argument names
+    - kwarg        (arg)   single arg nodes and keyword only arguments
+    - defaults     (list)  n-tuple of the default values of the last n
```
david-yz-liu: Use the type `Tuple` to differentiate from `List`; also, if a value can be `None`, indicate this with a type like `Tuple | None`

nodes/Arguments.py
```diff
@@ -1,2 +1,25 @@
...
+                           arguments, or None if no default arguments
+    - kw_defaults  (list)  list of default values for keyword-only arguments,
+                           but if one is None, the corresponding argument is
+                           required
+
+Example:
+    - args         -> arg
```
david-yz-liu: If `args` is a list, make sure to put square brackets around this.

nodes/Assert.py
```diff
@@ -1 +1,13 @@
+"""
+Assert astroid node
+
+Attributes:
+    - test  (compare)  the condition to be evaluated True or False
```
david-yz-liu: I'm guessing the `test` can really be an arbitrary expression, not necessarily a comparison node. For example, if you had a helper function `def f(n): return n > 5` and then did `assert f(10)`, the `test` would be a function call expression, right?

nodes/Assign.py
```diff
@@ -1 +1,13 @@
+"""
+Assign astroid node
+
+Attributes:
+    - targets  (list)  a list of nodes
+    - value    (type)  a single node
```
david-yz-liu: This should probably be an arbitrary `Node` or at least `Expr` class, and definitely not "type" (since you're assigning a value to a name).

nodes/AsyncFor.py
```diff
@@ -1,3 +1,29 @@
...
+
+Attributes:
+    - target  (Node(Name|Tuple|List))
+        - Initial value for the iteration variable and may be any type,
+          not just a number. Equivalent to for loop's i = 0.
+    - iter    (Node)
+        - Holds the item to be looped over.
+    - body    (List[Node])
+        - Contains lists of nodes to execute.
+    - orelse  (List[Node])
```
david-yz-liu: It looks like this attribute could be `None` as well? If so, this should be reflected in the type.

nodes/AsyncFor.py
```diff
@@ -1,3 +1,29 @@
+"""
+AsyncFor astroid node
+
+This node iterates over async code with a for-loop like syntax. Asynchronous
+code doesn't wait for an operation to complete, rather the code executes all
+operations in one go. Only valid in body of an AsyncFunctionDef astroid node.
+
+Attributes:
+    - target  (Node(Name|Tuple|List))
```
david-yz-liu: The classes `Name`, `Tuple`, and `List` are all subclasses of `Node`, so you don't need to write the "Node" explicitly in the type. Doing just `(Name | Tuple | List)` conveys the right information.

nodes/Attribute.py
```diff
@@ -1 +1,20 @@
+"""
+Attribute astroid node
+
+To give attribute access.
+
+Attributes:
+    - value  (Name|Other)
+        - Typically a Name node.
+    - attr   (Str)
```
david-yz-liu: The `str` type shouldn't be capitalized (its name is lowercase `str` in Python).

nodes/AugAssign.py
```diff
@@ -1 +1,22 @@
...
+AugAssign astroid node
+
+An augmented assignment, which is the combination, in a single statement, of a
+binary operation and an assignment statement.
+
+Attributes:
+    - target  (Node(Name|Subscript|Attribute))
+        - A list of nodes. Multiple nodes represents assigning the same value.
+          to each.
+    - op      (Node(Add))
```
david-yz-liu: I think this was pointed out in the meeting, but the `op` isn't always a `+=`, so it's misleading to say this is always an `Add` node.

nodes/Comprehension.py
```diff
@@ -1 +1,16 @@
+"""
+Comprehension astroid node
+
+Attributes:
+    - target  (class)  typically a name or tuple node; the reference to use for
```
david-yz-liu: This should probably be updated so the type is some kind of node, rather than "class."

nodes/AssignAttr.py
```diff
@@ -1,3 +1,19 @@
...
+
+Attributes:
+    - expr      (Expr)
+        - An expression to be assigned.
+    - attrname  (Node)
+        - The name that is assigned to the expression.
+
+Example:
+    - expr      -> 'name'
+    - attrname  -> 'self.name'
```
rebeccakay: unsure about the examples

PR #123, opened 2016-09-29
typecheck: Type Constraints

hayleyZHL:
BinOp Components: 
- Supports List/Tuple/String Concatenation using `+` or `*` operators.
- Operators includes `+,  -,  *,  /,  **,  //,  %`.
- Raise `ValueError` exception when invalid operations are detected, for instance, `'abc' * 'sss`, `[1, 2] * [3, 4]`.

Other Node Types:
- Compare Node
- Subscript Node
- List, Tuple, Dict Node
- Const Node
- UnaryOp Node

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -46,25 +46,65 @@ def set_binop_type_constraints(node):
     left_type = node.left.type_constraints
     right_type = node.right.type_constraints
 
-    if left_type == right_type:
+    # '*' does not work for same type of operands such as str, list... etc
+    if node.op != '*' and left_type == right_type:
         node.type_constraints = node.left.type_constraints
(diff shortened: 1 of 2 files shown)


david-yz-liu:
To follow up from the meeting's discussion: I looked into the source code for the typing module (https://hg.python.org/cpython/file/3.5/Lib/typing.py). It turns out that the implementation of these values are quite complex. However, you should be able to use special attributes (marked with double underscores) to access data. E.g., for lists:

```python
>>> list_type = List[int]
>>> list_type.__origin__
typing.List  # has some other stuff you can ignore.
>>> list_type.__origin__ == List
True
>>> list_type.__args__
(<class 'int'>,)
```

Use `dir` on these type values to see what attributes they have, and do some investigation to see how to get them.

david-yz-liu:
@hayleyZHL great, nice work.

PR #125, opened 2016-10-02
Node docs

yunyiwendy:
Refined all the nodes from Decorators.py to Lambda.py.
Moved all the previous commits from master branch to this node-docs branch.
Please ignored the "18 nodes refined" pull request.

### DIFF
--- a/nodes/For.py
+++ b/nodes/For.py
@@ -1,2 +1,25 @@
+"""
+For astroid node
+
+A for loop.
+
+Attributes:
+    - target  (Node(Name|Tuple|List))
+        - Holds the variable(s) the loop assigns to as a single node.
+    - iter    (Node)
+        - Holds the item to be looped over.
+    - body    (List[Node])
+        - Contains lists of nodes to execute.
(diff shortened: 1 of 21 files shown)


nodes/For.py
```diff
@@ -1,2 +1,25 @@
...
+For astroid node
+
+A for loop.
+
+Attributes:
+    - target  (Node(Name | Tuple | List))
+        - Holds the variable(s) the loop assigns to as a single node.
+    - iter    (Node)
+        - The node to be looped over.
+    - body    (Node)
```
IskyN: Shouldn't the type for body be `List[Stmt]`?

yunyiwendy: I'm quite confused about how to know the type of the attribute by looking at the source code. But I think if the type of the attribute is a list, it will be presented in a source code like self.items = [] in the __init__ function of that node, like the one in dict. So I assume that self.body = body just means that body is a Node type instead of a list??? Also I don't know what Stmt is Q__Q.

PR #128, opened 2016-10-13
PyTA Statistics Package

IskyN:
- [x] Fix `is` problem
- [x] Change lists of tuples to OrderedDicts
- [x] Isolate and fix minor errors
- [x] Fix multi-folder directory problem
- [x] Add completed README
- [ ] Perform comprehensive tests to make sure correct stats are calculated & printed

This is a mostly correct statistics aggregation and analysis module for Python TA. (See also the new `StatReporter` object that is used by the `sample_usage` module.)

Contributors: @yunyiwendy, @rebeccakay, @iskyn


yunyiwendy:
Debug note: 

- python_ta.check_all() doesn't recognize the whole path, instead it recognizes module names, so I changed line 30 to this:
  >>> python_ta.check_all(os.path.join(student_id, file), reporter=StatReporter)
So now when I pass in a directory named examples, the module name that got passed in will be: 
  >>> python_ta.check_all('examples.forbidden_import_example', reporter=StatReporter) 
But the problem is that students' utorid won't be recognized as a module, unless if it is inside the pyta directory by itself (It also won't be recognized as a module if it is a subfolder of a folder that's in pyta directory).

sample_usage/pyta_stats.py
```diff
@@ -0,0 +1,102 @@
...
+    all_errors = {}
+    for root_str, dir_list, file_list in os.walk(directory):
+        # check if directory is student directory or assignment parent directory
+        if not dir_list:
+            # It is a student folder
+            student_errors = []
+            student_style = []
+            for file in file_list:
+                # check if file is a .py file
+                if file[-3:] == ".py":
```
david-yz-liu: use `endswith`

sample_usage/pyta_stats.py
```diff
@@ -0,0 +1,102 @@
...
+                    python_ta.check_all(os.path.join(root_str, file),
+                                        reporter=StatReporter)
+                    # store all the msg objects of this student's files
+                    student_errors.extend(StatReporter.error_messages)
+                    student_style.extend(StatReporter.style_messages)
+                    student_id = os.path.basename(root_str)
```
david-yz-liu: This line and the next should be taken out of the loop.

sample_usage/pyta_stats.py
```diff
@@ -0,0 +1,102 @@
...
+            print("\t{}: {}".format(stat_type, value))
+
+        print("~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~" * 2)
+
+
+def _print_top_errors(stats, tabs=2):
+    """
+    Pretty prints a list of most frequent errors and their counts.
+
+    @param list[tuple(str, int)] stats: the errors and counts to be printed
```
david-yz-liu: Use the typing type representations `List[Tuple[str, int]]`.

sample_usage/stats_analysis.py
```diff
@@ -0,0 +1,160 @@
+from collections import OrderedDict
+
+def indiv_calc(error_msgs, style_msgs):
```
david-yz-liu: Use an unabbreviated name for this function.

sample_usage/stats_analysis.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @param list error_msgs: Message objects for all of a student's code errors
+    @param list style_msgs: Message objects for all of a student's style issues
+    @rtype: list[tuple]
+    """
+
+    # {msg.symbol + "(" + msg.object + ")": count}
+    all_msgs = error_msgs + style_msgs
+
+    all_num = _calc_helper(all_msgs)
```
david-yz-liu: Better to use unpacking assignment here:

```python
x, y = _calc_helper(all_msgs)
```

sample_usage/stats_analysis.py
```diff
@@ -0,0 +1,160 @@
...
+        code_errors.append(errors)
+        style_errors.append(styles)
+
+    error_num = frequent_messages(_message_counter(code_errors))
+    style_num = frequent_messages(_message_counter(style_errors))
+
+    # Calculating the Five Number Summary for all errors (per student)
+    stu_errors.sort(reverse=True)
+
+    if len(stu_errors) == 1:
```
david-yz-liu: For this part, import `statistics` and use [statistics.median](https://docs.python.org/3/library/statistics.html#statistics.median) rather than computing it yourself.

sample_usage/stats_analysis.py
```diff
@@ -0,0 +1,160 @@
...
+    @rtype: list
+    """
+    # get dict of values {id:int, id2:int}
+    msgs_dict = _message_counter(msgs)
+    # sort into list of tuple, highest on top
+    freq_nums = frequent_messages(msgs_dict)
+    total_msgs = len(msgs)
+    # divide each value by total and round to two places
+    for message in msgs_dict:
+        msgs_dict[message] = (msgs_dict[message]/total_msgs * 100).__round__(2)
```
david-yz-liu: Don't call special methods directly. There's a `round` function you can use.

sample_usage/stats_analysis.py
```diff
@@ -0,0 +1,160 @@
...
+    @rtype: dict[str : int]
+    """
+
+    msgs = msgs.copy()  # prevent aliasing
+    msgs_dict = {}
+
+    for msg in msgs:
+        key = msg.msg_id + " (" + msg.symbol + ")"
+        if key not in msgs_dict:
+            msgs_dict[key] = sum(1 for m in msgs if m.msg_id == msg.msg_id)
```
david-yz-liu: Use `len` rather than `sum` here.

sample_usage/stats_analysis.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    # get key-value pair in a list
+    most_frequently = list(comp_dict.items())
+    # sort the list according to the value
+    most_frequently.sort(key=lambda p: p[1])
+    # from larger number to lower number
+    most_frequently.reverse()
+    # So the name of the error first and then the number of its occurrence.
+    # return the top whatever number
+    if isinstance(top, int) and top > 0:
```
david-yz-liu: You can remove this type check.

sample_usage/stats_analysis.py
```diff
@@ -0,0 +1,160 @@
...
+        if top > len(most_frequently):
+            top = len(most_frequently)
```
david-yz-liu: You don't actually need to do this. If a slice upper index is out of bounds, the slice will correctly end (without error) at the end of the list.

IskyN:
Re: 77faa4b:
It turns out that the problem with calling `pyta_statistics` on a `directory` with multiple student directories inside it was that `importlib` would be trying to look for modules in directories specified in the PYTHONPATH, and `directory` wasn't in the PYTHONPATH. So appending `directory` to `sys.path` was a simple fix (no temporary `__init__.py` files required).
(Actually, the [pylint docs](https://pylint.readthedocs.io/en/latest/user_guide/run.html#invoking-pylint) hint at this, but not very helpfully.)

david-yz-liu:
Note: this has been merged in.

PR #143, opened 2016-11-24
Fix single-student stats problem

IskyN:
This fixes the problem pointed out by @rebeccakay where a non-nested directory of student files would not run correctly (stats would not be calculated since the module could not be found). It was a simple (yet again) PYTHONPATH problem that has now been fixed.


david-yz-liu:
Thanks, @IskyN!

PR #145, opened 2017-01-01
Node-docs-TODO last three nodes

yunyiwendy:
I've finished fixing the documentations for the last three nodes (Import, ImportFrom, Index).
I also deleted those three nodes from TODO.md.

Side Notes: 
  One interesting thing that I found was that Import and ImportFrom don't have any attributes in the astroid fields (their attributes are in _other_fields); therefore, when I tried printing the astroid node tree for Import and ImportFrom, it showed that there weren't any attributes for those nodes. I needed to use the debugger in order to see more details about their attributes.

### DIFF
--- a/nodes/Import.py
+++ b/nodes/Import.py
@@ -1,15 +1,28 @@
 """
 Import astroid node
 
-An import statement
+Represents an import statement. Unlike ImportFrom, Import node doesn't have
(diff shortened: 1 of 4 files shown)


david-yz-liu:
Thanks, @yunyiwendy!

PR #150, opened 2017-02-22
Ending locations

nigef:
Much Progress 💪

**[High Priority]** Tricky nodes TODO still: `If, Expr, ListComp, Binop, ...`

Other TODO: write tests for the remaining nodes in `/nodes/` that are not in `/examples/ending_locations/` in order to discover other tricky edge cases, and have stronger regression tests (important in my experience).

TODO later: write tiny domain specific language to embed the expected test values within each respective file in `/examples/ending_locations/`. TBD: does comment string go at end or in beginning..


python_ta/transforms/setendings.py
```diff
@@ -191,6 +295,60 @@ def discover_nodes(node):
...
+
+        if not node.last_child():
+            if not hasattr(node, 'end_lineno'): set_without_children(node)
+
+            line_i = node.parent.fromlineno - 1  # 1-based
+            char_i = node.parent.col_offset  # 0-based
+
+            # To solve the problem created by variations of "a[:][:]",
+            # all sibling Subscript node characters to consume.
+            sibling_buffer = ''
```
david-yz-liu: This is not the correct approach. In particular, `as_string` is a poor choice. Instead, search for the first `[` after the ending position of the parent's `value` node.

python_ta/transforms/setendings.py
```diff
@@ -41,63 +63,36 @@
...
-    # TODO: missing *both* outer parens
     astroid.GeneratorExp,
+    
     # TODO: need to fix elif (start) col_offset
     astroid.If,
     astroid.IfExp,
-    # TODO: missing *both* outer brackets
     astroid.Index,
-    # TODO: would be good to see the name of the keyword as well
     astroid.Keyword,
```
nigef: This is needed here, for example when the `end_lineno` is set from its `Const` child.

david-yz-liu: What is an example of a `Keyword` node that has a `Const` child?

python_ta/transforms/setendings.py
```diff
@@ -107,24 +100,97 @@
...
+        @type node: Astroid node
+        @rtype: bool
+        """
+        return s[index : index + len(keyword)] == keyword
+    return _is_keyword
+
+def _is_within_close_bracket(s, index, node):
+    """Fix to include right ']'."""
+    # print('>>>>>>', s, s[index], index, node.end_col_offset)
+    if index >= len(s)-1: return False
```
david-yz-liu: I believe I have already asked you to stop using one-line ifs. Please change this and all the others.

nigef:
I have removed all 1-line if/elif/else statements, sorry I forgot some.

FIXME: The `_consume_subscripts` in the `fix_slice` function still uses `as_string`, since I haven't found a more robust solution that works on all test cases..

python_ta/transforms/setendings.py
nigef:
```
┃━ Expr(), name: None,  value: Call()
┃  fromlineno:        1 ... end_lineno:        1
┃  col_offset:        0 ... end_col_offset:   13
┃  >>>str(object=2)
┃  >>>str(object=2)
┃  ┃━ Call(), name: None,  value: None
┃  ┃  fromlineno:        1 ... end_lineno:        1
┃  ┃  col_offset:        0 ... end_col_offset:   13
┃  ┃  >>>str(object=2)
┃  ┃  >>>str(object=2)
┃  ┃  ┃━ Name(str), name: str,  value: None
┃  ┃  ┃  fromlineno:        1 ... end_lineno:        1
┃  ┃  ┃  col_offset:        0 ... end_col_offset:    3
┃  ┃  ┃  >>>str
┃  ┃  ┃  >>>str
┃  ┃  ┃━ Keyword(), name: None,  value: Const(int)
┃  ┃  ┃  fromlineno:        1 ... end_lineno:        1
┃  ┃  ┃  col_offset:        4 ... end_col_offset:   12
┃  ┃  ┃  >>>object=2
┃  ┃  ┃  >>>object=2
┃  ┃  ┃  ┃━ Const(int), name: int,  value: 2
┃  ┃  ┃  ┃  fromlineno:        1 ... end_lineno:        1
┃  ┃  ┃  ┃  col_offset:       11 ... end_col_offset:   12
┃  ┃  ┃  ┃  >>>2
┃  ┃  ┃  ┃  >>>2
┣━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
```

python_ta/transforms/setendings.py
nigef: Sadly (because it would have been very nice) this will not work to set Slice() in the cases where the `value` node is a `Subscript()` node. Because the ending locations of `Subscript()` nodes are set from the child nodes. This sequence should explain what I am trying to say:
-- Subscript() ending locations are set from child node,
-- Slice() is a child of Subscript(),
-- Slice() needs to be set before Subscript(),
-- therefore the ending position of the parent's `value` node (in the cases where the value node is a Subscript()) cannot be used to set the Slice() ending locations..

```
...
┃  ┃  ┃━ Subscript(), name: None,  value: Subscript()                 💥💥💥
┃  ┃  ┃  fromlineno:       15 ... end_lineno:       15
┃  ┃  ┃  col_offset:        0 ... end_col_offset:    7
┃  ┃  ┃  >>>n[:][:]    <============== cannot use end_col_offset here though :-(
┃  ┃  ┃  >>>n[:][:]
┃  ┃  ┃  ┃━ Subscript(), name: None,  value: Name(n)                 💥💥💥
┃  ┃  ┃  ┃  fromlineno:       15 ... end_lineno:       15
┃  ┃  ┃  ┃  col_offset:        0 ... end_col_offset:    4
┃  ┃  ┃  ┃  >>>n[:]
┃  ┃  ┃  ┃  >>>n[:]
┃  ┃  ┃  ┃  ┃━ Name(n), name: n,  value: None
┃  ┃  ┃  ┃  ┃  fromlineno:       15 ... end_lineno:       15
┃  ┃  ┃  ┃  ┃  col_offset:        0 ... end_col_offset:    1
┃  ┃  ┃  ┃  ┃  >>>n
┃  ┃  ┃  ┃  ┃  >>>n
┃  ┃  ┃  ┃  ┃━  Slice(), name: None,  value: None
┃  ┃  ┃  ┃  ┃  fromlineno:       15 ... end_lineno:       15
┃  ┃  ┃  ┃  ┃  col_offset:        2 ... end_col_offset:    3
┃  ┃  ┃  ┃  ┃  >>>:
┃  ┃  ┃  ┃  ┃  >>>:
┃  ┃  ┃  ┃━ Slice(), name: None,  value: None                 💥💥💥
┃  ┃  ┃  ┃  fromlineno:       15 ... end_lineno:       15
┃  ┃  ┃  ┃  col_offset:        5 ... end_col_offset:    6
┃  ┃  ┃  ┃  >>>:    <============== want to set the col_offset, end_col_offset.
┃  ┃  ┃  ┃  >>>:
...
```

nigef: Correction, this approach mentioned above actually *does* work, because all the transforms are applied to nodes as they are encountered as the AST is walked -- rather than one node type at-a-time.

PR #152, opened 2017-03-22
Update TODO list for nodes

IskyN:
I realised these changes have already been made, so this is just cleanup.


david-yz-liu:
Great, thanks @IskyN!

Issue #153, opened 2017-03-28
Highlighting for trailing-newlines needs improvement


IskyN:
As of PR #151, ColorReporter and HTMLReporter, when encountering a `trailing-newlines` style error, highlight the number of the line on which the error was made, and provide surrounding context lines in grey as usual. However, as per @david-yz-liu's suggestion, the reporters should show/highlight "*all* the trailing newlines (including the last one) to make it more obvious what the problem is." 

There are two problems with this proposition:
1. It seems that the Python `readlines` method automatically omits an end-of-file newline if there is one, so that in a file with 2 trailing newlines, the `_source_lines` will only have 1 stored; in a file with 3 trailing newlines, there will be 2 in `_source_lines`; and so on. 
_Possible workaround:_ assume there is always 1 more trailing newline than reported, and go from there.
2. In a file with more than 2 trailing newlines (so more than 1 extra), the checker finding trailing-newlines errors only identifies the last newline (really the second-last, see problem 1) as erroneous. Thus, highlighting _all_ extraneous newlines is not feasible at the reporter level. This seems to be a fix better suited to a checker or patch.

david-yz-liu:
We now display all trailing newlines that should be deleted.

PR #154, opened 2017-03-28
Make some suggested amendments to reporters and add to TODO

IskyN:
The following suggested changes were completed (numbering consistent with ROP A4 feedback page):
1. The overall formatting of the ColorReporter output takes up too much vertical space per error message. Work on reducing the number of blank lines that appear. Consider reducing indentation of text output to 2 spaces instead of 4. **(removed most blank lines, reduced indent)**
2. The heading "Your Code Starts Here" should not have every word capitalized. Moreover, I don't actually think this heading is necessary at all in either reporter type. It should be pretty clear based on formatting which parts are the student code. **(removed heading)**
3. The style of bolding in ColorReporter doesn't match the logical structure of headings. For example, the short name of the message (e.g., "trailing-newlines") is not bolded, but the actual error message itself is. **(bolded short name)**
4. Speaking of trailing-newlines, I'd recommend showing *all* the trailing newlines (including the last one) to make it more obvious what the problem is. **(not easily fixable, see issue #153)**
---
6. The use of the `_COLOURING` dictionary is excellent. However, it can be made even better by putting *all* styling information in there. For example, this same dictionary should be used to format all the headings, including the `=== Code errors ... ===` and `=== Style errors ... ===` headings. 
**(added to `_COLOURING`, removed try-except in `_colourify`)**
7. `sort_messages`:
    - the variable `count` is unnecessary **(gone)**
    - it is no longer necessary to sort the existing messages (`messages.sort`) before adding them to the dictionary **(gone)**
    - you can use a `defaultdict` class here to store the messages, with the default value being an empty list **(added to constructor, `while` loop reworked)**
8. `_colour_messages_by_type` & `_add_line`:
    - these methods are far too long, and should be split up into helper methods **(split `_colour_messages_by_type`, but splitting `_add_line` seems too inefficient)**
    - don't use single letters as arguments to `_add_line`. Instead, define an `Enum` class to represent the different types **(added, plus TODO for future)**
9. Use another class attribute to represent a line break (either `\n` or `<br/>`) **(added)**
---
11. In general, use single-quotes for string literals, not double-quotes. **(fixed)**

Left:
- [ ] Suggestion 5 (the `assigning_to_self_example` error) needs to be worked on.
- [ ] Suggestion 10 ("The source lines should be processed in the constructor to remove newline characters.") would add an extra iteration through the source lines, without any apparent gain, so this is left for now.
- [ ] Suggestions 12-20 remain.
- Some reporter TODOs added as well.

### DIFF
--- a/python_ta/reporters/TODO.md
+++ b/python_ta/reporters/TODO.md
@@ -19,10 +19,19 @@
 
 ## `HTMLReporter`
 
+* Fix the following `_COLOURING` dict entries, which need to be included so that HTMLReporter can call `_colour_messages_by_type` safely (currently useless):
+    * 'bold', 'code-heading', 'style-heading', 'code-name', 'style-name'
+
+* Use javascript to expand and close the source code of each error messages.
+
+* The ColorReporter shows the correct output for pylint.W0631_undefined_loop_variable, but the HTMLReporter doesn't.
+
+## Other
+
+* Consider changing ColorReporter & HTMLReporter's `_COLOURING` dict to an Enum.
+
 * Not all error messages are explained on the website: http://www.cs.toronto.edu/~david/pyta/:
 
     * `C0305 (trailing-newlines)`
 
-* Use javascript to expand and close the source code of each error messages.
-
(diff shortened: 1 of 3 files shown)


david-yz-liu:
Great work, thanks @IskyN!

PR #162, opened 2017-05-02
ending-locations: capture enclosing parentheses for Const nodes

david-yz-liu:
This adds a new transform method, `_add_parens`, which searches for pairs of parentheses enclosing a given node, and updates both the start and ending locations accordingly.

An exception is made for `Const` nodes that occur as the single argument of a function call.

Note: this duplicates a lot of code, requires refactoring, and does not properly handle comments. I'll probably do this soon in a separate pull request.


nigef:
Some other edge cases to handle, as far as I understand:

• we want to look up, and down a line for a match,
```
(
11
)
```

• and we want to include all parens except for the ones directly associated to the call node,
```
range(((2)))
```

david-yz-liu:
Thanks, @nigef. I'll fix the nested parentheses in a function call in a later update.

nigef:
I came back to revisit this Const node, but I was not able to figure out a solution for the following edge case, so I will leave it here as a comment. Example:
```
( 
# paren in this comment should not be "used" in Const node ending locations: (
    10    )
```

I couldn't figure out how to prevent the left paren in the comment from being paired with the close paren. What makes it difficult is we can't make assumptions on what is left or right of the particular paren. e.g. case when we would want to match the left paren is if the hash mark is in a string and therefore its not a comment.
```
...
... " ... # ... " ... (
    10    )
```
vs a case where we don't want to match the left paren, is if it is in a comment.
```
...
... # ... ( ...
    10    )
```

david-yz-liu:
@nigef this isn't worth the effort right now.

PR #164, opened 2017-05-03
Index test

RyanDJLee:
Added test cases for:
- String indexing
- Tuple indexing
- List indexing
- Dictionary indexing

### DIFF
--- a/tests/type_inference/test_literals.py
+++ b/tests/type_inference/test_literals.py
@@ -100,7 +100,7 @@ def test_string_index(index):
 @hs.composite
 def tuple_and_index(draw, elements=PRIMITIVE_VALUES):
     xs = draw(hs.tuples(elements, elements))
-    i = draw(hs.integers(min_value=0, max_value=1))
+    i = draw(hs.integers())
     return [repr(xs) + "[" + repr(i) + "]", i]
 @given(tuple_and_index())
 def test_tuple_index(index):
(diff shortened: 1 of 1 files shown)


tests/type_inference/test_literals.py
```diff
@@ -78,6 +84,58 @@ def test_heterogeneous_dict(dictionary):
     assert [Dict[Any, Any]] == result
 
 
+@hs.composite
+def string_and_index(draw):
+    xs = draw(INDEX_TYPES)
+    i = draw(hs.integers(min_value=0, max_value=len(str(xs)) - 1))
+    return ["\"" +str(xs) + "\"" + "[" + str(i) + "]", i]
```
david-yz-liu: You should use `repr` here and below, as well.

david-yz-liu: Also you can return a tuple, not a list. The subscripting below should still work.

tests/type_inference/test_literals.py
```diff
@@ -78,6 +84,58 @@ def test_heterogeneous_dict(dictionary):
...
+    """Test index nodes representing a subscript"""
+    module = _parse_text(index[0])
+    result = [n.type_constraints.type for n in module.nodes_of_class(astroid.Index)]
+    assert [type(index[1])] == result
+
+
+@hs.composite
+def tuple_and_index(draw, elements=PRIMITIVE_VALUES):
+    xs = draw(hs.tuples(elements, elements))
+    i = draw(hs.integers(min_value=0, max_value=1))
```
david-yz-liu: You don't need to specify a min or max value here.

PR #167, opened 2017-05-04
Refactor tests

RyanDJLee:
- Extract helpers and strategies and create new module for supporting hypothesis testing.
- Refactor test cases
- Need to change strategies for index testing


tests/custom_hypothesis_support.py
```diff
@@ -0,0 +1,42 @@
...
+
+INDEX_TYPES = hs.sampled_from([
+    hs.integers,
+    lambda: hs.text(alphabet="abcdefghijklmnopqrstuvwxyz", min_size=1),
+])
+INDEX_VALUES = INDEX_TYPES.flatmap(lambda s: s())
+
+TUPLE =  (hs.lists(PRIMITIVE_VALUES, min_size=1)).map(tuple)
+
+HOMO_LIST = PRIMITIVE_TYPES.flatmap(lambda s: hs.lists(s(), min_size=1))
```
david-yz-liu: Use full names for the strategies.

tests/custom_hypothesis_support.py
```diff
@@ -0,0 +1,42 @@
...
+
+HETERO_LIST = hs.lists(PRIMITIVE_VALUES, min_size=2)
```
david-yz-liu: This is actually just an arbitrary list.

tests/custom_hypothesis_support.py
```diff
@@ -0,0 +1,42 @@
+import astroid
+import hypothesis
+import hypothesis.strategies as hs
+from typing import Any, Dict, List, Tuple
+
+# Custom strategies for hypothesis testing framework
+PRIMITIVE_TYPES = hs.sampled_from([
```
david-yz-liu: Now that we are importing these, we should stick with the hypothesis style and not use all-caps. Instead, use the standard `snake_case` for naming these.

david-yz-liu:
This is on the right track. You should try to extend the list and tuple strategies to be real functions that return strategies, and that take in keyword arguments for min size (rather than hard-coding this to 1 or 2).

You should also add documentation for each of the strategies.

PR #168, opened 2017-05-04
Refactor python_ta/__init__.py to check multiple files or directories

nigef:


nigef:
This is missing recent changes from master. Will pull and redo PR..

PR #169, opened 2017-05-04
Refactor python_ta/__init__.py to check multiple files or directories

nigef:


python_ta/reporters/color_reporter.py
```diff
@@ -57,6 +57,12 @@ def print_messages(self, level='all'):
 
         print(result)
 
+        # reset messages
+        self.reset_messages()
```
david-yz-liu: Make the call to `reset_messages` in the client code, not in the body of `print_messages`. You should override `reset_messages` in this class to include the following two lines.

python_ta/__init__.py
```diff
@@ -117,42 +80,87 @@ def _check(module_name='', reporter=ColorReporter, number_of_messages=5, level='
...
+                print('Error: {}, which means the file "{}" could not be found.\n'
+                      .format(error['ex'], error['mod']))
+            else:
+                print('Error: {}\n'.format(error['ex']))
+        print()
+
+        for descr in expanded_files:
+            modname, filepath, is_arg = descr['name'], descr['path'], descr['isarg']
+            if not _find_inline_pylint_comment(filepath):
+                return
```
david-yz-liu: You can use `continue` here to check the other files.

python_ta/__init__.py
```diff
@@ -117,42 +80,87 @@ def _check(module_name='', reporter=ColorReporter, number_of_messages=5, level='
...
+        expanded_files, errors = utils.expand_modules(module_name, 
+                        linter.config.black_list, linter.config.black_list_re)
+        for error in errors:
+            current_reporter.show_file_linted(error['mod'])
+            if isinstance(error['ex'], ImportError):
+                print('Error: {}, which means the file "{}" could not be found.\n'
+                      .format(error['ex'], error['mod']))
+            else:
+                print('Error: {}\n'.format(error['ex']))
+        print()
```
david-yz-liu: This call to `print` isn't very elegant. You should combine this with a previous call, or a future one.

david-yz-liu:
It looks as though you misread my previous comment. The input to `_check` should be a list of strings *or* a single string.

python_ta/__init__.py
```diff
@@ -117,45 +80,102 @@ def _check(module_name='', reporter=ColorReporter, number_of_messages=5, level='
...
+    valid_module_names = []
+
+    # Allow call to check with empty args
+    if module_name == '':
+        m = sys.modules['__main__']
+        spec = importlib.util.spec_from_file_location(m.__name__, m.__file__)
+        module_name = [spec.origin]
+    # Enforce API to expect 1 file or directory if type is list
+    elif isinstance(module_name, str):
+        module_name = list(filter(None, module_name.split(' ')))
```
david-yz-liu: This is interesting. We should allow students to enter a path that contains spaces (students often have directories that contain spaces), so I think it's just better to remove this check.

Issue #166, opened 2017-05-04
global_variables_checker regex issue


nigef:
file: examples/ending_locations/Assign.py
checker: global_variables_checker.py

run: `python_ta.check_all('examples/ending_locations/Assign.py')`

issue: just noticed that checker is not detecting global variable on line 4 of the file. probably an issue with the regex. todo later. probably something like [regex change not shown],

```
regex = str(node.targets[0])
s = re.findall('\((.*?)\)', regex)
if not s:
    return
a = re.match(CONST_NAME_RGX, s[0])
# Raise an error only if it's not a constant
if a is None:
    args = "you declared the global variable '{}' on line {}".\
        format(s[0], node.lineno)
    self.add_message('forbidden-global-variables', node=node,
                     args=args)
```

PR #171, opened 2017-05-04
Fix index error, for empty strings. issue #166

nigef:


david-yz-liu:
This is a good error to point out, but I think you should work on a proper fix that doesn't actually use regular expressions, but instead the structure of the AST. (The `Assign` node has different targets for multiple assignments and unpacking assignments.)

PR #172, opened 2017-05-05
Fixes #166 RE: index error.  Use AST to get node name for global, non-constant variables.

### LINKED ISSUE #166: global_variables_checker regex issue

nigef:


nigef:
lets hold off on merging this for about an hour. i want to examine `visit_global` and `visit_nonlocal` a bit further.

nigef:
added check for visit_name. good now.

PR #175, opened 2017-05-08
Environment class helpers

RyanDJLee:
- Implement helper methods to access and mutate private attributes.

### DIFF
--- a/python_ta/typecheck/base.py
+++ b/python_ta/typecheck/base.py
@@ -240,5 +240,37 @@ class Environment:
         self.nonlocals = nonlocals_ or {}
         self.globals = globals_ or {}
 
+
+    def get_locals(self):
+        """Helper method for accessing local variables in node's environment"""
+        # _locals? private attributes?
+        return self.locals
+
+
+    def add_to_locals(self, variable_name, variable_value):
+        """Helper to add a pair of variable name and value to the locals environment"""
+        self.locals[variable_name] = variable_value
+
+
+    def get_globals(self):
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@RyanDJLee these helpers aren't that helpful. You really want something higher-level that takes in a single string and returns the type variable associated in that string, which internally searches through the different dictionaries (but doesn't require the user to search through them).

python_ta/typecheck/base.py
```diff
@@ -240,5 +240,49 @@ def __init__(self,
...
+    def lookup_in_env(self, variable_name):
+        """Helper to search for a variable in the environment of a node by name."""
+        if variable_name in self.locals:
+            return self.locals[variable_name]
+        elif variable_name in self.globals:
+            return self.globals[variable_name]
+        elif variable_name in self.nonlocals:
+            return self.nonlocals[variable_name]
+        else:
+            return -1
```
david-yz-liu: Don't use return values to indicate an error. Instead, raise an exception. In this case, you can raise a `KeyError`.

python_ta/typecheck/base.py
```diff
@@ -240,5 +240,49 @@ def __init__(self,
         self.nonlocals = nonlocals_ or {}
         self.globals = globals_ or {}
 
+
+    def get_locals(self):
```
david-yz-liu: These methods are not necessary and can be removed. If absolutely necessary, external code can access the attributes directly, but you should attempt to minimize this as much as possible.

PR #177, opened 2017-05-08
Reimplement set env

RyanDJLee:
- Reimplement helper for setting environment using lookup helper.
- Reimplement set environment method

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -258,27 +258,39 @@ def register_type_constraints_setter():
     return type_visitor
 
 
-def _populate_type_env(node):
-    """Helper to prevent overlap in populating locals and globals attributes in type environment of given node."""
-    if isinstance(node, astroid.FunctionDef):
-        node.type_environment = Environment(locals_={name: TYPE_CONSTRAINTS.fresh_tvar() for name in node.locals})
-    node.type_environment = Environment(globals_={name: TYPE_CONSTRAINTS.fresh_tvar() for name in node.globals})
+def _populate_local_env(node):
+    """Helper to populate locals attributes in type environment of given node."""
     for var_name in node.locals:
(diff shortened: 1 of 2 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -258,11 +258,21 @@ def register_type_constraints_setter():
     return type_visitor
 
 
+def _populate_type_env(node):
+    """Helper to prevent overlap in populating locals and globals attributes in type environment of given node."""
+    if isinstance(node, astroid.FunctionDef):
+        node.type_environment = Environment(locals_={name: TYPE_CONSTRAINTS.fresh_tvar() for name in node.locals})
+    node.type_environment = Environment(globals_={name: TYPE_CONSTRAINTS.fresh_tvar() for name in node.globals})
+    for var_name in node.locals:
+        if node.type_environment.lookup_in_env(var_name) == -1: var_value = TYPE_CONSTRAINTS.fresh_tvar()
```
david-yz-liu: Don't use inline statements for any type of control flow.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -258,11 +258,21 @@ def register_type_constraints_setter():
     return type_visitor
 
 
+def _populate_type_env(node):
+    """Helper to prevent overlap in populating locals and globals attributes in type environment of given node."""
+    if isinstance(node, astroid.FunctionDef):
+        node.type_environment = Environment(locals_={name: TYPE_CONSTRAINTS.fresh_tvar() for name in node.locals})
+    node.type_environment = Environment(globals_={name: TYPE_CONSTRAINTS.fresh_tvar() for name in node.globals})
```
david-yz-liu: This does not actually fix the problem mentioned earlier. The implementations of the `_set_environment` method are different enough for the two classes (`Module` and `FunctionDef`) that you should just split them up into two entirely different functions.

PR #185, opened 2017-05-11
Add test for the python_ta check method.

nigef:


python_ta/__init__.py
```diff
@@ -23,7 +23,9 @@
 import pylint.lint as lint
 import pylint.utils as utils
 import pylint
+
 from astroid import MANAGER
+from astroid import modutils
```
david-yz-liu: Add `modutils` to the previous import line.

python_ta/__init__.py
```diff
@@ -111,6 +113,20 @@ def _verify_pre_check(filepath):
...
+def get_file_paths(rel_path):
+    """A generator for iterating python files within a directory.
+    `rel_path` is a relative path to a file or directory.
+    Returns paths to all files in a directory.
+    TODO: refactor helpers into a new file, `python_ta/utils.py`
+    """
+    if not os.path.isdir(rel_path):
+        yield rel_path  # Don't do anything; return the file name.
+
+    for root, _, files in os.walk(rel_path):
```
david-yz-liu: Put this block in an `else` related to the `if` from above.

tests/test_check.py
```diff
@@ -0,0 +1,24 @@
...
+"""
+import unittest
+import os
+import sys
+import python_ta
+
+_NODES_PATH_TOP_LEVEL = 'nodes'
+
+
+class TestCheck(unittest.TestCase):
```
david-yz-liu: Whoops, didn't notice this one earlier. We aren't using `unittest`; you can write a top-level `test` function for running with `nosetests`.

PR #186, opened 2017-05-11
Test single target assign

RyanDJLee:
- Import TYPE_CONSTRAINTS object.
- Create test case for single variable assignment.


david-yz-liu:
This is definitely the right idea. A couple of overall comments:

1. You should be using hypothesis here.
2. Use the `nodes_of_class` method with `AssignName` to access those nodes in a better way.

PR #188, opened 2017-05-12
Search directory path for a `.pylintrc` config file

nigef:
Check for `.pylintrc` file provided in user location (i.e. the student starter code). Does not search upward, in parent directories.

In the case where `_check` is called on a directory and files are checked recursively, a cache of pylintrc locations is _not_ used for now, so there is more work done loading a config files.

Aside: although different from our approach, for background info on how pylint finds a `.pylintrc` file, see, `pylint.config.find_pylintrc`

### DIFF
--- a/python_ta/__init__.py
+++ b/python_ta/__init__.py
@@ -55,11 +55,11 @@ def _find_pylintrc():
     """Search bottom-up for a `.pylintrc` file provided in user location.
     Return absolute path to the file, or None.
     • When run in an IDE (like PyCharm), sys.argv contains the full path,
-        - sys.argv[0] is: /Users/fong/Desktop/pyta/f0/f1/f2/new.py
-        - os.getcwd() is: /Users/fong/Desktop/pyta/f0/f1/f2
+        - sys.argv[0] is: /<users>/<user>/<path_to>/dir1/dir2/file.py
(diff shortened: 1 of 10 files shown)


nigef:
RE: 1. I don't see a realistic way around this.

nigef:
RE: 4. cache removed! aside: I'm curious why we don't want it.

python_ta/__init__.py
```diff
@@ -155,7 +129,7 @@ def _verify_pre_check(filepath):
         # Check for inline "pylint:" comment, which may indicate a student
         # trying to disable a check.
         with tokenize.open(filepath) as f:
-            for (tok_type, content, _, _, _) in tokenize.generate_tokens(f.readline):
+            for (tok_type,content,_,_,_) in tokenize.generate_tokens(f.readline):
```
david-yz-liu: Please revert this change.

david-yz-liu:
For (1), you could create a new linter for each file.
For (4), in general caching introduces an extra level of complexity that should be avoided until deemed necessary.

nigef:
ready to merge?

david-yz-liu:
I still see a problem with the "#1" I pointed out earlier.

PR #189, opened 2017-05-12
Throw away unused variables in test_examples

nigef:


tests/test_examples.py
```diff
@@ -23,7 +23,7 @@ def get_file_paths():
     _EXAMPLES_PATH prefix followed by the file name for each element.
     A list of all the file paths will be returned."""
     test_files = []
-    for root, directories, files in os.walk(_EXAMPLES_PATH, topdown=True):
+    for _, _, files in os.walk(_EXAMPLES_PATH, topdown=True):
```
david-yz-liu: This is interesting. If there are nested directories under the examples, it seems like we actually should be using `root` to actually specify the correct path to the files in these nested directories.

PR #191, opened 2017-05-15
Add helper to generate a dictionary of variable name and value pairs.

RyanDJLee:


david-yz-liu:
@RyanDJLee this is good!

PR #195, opened 2017-05-15
Reimplement visitor for tuple to account for context.

RyanDJLee:


python_ta/transforms/type_inference_visitor.py
```diff
@@ -43,10 +43,13 @@ def set_const_type_constraints(node):
 
 
 def set_tuple_type_constraints(node):
-    # node_types contains types of elements inside tuple.
-    node.type_constraints = TypeInfo(
-        Tuple[tuple(x.type_constraints.type for x in node.elts)]
-    )
+    # If tuple is on RHS, find and set Tuple type
```
david-yz-liu: The "load" context is for more than just the right-hand side of an assignment statement. The key idea here is that the tuple is being evaluated, rather than assigned to.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -43,10 +43,13 @@ def set_const_type_constraints(node):
...
+    if node.ctx == astroid.Load:
+        node.type_constraints = TypeInfo(
+                Tuple[tuple(x.type_constraints.type for x in node.elts)])
+    else:
+        # Tuple is on LHS; has no type - yet.
```
david-yz-liu: The node in this case will *never* have a type other than `NoType`, so the "yet" is misleading.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -43,10 +43,13 @@ def set_const_type_constraints(node):
...
 
 def set_tuple_type_constraints(node):
-    # node_types contains types of elements inside tuple.
-    node.type_constraints = TypeInfo(
-        Tuple[tuple(x.type_constraints.type for x in node.elts)]
-    )
+    # If tuple is on RHS, find and set Tuple type
+    if node.ctx == astroid.Load:
+        node.type_constraints = TypeInfo(
+                Tuple[tuple(x.type_constraints.type for x in node.elts)])
```
david-yz-liu: Make the indentation here 4 spaces more than the previous line.

PR #198, opened 2017-05-15
Add test for many different inputs.

nigef:
Merge after pull #185

This is the lot of tests I was running while working on `_check` in our `__init__.py`
Good to have as regression test.


david-yz-liu:
This is okay, but I will consider this only after my comments on #185 have been addressed.

PR #199, opened 2017-05-16
Refactor testing.

RyanDJLee:
- Extract test cases for assignment statements into separate file.
- Fix import statements accordingly.
- Move parsing function into helpers module.


david-yz-liu:
Seems like 6 tests were deleted, but only 4 are in the new file?

PR #202, opened 2017-05-17
Add helper to return valid identifiers.

RyanDJLee:
- Add helper function to return generator for valid identifiers.


david-yz-liu:
Yeah, I don't think the `kwargs` is necessary. Fix the `min_size` to be 1.

PR #203, opened 2017-05-17
Add helper for getting types from a Callable

RyanDJLee:
- Implement helper method to get types in Callable representing a function.


david-yz-liu:
1. This is a good place for a list comprehension.
2. It would be better to return a tuple of `(arg_types, return_type)`.

PR #204, opened 2017-05-19
Pass custom config options from `pylintrc` to checkers/reporters.

nigef:
We can now define custom config options in a `.pylintrc` file by passing in a tuple of options at the constructor of our Linter (which is a type of checker). This is similar to how PyLinter.options are set.

Class variable `name` is added to reporters to for compatibility with setting the Reporter using the `output-format` config option in our pylintrc, which is how pylint does it.

Config options are available in Reporters, by making the linter (checker) an instance variable of reporters.

### DIFF
--- a/python_ta/.pylintrc
+++ b/python_ta/.pylintrc
@@ -1,3 +1,17 @@
+[CUSTOM PYTA OPTIONS]
+
+# Make sure to register custom options tuple first in `python_ta/__init__.py`
+# ===========================================================
+# Default max amount of messages for reporter to display.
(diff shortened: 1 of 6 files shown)


nigef:
Note the checks fail bc of the change to `python_ta/.pylintrc`, which pylint cannot use anymore. 
To fix, the `tests/text_examples` should use _pyta_ instead of pylint to check for the existence of `checker_name` in message output.

david-yz-liu:
@nigef good job, but this branch should not include the commits from #188.

PR #206, opened 2017-05-19
Fix helper method for returning types contained in Callable

RyanDJLee:
- Initialize list for returning in helper.

### DIFF
--- a/python_ta/typecheck/base.py
+++ b/python_ta/typecheck/base.py
@@ -212,8 +212,7 @@ class TypeConstraints:
     ### HELPER METHODS
     def types_in_callable(self, callable_function):
         """Return a tuple of types corresponding to the Callable function's arguments and return value, respectively."""
-        arg_type_lst = []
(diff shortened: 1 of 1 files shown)


python_ta/typecheck/base.py
```diff
@@ -212,8 +212,9 @@ def lookup_concrete(self, tvar):
     ### HELPER METHODS
     def types_in_callable(self, callable_function):
         """Return a tuple of types corresponding to the Callable function's arguments and return value, respectively."""
-        types_lst = [arg_type_lst.append(self.lookup_concrete(argument)) for argument in callable_function.__args__]
-        return types_lst[:-1], types_lst[-1]
+        arg_type_lst = []
+        [arg_type_lst.append(self.lookup_concrete(argument)) for argument in callable_function.__args__]
```
david-yz-liu: Sorry I missed this last time. You should not use `append` here. Just do `types_lst = [self.lookup_concrete(argument) for ...]`

PR #209, opened 2017-05-19
Pass custom config options from `pylintrc` to checkers/reporters.

nigef:
We can now define custom config options in a `.pylintrc` file by passing in a tuple of options at the constructor of our Linter (which is a type of checker). This is similar to how PyLinter.options are set.

Config options are available in Reporters, by making the linter (checker) an instance variable of reporters.

Check for .pylintrc file provided in user location (i.e. the student starter code). Does not search upward, in parent directories.

In the case where _check is called on a directory and files are checked recursively, a cache of pylintrc locations is not used for now, so there is more work done loading a config files.

Aside: although different from our approach, for background info on how pylint finds a .pylintrc file, see, pylint.config.find_pylintrc


python_ta/__init__.py
```diff
@@ -22,39 +22,38 @@
...
 from astroid import MANAGER, modutils
 
-from .reporters import ColorReporter
+from .reporters import *
 from .patches import patch_all
 
 # Local version of website; will be updated later.
 HELP_URL = 'http://www.cs.toronto.edu/~david/pyta/'
 
+REPORTERS = (ColorReporter, PlainReporter, HTMLReporter, StatReporter)
```
david-yz-liu: I recommend having the `reporter/__init__.py` export a tuple of the different reporter classes, rather than putting it here. (I believe `pylint` does something similar with `ALL_NODE_CLASSES`)

python_ta/reporters/plain_reporter.py
```diff
@@ -106,6 +107,10 @@ def print_messages(self, level='all'):
         print('\n')
 
     def sort_messages(self):
+        max_messages = float('inf')
```
david-yz-liu: This should be 5, which is the reported default. On the other hand, It should be documented in the sample `.pylintrc` file how to configure pyta to display an unlimited number of messages (by setting the option's value to `float('inf')`?)

python_ta/__init__.py
```diff
@@ -153,8 +153,35 @@ def _check(module_name='', reporter=ColorReporter, number_of_messages=5, level='
         print('No checks run. Input to check, `{}`, has invalid type, must be a list of strings.\n'.format(module_name))
         return
 
-    current_reporter = reporter(number_of_messages)
-    linter = _load_pylint_plugins(current_reporter, local_config, pep8)
+    # see 'type' in pylint/config.py `VALIDATORS` dict.
+    new_options = (
```
david-yz-liu: This is good stuff, and it seems like it could be entirely pulled out into a helper function that returns a linter.

python_ta/reporters/plain_reporter.py
```diff
@@ -53,13 +53,14 @@
 
 
 class PlainReporter(BaseReporter):
-    def __init__(self, number_of_messages, source_lines=None, module_name=''):
+
```
david-yz-liu: Please don't add this blank line.

python_ta/reporters/plain_reporter.py
nigef: We can use the default set in `init_linter()` in `python_ta/__init__.py`
I didn't see an easy way to allow a setting for infinite messages, should I make a new issue for that?

nigef:
yes now it does play well with the `config` dictionary argument

nigef:
lets see if this #209 can be merged, then i will try to fix #188 according to the changes.

python_ta/__init__.py
```diff
@@ -22,10 +22,11 @@
 
 import pylint.lint
 import pylint.utils
+from pylint.config import VALIDATORS, _call_validator
 
 from astroid import MANAGER, modutils
 
-from .reporters import ColorReporter
+from .reporters import *
```
david-yz-liu: Import `REPORTERS` here.

python_ta/__init__.py
```diff
@@ -37,52 +38,135 @@
...
+    """
+    found_pylintrc_location = None
+    if os.path.exists(os.path.join(curr_dir, '.pylintrc')):
+        found_pylintrc_location = os.path.join(curr_dir, '.pylintrc')
+    elif os.path.exists(os.path.join(curr_dir, 'pylintrc')):
+        found_pylintrc_location = os.path.join(curr_path, 'pylintrc')
+    return found_pylintrc_location
+    
+
+def _apply_nearest_config(linter=None, file_abs_path=''):
```
david-yz-liu: This is better but still too complicated. This helper should be merged with `init_linter`. The merged method should do the following:

0. It should *not* take in a `linter`.
1. Given the `file_abs_path`, determine which configuration to use:
    - If the user's `config` argument is a string, use the config found at that location.
    - If the user's `config` argument is a dictionary, use those options in addition to either the pylintrc file found at the directory of the file being linted, or the default one.
    - If the user's `config` argument is empty, use the pylintrc file found at the directory of the file being linted, or the default one.
2. Return a new `linter` object.

More specifically, do not ever try to re-use a linter object.

PR #212, opened 2017-05-23
Hypothesis settings

RyanDJLee:
- Change settings of Hypothesis testing framework so that Hypothesis generates at most 10 examples per test case for assignments and functions testing.


david-yz-liu:
@RyanDJLee it is better to create a new profile (http://hypothesis.readthedocs.io/en/latest/settings.html#settings-profiles). You don't need to pass the profile name at the command-line - just call`settings.load_profile` at the top of each test file.

PR #216, opened 2017-05-25
Reimplement type store

RyanDJLee:
- Rename methods attribute to functions
- Import Any
- Change value of dictionaries to tuples; instantiate Callables with types not strings.


david-yz-liu:
@RyanDJLee For the function dictionary, each list actually should have a tuple, whose second element is the name of the *class* that function is from.

PR #217, opened 2017-05-26
Add method type store

RyanDJLee:
- Rename methods attribute to functions.
- Add method lookup_functions to TypeStore class to replace lookup_methods.
- Instantiate TYPE_STORE singleton.
- Fix path setting of TypeShed source code.

### DIFF
--- a/python_ta/typecheck/type_store.py
+++ b/python_ta/typecheck/type_store.py
@@ -49,7 +49,7 @@ class TypeStore:
                     found = True
                     break
             if not found:
-                raise TypeError
+                raise KeyError
         return function_type
 
 TYPE_STORE = TypeStore()
--- a/python_ta/typecheck/type_store.py
+++ b/python_ta/typecheck/type_store.py
@@ -35,3 +35,18 @@ class TypeStore:
                     concrete_arg_types.append(concrete_arg_type)
                 self.classes[class_def.name][function_def.name] = (Callable[concrete_arg_types, rtype], class_def.name)
                 self.functions[function_def.name].append(Callable[concrete_arg_types, rtype])
+
+    def look_up_function(self, operator, *args):
+        """Helper method to lookup a function type."""
+        function_type = None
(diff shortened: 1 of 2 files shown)


python_ta/typecheck/type_store.py
```diff
@@ -1,12 +1,13 @@
 import astroid
 from collections import defaultdict
 from typing import Callable, Any
-
+import os
+TYPE_SHED_PATH = os.path.join(os.path.dirname(__file__), 'typeshed/builtins.pyi')
```
david-yz-liu: Don't hard-code the forward-slash (different on Windows). Simply pass `'typeshed'` and `'builtins.pyi'` as arguments to the `os.path.join` -- it'll join them properly.

python_ta/typecheck/type_store.py
```diff
@@ -35,3 +36,20 @@ def __init__(self):
...
+            for funct_type in funct_types_list:
+                if funct_type.__args__[:-1] == args:
+                    function_type = funct_type
+                    found = True
+                    break
+            if not found:
+                raise KeyError
+        return function_type
+
+TYPE_STORE = TypeStore()
```
david-yz-liu: Make sure to configure your editor to ensure all files end with a blank line.

python_ta/typecheck/type_store.py
```diff
@@ -35,3 +36,20 @@ def __init__(self):
...
+
+    def lookup_function(self, operator, *args):
+        """Helper method to lookup a function type given the operator and types of arguments."""
+        function_type = None
+        if args:
+            funct_types_list = self.functions[operator]
+            found = False
+            for funct_type in funct_types_list:
+                if funct_type.__args__[:-1] == args:
+                    function_type = funct_type
```
david-yz-liu: You can just return `funct_type` here directly.

Also, the more conventional abbreviation for "function" is "func".

PR #218, opened 2017-05-26
Add helpers.

RyanDJLee:
- Add function that returns strategy for generating boolean operator symbol.
- Helper to get non-boolean primitive values.

### DIFF
--- a/tests/custom_hypothesis_support.py
+++ b/tests/custom_hypothesis_support.py
@@ -79,7 +79,7 @@ def non_boolean_operator():
 
 def return_one_of(list_of_strings):
     """Return a strategy that returns one of given string arguments."""
-    return hs.one_of([hs.text(alphabet=symbol, min_size=1, max_size=2) for symbol in list_of_strings])
+    return hs.one_of([hs.text(alphabet=symbol, min_size=len(symbol)) for symbol in list_of_strings])
 
 
 # Helper functions for testing
--- a/tests/custom_hypothesis_support.py
+++ b/tests/custom_hypothesis_support.py
@@ -79,7 +79,7 @@ def non_boolean_operator():
 
 def return_one_of(list_of_strings):
     """Return a strategy that returns one of given string arguments."""
-    return hs.one_of([hs.text(alphabet=symbol, min_size=1, max_size=1) for symbol in list_of_strings])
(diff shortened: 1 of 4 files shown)


tests/custom_hypothesis_support.py
```diff
@@ -63,6 +71,13 @@ def heterogeneous_dictionary(**kwargs):
     return hs.dictionaries(index_values, primitive_values, **kwargs)
 
 
+def boolean_operator():
+    """Return a strategy which generates a boolean operator symbol (string)."""
+    return hs.one_of(hs.text('+-/*%', min_size=1, max_size=1),
```
david-yz-liu: This is not an appropriate use of `hs.text`. You should use a strategy that takes as input a list of strings, and returns one of the strings.

tests/custom_hypothesis_support.py
```diff
@@ -63,6 +71,17 @@ def heterogeneous_dictionary(**kwargs):
     return hs.dictionaries(index_values, primitive_values, **kwargs)
 
 
+def non_boolean_operator():
+    """Return a non-boolean operator symbol (string)."""
+    non_bool_operators = ['+', '-', '*', '//', '%', '/', '**', '&', '^', '~', '|', '<<', '>>']
+    return return_one_of(non_bool_operators)
```
david-yz-liu: Better, but instead of using `hs.one_of`, use `hs.sampled_from`. This will eliminate the need for a helper.

tests/custom_hypothesis_support.py
```diff
@@ -19,6 +19,14 @@
 ])
 primitive_values = primitive_types.flatmap(lambda s: s())
 
+non_bool_primitive_types = hs.sampled_from([
```
david-yz-liu: You don't need these. Just use the already-defined primitive types in the new test.

tests/type_inference/test_ops.py
```diff
@@ -0,0 +1,24 @@
+import nose
+from hypothesis import given, settings, assume
+import tests.custom_hypothesis_support as cs
+from python_ta.transforms.type_inference_visitor import *
```
david-yz-liu: Never import `*`

PR #221, opened 2017-05-30
Output htmlreporter

nigef:
Embed assets into html, inline. Output to cwd.


python_ta/reporters/html_reporter.py
```diff
@@ -3,11 +3,14 @@
...
 from jinja2 import Environment, FileSystemLoader
 from datetime import datetime
+from base64 import b64encode
 
 from .color_reporter import ColorReporter
 
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
-
+OUTPUT_DIR = os.getcwd()
+TEMPLATE_FILE = '/templates/template.txt'
```
david-yz-liu: Use `os.path.join` for file paths. (This was done incorrectly before.)

PR #224, opened 2017-05-31
Output htmlreporter

nigef:


python_ta/reporters/html_reporter.py
```diff
@@ -32,20 +35,26 @@ def print_messages(self, level='all'):
         self._colour_messages_by_type(style=False)
         self._colour_messages_by_type(style=True)
 
-        template = Environment(loader=FileSystemLoader(THIS_DIR)).get_template('templates/template.txt')
-        output_path = THIS_DIR + '/templates/output.html'
+        template = Environment(loader=FileSystemLoader(THIS_DIR)).get_template(TEMPLATE_FILE)
```
david-yz-liu: This suggests that `THIS_DIR` can be replaced entirely with `TEMPLATES_DIR` (which you need for the image below anyway).

PR #227, opened 2017-06-01
Output htmlreporter recursive linting

nigef:
• Output the html file in current working directory.
• Display all files' messages in htmlreporter, when linting done on a directory. Previously multiple files unsupported by htmlreporter.


david-yz-liu:
@nigef seems like CircleCI failed with an odd error - does this happen when you run the tests locally?

nigef:
strange, no, it doesn't happen locally. i'm looking into it

PR #228, opened 2017-06-02
Inference for lists

RyanDJLee:
- Use helper to add polymorphic type variables in type store function types.
- Implement can_unify helper and update methods in TypeConstraints as well as visitor to properly unify function calls.
- Add test case for Subscript node representing homogeneous list with integer index access (__getitem__ builtin).


python_ta/typecheck/base.py
```diff
@@ -219,6 +221,14 @@ def can_unify(self, t1, t2):
...
         if isinstance(t1, TypeVar) or isinstance(t2, TypeVar):
             return True
+        elif isinstance(t1, GenericMeta) and isinstance(t2, GenericMeta):
+            if not _geqv(t1, t2):
+                return False
+            elif t1.__args__ is not None and t2.__args__ is not None:
+                for a1, a2 in zip(t1.__args__, t2.__args__):
+                    if not self.can_unify(a1, a2):
+                        return False
+                return True
```
david-yz-liu: You should have a final `else` block here; otherwise this method might return `None`.

python_ta/typecheck/type_store.py
```diff
@@ -36,7 +37,15 @@ def __init__(self, type_constraints):
                     if gen == 'Generic':
                         self.classes[class_def.name]['__pyta_tvars'] = tvars
             for function_def in class_def.nodes_of_class(astroid.FunctionDef):
+                poly_tvars = []
                 arg_types = []
+                try:
```
david-yz-liu: You don't need a try-except here. You can use the `dict` method `get` to avoid a `KeyError`.

python_ta/typecheck/type_store.py
```diff
@@ -36,7 +37,15 @@ def __init__(self, type_constraints):
...
+                    tvars = self.classes[class_def.name]['__pyta_tvars']
+                    for tvar in tvars:
+                        if isinstance(eval(tvar), TypeVar):
```
david-yz-liu: You can assume that `tvar` always evaluates to a `TypeVar`. Do `eval(tvar, globals())` so that the type variables defined at the top of this file are in scope.

python_ta/typecheck/type_store.py
```diff
@@ -36,7 +37,15 @@ def __init__(self, type_constraints):
                     if gen == 'Generic':
                         self.classes[class_def.name]['__pyta_tvars'] = tvars
             for function_def in class_def.nodes_of_class(astroid.FunctionDef):
+                poly_tvars = []
                 arg_types = []
+                try:
+                    tvars = self.classes[class_def.name]['__pyta_tvars']
+                    for tvar in tvars:
```
david-yz-liu: This entire block should be simplified to a list comprehension.

python_ta/typecheck/type_store.py
```diff
@@ -97,3 +104,5 @@ def _builtin_to_typing(self, type_name):
             tvar_string = ''
 
         return base_name + tvar_string
+
```
david-yz-liu: Revert this change.

david-yz-liu:
@RyanDJLee whoops I think I approved the changes too soon. There is still some work to be done here.

PR #230, opened 2017-06-05
Create snippets for messages C0326, C0301, and fix R1705.

nigef:
• Add snippets to messages for C0326 bad-whitespace, C0301 line-too-long. Add `msg_truncated` attribute to namedtuple object. Print message from the `msg.msg_truncated` property now.

• Fix off by one error on line 68, that would wrap around to the last [-1] line and display it as line 0.

• R1705 `no-else-return` was missing some lines, and double counting others. This fix makes the snippet great again.

• Also fixes C0330 to remove redundant ` }\n| |^` from message output.

### DIFF
--- a/python_ta/reporters/node_printers.py
+++ b/python_ta/reporters/node_printers.py
@@ -23,7 +23,7 @@ def render_generic(msg, source_lines=None):
             yield (start_line, slice(start_col, end_col), LineType.ERROR, source_lines[start_line-1])
         else:
             yield (start_line, slice(start_col, None), LineType.ERROR, source_lines[start_line-1])
-            yield from ((line, slice(None, None), LineType.ERROR, source_lines[line-1]) for line in range(start_line, start_col))
+            yield from ((line, slice(None, None), LineType.ERROR, source_lines[line-1]) for line in range(start_line+1, end_line))
             yield (end_line, slice(None, end_col), LineType.ERROR, source_lines[end_line-1])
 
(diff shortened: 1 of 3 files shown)


david-yz-liu:
I'm not sure this is working properly for `bad-whitespace`.

1. Passing `args` to `handle_node` doesn't seem to do anything (`handle_node` doesn't use it?).
2. In general, I believe `args` is consumed when creating the `msg` attribute (passed into the format string). So in principle we shouldn't have to access it if we have the message string.
3. I believe the real problem is with Lines 211-214 of `plain_reporter.py`, which seems to truncate `msg.msg`. This should be removed, and instead the post-processing should happen in the new `render_bad_whitespace`, and when the new tuple is created (with the snippet added).

python_ta/reporters/node_printers.py
```diff
@@ -65,14 +65,31 @@ def render_trailing_newlines(msg, source_lines=None):
...
+            stop = start + 1
+            break
+
+    line = msg.line
+    yield from render_context(line - 2, line, source_lines)
+    yield (line, slice(start, stop), LineType.ERROR, source_lines[line - 1])
+    yield from render_context(line + 1, line + 3, source_lines)
+
+
+
```
david-yz-liu: Please remove one blank line.

david-yz-liu:
I'd rather not add more fields to `NewMessage`. Instead compute a truncated version to display when printing the message.

PR #231, opened 2017-06-05
Remove redundant `     }\n|   |^` from message output.

nigef:
Example,
```
C0330 (bad-continuation)  Number of occurrences: 1.
  [Line 9] Wrong hanging indentation.
     }
|   |^
    7      'b': 1,
    8      'c': 2
    9       }
```

is now simply,
```
C0330 (bad-continuation)  Number of occurrences: 1.
  [Line 9] Wrong hanging indentation.
    7      'b': 1,
    8      'c': 2
    9       }
```

with the close brace still highlighted by color reporter (not shown).


david-yz-liu:
Hmm, interesting interaction with my comment on the other pull request. Okay, so in this case why don't we just create a new variable with the truncated message (for use in the subsequent `format`), but not actually replace the `msg` tuple (so that location information can be parsed from the string).

PR #235, opened 2017-06-06
expand/collapse sections of html output with javascript

rebeccakay:


python_ta/reporters/templates/template.txt
```diff
@@ -94,6 +107,12 @@
 		</a>
 	</div>
 </body>
+<script
```
nigef: Now we will download and place this minified jquery in pyta/python_ta/reporters/templates/ and include it, 
similar to the include in the link below, but with different enclosing element, for inline scripts. look up how to do this.

Similarly, directly after the above script element, we can "include" the script.js -- we may be to put the includes in the same enclosing script element, otherwise in a new one directly after.

https://github.com/pyta-uoft/pyta/blob/master/python_ta/reporters/templates/template.txt#L4

python_ta/reporters/templates/template.txt
```diff
@@ -35,12 +35,17 @@
 			</div>
 			<u1>
 				{% for message in file.code %}
+				    <div class="container">
+				    <div class="slider">
```
nigef: indent nested elements, and vertically match the opening tags with their closing tags.

python_ta/reporters/templates/template.txt
```diff
@@ -35,12 +35,17 @@
...
 					<p style="font-size: 110%">
 						<span class="code_id">
 						    <a href="http://www.cs.toronto.edu/~david/pyta/&#35;{{message}}" target="_blank">{{ message }} ({{ file.code[message][0].symbol }})</a> 
                         </span>
 						&nbsp;&nbsp;Number of occurrences: {{ file.code[message]|length }}
 					</p>
+					</div>
+
+					<div id ="content" class="collapse in">
```
nigef: are the `in` or `content` selectors necessary?

python_ta/reporters/templates/stylesheet.css
```diff
@@ -134,3 +134,25 @@ body {
     text-align: center;
     padding-top: 10px;
 }
+
+#.hidden {
```
nigef: try to avoid using `id` selectors.

remember, id selected with `#`, class selected with `.`

nigef:
@rebeccakay a reminder about the work here, since it was not done for the last meeting.

PR #237, opened 2017-06-06
Fix forbidden global variables (import names considered global variables)

nigef:
Prevent import names from being considered global variables, e.g. `doctest`, `python_ta`:
```
E9997 (forbidden-global-variables)  Number of occurrences: 3.
  [Line 57] Global variables should not be used in CSC108/CSC148 - a global variable 'doctest' is declared on line 57
   55  [line omitted]
   56  
   57  doctest.testmod()
   58  python_ta.check_all()
   59  

  [Line 58] Global variables should not be used in CSC108/CSC148 - a global variable 'python_ta' is declared on line 58
   56  
   57  doctest.testmod()
   58  python_ta.check_all()
   59  
   60  
```

### DIFF
--- a/python_ta/checkers/global_variables_checker.py
+++ b/python_ta/checkers/global_variables_checker.py
@@ -18,6 +18,10 @@ class GlobalVariablesChecker(BaseChecker):
     # this is important so that your checker is executed before others
     priority = -1
 
+    def __init__(self, linter=None):
+        super().__init__(linter)
+        self.import_names = []
+
     def visit_global(self, node):
         args = "the keyword 'global' is used on line {}".format(node.lineno)
         self.add_message('forbidden-global-variables', node=node, args=args)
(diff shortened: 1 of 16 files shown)


david-yz-liu:
This is good, but I believe there are other types of import nodes (`ImportFrom`) that should also be considered.

python_ta/checkers/global_variables_checker.py
```diff
@@ -38,10 +42,26 @@ def visit_name(self, node):
...
+    def visit_import(self, node):
+        """Save the names of imports, to prevent mistaking for global vars."""
+        self._store_name_or_alias(node)
+
+    def visit_importfrom(self, node):
+        """Save the names of imports, to prevent mistaking for global vars."""
+        self._store_name_or_alias(node)
+
+    def _store_name_or_alias(self, node):
+        if node.names[0][1] is not None:
```
david-yz-liu: I'm not really sure about the structure of `node.names` (it would be good to update the documentation in the `nodes/` folder as well), but it seems like it's a list; this would suggest that a single `Import` node can actually have a bunch of imports (e.g., `import a, b, c`) and this method is only adding the first one.

PR #240, opened 2017-06-06
Dev test boolops

RyanDJLee:
- Re-document BoolOp visitor.
- Implement visitor for BoolOp using Union.
- Add test case for BoolOp representing binary boolean operation.

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -161,9 +161,21 @@ def set_compare_type_constraints(node):
 
 def set_boolop_type_constraints(node):
     """Boolean operators includes: 'and', 'or';can be "either" of the types."""
-    left_operand_type = node.values[0].type_constraints.type
-    right_operand_type = node.values[1].type_constraints.type
-    node.type_constraints = TypeInfo(Union[left_operand_type, right_operand_type])
+    if len(node.values) == 2:
+        left_operand_type = node.values[0].type_constraints.type
+        right_operand_type = node.values[1].type_constraints.type
+        node.type_constraints = TypeInfo(Union[left_operand_type, right_operand_type])
+    else:
+        homogeneous = True
+        operand_type = node.values[0].type_constraints.type
+        for operand in node.values:
+            if operand.type_constraints.type != operand_type:
+                homogeneous = False
+                break
+        if not homogeneous:
+            node.type_constraints = TypeInfo(Any)
+        else:
+            node.type_constraints = TypeInfo(operand_type)
 
 
 ##############################################################################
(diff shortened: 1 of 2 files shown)


david-yz-liu:
This is a good test; however, the fact that the attribute of `BoolOp` is a list of values makes me nervous. Can you try to inspect something like `x and y and z` and see what the AST looks like? If there are more than two possible arguments, then I think we should switch to the same model as lists, where the type of `x and y and z and ...` is a single type if all of the values have the same type, or else the type is `Any`.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -160,21 +160,22 @@ def set_compare_type_constraints(node):
...
-    if node.op == 'or':
-        if not node.values[0]:
-            node.type_constraints = node.values[1].type_constraints
-        else:
-            node.type_constraints = node.values[0].type_constraints
-    elif node.op == 'and':
-        if not node.values[0]:
-            node.type_constraints = node.values[0].type_constraints
+    """Boolean operators includes: 'and', 'or';can be "either" of the types."""
+    if len(node.values) == 2:
```
david-yz-liu: Don't use length 2 as a special case. Look at how the list type constraints are set and use that approach.

tests/type_inference/test_boolops.py
```diff
@@ -0,0 +1,34 @@
+import astroid
+import nose
+from hypothesis import given, settings, assume
+import hypothesis.strategies as hs
+import tests.custom_hypothesis_support as cs
+from typing import Any
+settings.load_profile("pyta")
+
+
+@given(hs.lists(cs.binary_bool_operator), hs.lists(cs.primitive_values))
```
david-yz-liu: Can you mix `and` and `or` in a single `BoolOp` node? If not, what you should do take in a single bool op, and a list of values (of length `>= 2`) and then join the list using the single operation.

You should also have two different cases, one with a homogeneous list, and one with a heterogeneous list.

tests/type_inference/test_boolops.py
```diff
@@ -0,0 +1,56 @@
...
+def test_homogeneous_binary_boolop(op, operand_list):
+    """Test type setting of binary BoolOp node(s) representing expression with same binary boolean operations."""
+    # get every permutation?
+    assume(len(operand_list) > 0)
+    pre_format_program = [repr(operand) for operand in operand_list]
+    program = (' ' + op + ' ').join(pre_format_program)
+    module = cs._parse_text(program)
+    boolop_node = list(module.nodes_of_class(astroid.BoolOp))[0]
+    operand_type = type(operand_list[0])
+    homogeneous = True
```
david-yz-liu: Use the right strategy to avoid doing this check.

PR #242, opened 2017-06-07
Add helper that returns 'type' of elements inside a given list 

RyanDJLee:


david-yz-liu:
This helper is not necessary if you use the custom strategies you created.

PR #245, opened 2017-06-08
Ignore C0301 line-too-long on doctests (monkey patch FormatChecker.check_lines)

nigef:
Rather than monkey patching `new_line` too, we simply do a regex search before the call to [`check_line`](https://github.com/PyCQA/pylint/blob/126c63db8010d23afd18cf65e4a6b45f4bf79efd/pylint/checkers/format.py#L1033)

Although it is not cheap, with linear time cost to regex match a line, it seems acceptable.

[The other approach would have been to search for the line number of the line-too-long message in the token object in the `new_line` function, possibly filtering by type string, the string starting with triple quotes, and then searching for the '>>>' substring...]


python_ta/patches/checkers.py
```diff
@@ -99,3 +103,63 @@ def _get_attribute_property_setter(name, klass):
...
+                continue
+
+            if unsplit:
+                unsplit.append(line)
+                line = ''.join(unsplit)
+                unsplit = []
+
+            # Skip error message for long doctest lines
+            doctest_tokens = compile(r'^\s*>>>.*?\n$')
+            if bool(match(doctest_tokens, line)):
```
david-yz-liu: You shouldn't need to call `bool` explicitly.

david-yz-liu:
@nigef please see my above comment about the documentation

PR #249, opened 2017-06-12
Dev test compare

RyanDJLee:
- Update dunder helper with compare operators
- Implement test case for Compare node representing comparisons between homogeneous-typed values.
- Add TODO for unification problem between a type and object type.

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -201,12 +201,6 @@ class TypeInferer:
             else:
                 node.type_constraints = TypeInfo(return_type)
 
-    # TODO: Add check in the visit_compare as in BinOp.
-    def visit_compare(self, node):
-        """Compare operators includes:
-        '<', '>', '==', '>=', '<=', '<>', '!=', 'is' ['not'], ['not'] 'in' """
-        node.type_constraints = TypeInfo(bool)
-
     def visit_boolop(self, node):
         """Boolean operators are 'and', 'or'; the result type can be either of the argument types."""
         node_type_constraints = {operand_node.type_constraints.type for operand_node in node.values}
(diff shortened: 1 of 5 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -215,6 +209,17 @@ def visit_boolop(self, node):
         else:
             node.type_constraints = TypeInfo(Any)
 
+    # TODO: Add check in the set_compare_type_constraints as in BinOp.
+    def visit_compare(self, node):
+        """Compare operators includes:
+        '<', '>', '==', '>=', '<=', '!='. """
+        function_type = self.type_store.lookup_function(op_to_dunder_binary(node.ops[0][0]),
```
david-yz-liu: You aren't interpreting `node.ops` correctly. It's for chained comparisons. Please consult https://docs.python.org/3/reference/expressions.html#comparisons.

Also it would be good to improve the documentation in the `nodes/Compare.py` file.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -215,6 +209,11 @@ def visit_boolop(self, node):
         else:
             node.type_constraints = TypeInfo(Any)
 
+    def visit_compare(self, node):
+        """Comparison operators are: '<', '>', '==', '<=', '>=', '!=', 'in', 'is'.
+        All comparisons yield boolean values."""
+        node.type_constraints = TypeInfo(bool)
```
david-yz-liu: You should still look up the function calls and unify them properly.

python_ta/typecheck/base.py
```diff
@@ -263,7 +265,13 @@ def can_unify(self, t1, t2):
                         return False
                 return True
             else:
-                False
+                return False
+        elif isinstance(t1, GenericMeta):
```
david-yz-liu: I'm not sure why these are necessary. Won't they be caught by the `t1 != t2` below?

python_ta/transforms/type_inference_visitor.py
```diff
@@ -212,7 +212,19 @@ def visit_boolop(self, node):
     def visit_compare(self, node):
         """Comparison operators are: '<', '>', '==', '<=', '>=', '!=', 'in', 'is'.
         All comparisons yield boolean values."""
-        node.type_constraints = TypeInfo(bool)
+        return_types = set({})
```
david-yz-liu: You can just do `= {}`.

RyanDJLee:
- Was unable to change line 215 in type_inference_visitor.py as Python3 seems to interpret {} as a default and raises Exception stating dictionaries do not have .add attributes.

david-yz-liu:
@RyanDJLee this is good. For `is` and `is not`, since these aren't special methods, you can have that be a special case in `visit_compare` which doesn't type check the arguments (instead, just `return_types.add(bool)`).

Omit `in` (and `not in`) from the tests entirely, and leave a `TODO` in `visit_compare`; this should be done in a separate pull request, since it involves other special methods.

Finally, good point about `{}`. Use `set()`, not `set({})`.

PR #251, opened 2017-06-14
Fix for edge cases.

nigef:
These changes fix the ending locations -- to capture the parens not included in the Call syntax -- for the edge case: `fun(((1), ))`

A breaking change (lines 433-434) was made to set the correct locations for Tuple node in the example, `(x, y) = [7, 8]`. As an alternative, a new transform function could be made.

The 6 breaking changes (below) could be corrected by passing in a binary flag to determine whether to return when whitespace (`CONSUMABLES`) encountered.

AssignAttr, Attribute
```self . name```

DelAttr
```del self.<...>```

ExtSlice, Slice, Subscript
```A [ 1, 2 :   ]```

### CODE REVIEW

File: python_ta/transforms/setendings.py [hunk 1]
```diff
@@ -430,6 +430,8 @@ def set_endings_from_source(node):
                 temp = node.end_col_offset
                 node.end_col_offset = j + 1
                 return
+            elif source_code[lineno][j] not in CONSUMABLES:
```
nigef: Adding this line does cause the last failing test for Tuple node to now pass.
But it causes 6 other nodes to break. 

Pushed this change (lines 433-434) to this PR more to talk about the approach first.
I think this change should be reverted, and instead, a new transform function should be made for capturing ending `,` in Tuple node?

PR #252, opened 2017-06-15
Dev test for

RyanDJLee:
- Implement visitor for For node.
- Add test cases to test inference of bound names in for loop statement as well as scope setting of locally bound target names.

### DIFF
--- a/nodes/For.py
+++ b/nodes/For.py
@@ -7,9 +7,8 @@ Attributes:
     - target  (Node)
         - Holds the variable(s) the loop assigns to as a single node. The type
           of the node can be Name, List, Tuple, etc.
-    - iter    (Call)
-        - A function call node which represents the part that iterates over
-          the loop.
+    - iter    (List Node)
+        - A List node representing the iterable.
     - body    (List[Statement])
         - The node to be executed.
     - orelse  (List[Statement])
@@ -18,12 +17,13 @@ Attributes:
 
 Example:
     - target  -> AssignName(i)
-    - iter    -> Call(func = Name(range), args = [3])
+    - iter    -> List(Const.int(Value=1), Const.int(Value=2),
+    Const.int(Value=3)))
     - body    -> [Break()]
     - orelse  -> [Pass()]
 """
 
-for i in range(3):
+for i in [1, 2, 3]:
     break
 else:
     pass
--- a/nodes/For.py
+++ b/nodes/For.py
@@ -7,8 +7,9 @@ Attributes:
     - target  (Node)
         - Holds the variable(s) the loop assigns to as a single node. The type
           of the node can be Name, List, Tuple, etc.
-    - iter    (List Node)
+    - iter    (Node)
         - A List node representing the iterable.
+        - Can also be a Call node (ie. representing a range function call).
     - body    (List[Statement])
         - The node to be executed.
     - orelse  (List[Statement])
@@ -27,3 +28,7 @@ for i in [1, 2, 3]:
     break
 else:
     pass
+
+# TODO: Add example of for-else loop iterating over generator returned by range
+# call."
+
--- a/nodes/For.py
+++ b/nodes/For.py
@@ -7,9 +7,9 @@ Attributes:
     - target  (Node)
         - Holds the variable(s) the loop assigns to as a single node. The type
           of the node can be Name, List, Tuple, etc.
-    - iter    (Node)
-        - A List node representing the iterable.
-        - Can also be a Call node (ie. representing a range function call).
+    - iter    (Call)
+        - A function call node which represents the part that iterates over
+          the loop.
     - body    (List[Statement])
         - The node to be executed.
     - orelse  (List[Statement])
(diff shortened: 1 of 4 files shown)


RyanDJLee:
Will update test cases using hypothesis after fixing Assign node visitor.

nodes/For.py
```diff
@@ -18,12 +18,19 @@
 
 Example:
     - target  -> AssignName(i)
-    - iter    -> Call(func = Name(range), args = [3])
+    - iter    -> List(Const.int(Value=1), Const.int(Value=2),
```
david-yz-liu: This is an okay change, but the *real* change should be made in the documentation of `iter` above, which incorrectly says that this must be a `Call` node. Really it's just a node that represents the value being iterated over - the part after the `in`.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -276,6 +276,18 @@ def visit_call(self, node):
...
+    def visit_for(self, node):
+        for_node = list(node.nodes_of_class(astroid.For))[0]
+        iterable_type = self.type_store.lookup_function('__iter__', for_node.iter.type_constraints.type)
+        rtype = self.type_constraints.unify_call(iterable_type, for_node.iter.type_constraints.type)
+        # there may be one target, or a Generic of targets to unify.
+        if isinstance(for_node.target, astroid.AssignName):
+            self.type_constraints.unify(rtype.__args__[0], node.frame().type_environment.lookup_in_env(for_node.target.name))
+        else:
+            target_tvars = [node.frame().type_environment.lookup_in_env(target_node.name) for target_node in for_node.target.elts]
+            for i in range(len(rtype.__args__) + 1):
```
david-yz-liu: This loop seems off. The counter range is based on `rtype.__args__`, but the value being indexed is `target_tvars`, whose length is independent of `rtype.__args__`.

python_ta/typecheck/base.py
```diff
@@ -268,6 +268,10 @@ def can_unify(self, t1, t2):
                 return True
             else:
                 return False
+        elif isinstance(t1, GenericMeta):
```
david-yz-liu: I think I already commented on this change in a previous pull request. You should be able to remove this.

tests/type_inference/test_for.py
```diff
@@ -0,0 +1,48 @@
...
+from typing import Callable
+import tests.custom_hypothesis_support as cs
+settings.load_profile("pyta")
+
+
+def test_for_list():
+    """Test whether visitors properly set the type constraint of the a For node representing for/else statement
+     iterating over a simple list.
+    """
+    program = f'for num in [1, 2, 3]:\n' \
```
david-yz-liu: It would be nice to have this be a hypothesis-generated list.

tests/type_inference/test_for.py
```diff
@@ -0,0 +1,48 @@
...
+              f'    x = 3 + num\n' \
+              f'else:\n' \
```
david-yz-liu: You don't need to include the `else` here (and in the example `nodes` documentation). That's not important here.

tests/type_inference/test_for.py
```diff
@@ -0,0 +1,48 @@
...
+    local_type_var = module.type_environment.lookup_in_env('x')
+    inferred_type = TypeInferrer.type_constraints.lookup_concrete(local_type_var)
+    assert inferred_type == int
+
+
+def test_inference_func_def_for():
+    """Test whether visitors properly set the type constraint of the a For node representing for/else statement
+     iterating over a more complex iterable (ie, tuples, dicts, nested iterables).
+    """
+    program = f'x = 5\n' \
```
david-yz-liu: There's too much stuff going on with this test. Just keep the function part, and remove the rest.

python_ta/typecheck/base.py
RyanDJLee: I originally put this in because isinstance raises an error when called on parametrized generics. I ran into the same problem with issubclass so I decided to put the check back in, which seems to handle the problem.

david-yz-liu: Hmm okay.

tests/type_inference/test_for.py
```diff
@@ -0,0 +1,56 @@
...
+    """Test whether visitors properly set the type constraint of the a For node representing for/else statement
+     iterating over a homogeneous list.
+    """
+    program = f'for elt in {iterable}:\n' \
+              f'    x = elt\n'
+    module, TypeInferrer = cs._parse_text(program)
+    for_node = list(module.nodes_of_class(astroid.For))[0]
+    local_type_var = module.type_environment.lookup_in_env('x')
+    inferred_type = TypeInferrer.type_constraints.lookup_concrete(local_type_var)
+    assert inferred_type == type(iterable[0])
```
david-yz-liu: Remember what I said about not calling `type`.

RyanDJLee: Sorry T_T

PR #253, opened 2017-06-15
Add test case for multi-target-assignment with a List.

RyanDJLee:


python_ta/transforms/type_inference_visitor.py
```diff
@@ -234,10 +234,18 @@ def visit_compare(self, node):
     def visit_assign(self, node):
         # multi-assignment; LHS is a tuple of AssignName target nodes as "elements"
         if isinstance(node.targets[0], astroid.Tuple):
-            target_type_tuple = zip(node.targets[0].elts, node.value.elts)
-            for target_node, value in target_type_tuple:
-                target_type_var = node.frame().type_environment.lookup_in_env(target_node.name)
-                self.type_constraints.unify(target_type_var, value.type_constraints.type)
+            if isinstance(node.value, astroid.Name):
```
david-yz-liu: This is not a good choice of special case.

The real special case should be `isinstance(node.value, astroid.Tuple)`. For everything else, lookup the type and then unify_call with `__iter__`.

tests/type_inference/test_assignments.py
```diff
@@ -89,5 +89,40 @@ def test_set_multi_assign(variables_list, value):
...
+        assume(not iskeyword(variable_name))
+    program = ("x = ["
+               + ", ".join([repr(value) for value in variables_dict.values()])
+               + "]\n"
+               + ", ".join(variables_dict.keys())
+               + " = x")
+    module, typeinferrer = cs._parse_text(program)
+    for variable_name in variables_dict:
+        var_type_var = module.type_environment.lookup_in_env(variable_name)
+        assert typeinferrer.type_constraints.lookup_concrete(var_type_var) == type(list(variables_dict.values())[0])
```
david-yz-liu: Don't use `type` here.

PR #254, opened 2017-06-16
Fix indentation. 

nigef:
This went unnoticed 4 months ago. As a bonus, this change improves runtime, noticeable when running nosetests.


david-yz-liu:
Nice one!

PR #264, opened 2017-06-19
Dev test ifexpr

RyanDJLee:
- Update documentation for IfExpr node.
- Implement IfExpr node visitor.
- Add test cases for IfExpr node representing an assignment of an if expression to a variable with both body/orelse suites having the same type as well as different types.

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -303,7 +303,8 @@ class TypeInferer:
                 self.type_constraints.unify(rtype.__args__[0], target_tvars[i])
 
     def visit_ifexp(self, node):
-        if node.body.type_constraints.type == node.orelse.type_constraints.type:
+        if self.type_constraints.can_unify(node.body.type_constraints.type, node.orelse.type_constraints.type):
+            self.type_constraints.unify(node.body.type_constraints.type, node.orelse.type_constraints.type)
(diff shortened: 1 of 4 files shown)


RyanDJLee:
- Will update using hypothesis testing.

RyanDJLee:
- Build is breaking due to name change in custom_hypothesis_support; fixed issue in separate branch.

david-yz-liu:
@RyanDJLee you should fix the error in this branch.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -302,6 +302,12 @@ def visit_for(self, node):
             for i in range(len(target_tvars)):
                 self.type_constraints.unify(rtype.__args__[0], target_tvars[i])
 
+    def visit_ifexp(self, node):
+        if node.body.type_constraints.type == node.orelse.type_constraints.type:
```
david-yz-liu: There should be a `unify` here.

tests/custom_hypothesis_support.py
```diff
@@ -36,6 +36,14 @@
 numeric_values = numeric_types.flatmap(lambda s: s())
 
 
+eval_types = hs.sampled_from([
```
david-yz-liu: Another strategy is probably not necessary here.

PR #269, opened 2017-06-20
Refactor visitors

RyanDJLee:
- Refactor visitors.
- Implement helper function that looks up a function, unifies the function call with given arguments and returns a return type wrapped in the TypeInfo wrapper.
    - Write more informative TypeErrorInfo messages in helper.
- Reimplement Name, BinOp, UnaryOp, Subscript, Assign, For node visitors using helper.


david-yz-liu:
@RyanDJLee please work on separating the changes here from the ones from `IfExp`.

PR #278, opened 2017-06-30
Reimplement iterables visitors

RyanDJLee:
- Implement helper to return least general unifier of two types.
- Update List, Dict, Set node visitors to use helper and set type constraints.
- Update test cases assumptions (bool and int unified).

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -133,31 +133,32 @@ class TypeInferer:
             node.type_constraints = TypeInfo(NoType)
 
     def visit_list(self, node):
-        # node_types contains types of elements inside list.
-        node_type = node.elts[0].type_constraints.type
-        for elt in node.elts:
-            node_type = self.type_constraints.least_general_unifier(elt.type_constraints.type, node_type)
-        node.type_constraints = TypeInfo(List[node_type])
-
-    def visit_set(self, node):
-        node_type = node.elts[0].type_constraints.type
-        for elt in node.elts:
-            node_type = self.type_constraints.least_general_unifier(elt.type_constraints.type, node_type)
-        node.type_constraints = TypeInfo(List[node_type])
-
-    def visit_set(self, node):
-        node_types = {node_child.type_constraints.type for node_child in node.elts}
-        if len(node_types) == 1:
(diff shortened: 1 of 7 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -102,28 +102,23 @@ def visit_tuple(self, node):
...
-        node_types = {node_child.type_constraints.type for node_child in node.elts}
-        # If list has more than one type, just set node.type_constraints to List.
-        # If list has only one type T, set the node.type_constraints to be List[T].
-        if len(node_types) == 1:
-            # node_types.pop() returns the only element in the set, which is a
-            # type object.
-            node.type_constraints = TypeInfo(List[node_types.pop()])
-        else:
-            node.type_constraints = TypeInfo(List[Any])
+        node_type = node.elts[0].type_constraints.type
```
david-yz-liu: This is good except that it will fail on an empty list. This is a bit trickier, so for now you can do what the old code did and set the type constraint to `List[Any]` in that case.

Similar for the other collection types.

python_ta/typecheck/base.py
```diff
@@ -217,6 +217,71 @@ def unify_call(self, func_type, *arg_types):
...
+        elif isinstance(t1, TypeVar):
+            i1 = self._find(t1)
+            if issubclass(i1, t2):
+                return t2
+            elif issubclass(t2, i1):
+                return i1
+            else:
+                return Any
+        elif isinstance(t2, TypeVar):
+            self.least_general_unifier(t2, t1)
```
david-yz-liu: Many parts of your code here are weird because there aren't any return statements. The whole point of this method is to return something.

tests/test_type_inference/test_literals.py
```diff
@@ -33,6 +33,10 @@ def test_homogeneous_lists(lst):
 def test_random_lists(lst):
     """Test List nodes representing a list of values of different primitive types."""
     assume(not isinstance(lst[0], type(lst[1])))
+    # bools and ints become unified; assume we have one or the other for this test case.
```
david-yz-liu: This check is incorrect. Fix the `assume` above so that `lst[1]` is not an instance of `type(lst[0])` either.

Same comment below. Don't use `==` to compare types.

PR #283, opened 2017-07-05
Week 6

ostrokach:
Goes with pyta-uoft/website#60


examples/pylint/C0121_singleton_comparison.py
```diff
@@ -1,9 +1,5 @@
-def is_true():
-    """
-    @rtype: bool
-    """
-    temp = 5
-    if temp == None:
-        return False
+def square(number):
```
david-yz-liu: Don't need to wrap this in a function. (Generally prefer module-level examples.)

PR #287, opened 2017-07-14
Dev test method call

RyanDJLee:
- Reimplement Call node visitor to handle method calls.
- RyanDJLee | Implement Attribute node visitor representing an attribute access.
- RyanDJLee | Do concrete lookups when assigning types.
- RyanDJLee | Implement test case for method calls.

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -37,6 +37,7 @@ class TypeInferer:
     """
     type_constraints = TypeConstraints()
     type_store = TypeStore(type_constraints)
+    type_environments = {}
 
     def __init__(self):
         self.type_constraints.clear_tvars()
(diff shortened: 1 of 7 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -329,19 +338,27 @@ def visit_functiondef(self, node):
...
-                node.frame().locals[func_name][0].type_environment.locals['__init__'])
-            arg_types = [_ForwardRef(func_name)] + [arg.type_constraints.type for arg in node.args]
-            self.type_constraints.unify_call(func_t, *arg_types)
-            node.type_constraints = TypeInfo(_ForwardRef(func_name))
-        else:
-            func_t = self.type_constraints.lookup_concrete(node.frame().type_environment.locals[func_name])
-            arg_types = [arg.type_constraints.type for arg in node.args]
+        try:
+            func_name = node.func.name
+        except AttributeError:  # Method Call
```
david-yz-liu: This is not a good way to check this. Check the type of `node.func` explicitly.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -196,6 +196,12 @@ def _closest_frame(self, node, name):
 
     def visit_name(self, node):
         try:
+            node.type_constraints = TypeInfo((self.type_constraints.lookup_concrete(
+                self._closest_frame(node, node.name)).type_environment.lookup_in_env(node.name)))
+        except KeyError:
```
david-yz-liu: I don't know what's going on here exactly - possibly a bad merge. This `except` clause catches the same error as the following one.

PR #292, opened 2017-07-20
Dev test method call annotations.

RyanDJLee:
- Add test case for checking types of annotated method call.
- Reimplement annotation parser to handle instance method definitions.

### DIFF
--- a/python_ta/typecheck/base.py
+++ b/python_ta/typecheck/base.py
@@ -448,7 +448,11 @@ def parse_annotations(node, class_tvars=None):
     """Return a type specified by the type annotations for a node."""
     if isinstance(node, astroid.FunctionDef):
         arg_types = []
-        if class_tvars is None or not isinstance(node.parent, astroid.ClassDef):
+        arg_annotation_list = list(zip(node.args.args, node.args.annotations))
+        # Special case; Instance method definition based on syntactical conventions.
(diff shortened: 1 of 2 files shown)


python_ta/typecheck/base.py
```diff
@@ -448,8 +448,14 @@ def parse_annotations(node, class_tvars=None):
     """Return a type specified by the type annotations for a node."""
     if isinstance(node, astroid.FunctionDef):
         arg_types = []
+        arg_annotation_list = list(zip(node.args.args, node.args.annotations))
+        # Special case; Instance method definition based on syntactical conventions.
         if class_tvars is None or not isinstance(node.parent, astroid.ClassDef):
-            self_type = None
+            if isinstance(node.parent, astroid.ClassDef) and getattr(
```
david-yz-liu: The nested `if` structure here isn't great. Refactor this code so that the three cases are explicit in one `if-elif-else` chain. Also remove `arg_annotation_list` - its only use here does not warrant its existence.

tests/test_type_inference/test_classdef.py
```diff
@@ -46,5 +46,30 @@ def test_classdef_method_call():
...
+    program = f'class Network:\n' \
+              f'    def __init__(self, name: str) -> None:\n' \
+              f'        self.name = name\n' \
+              f'        status = 0\n' \
+              f'    def set_status(self, status: int) -> int:\n' \
+              f'        self.status = status\n' \
+              f'        return self.status\n' \
+              f'\n'
+    module, inferer = cs._parse_text(program)
+    # for self parameter
```
david-yz-liu: This comment is either unnecessary or misplaced and too vague.

PR #293, opened 2017-07-20
Refactor code lookups

RyanDJLee:
- Implement helper to facilitate type lookups.
- Update visitors accordingly.

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -122,15 +122,15 @@ class TypeInferer:
     ##############################################################################
     # Literals
     ##############################################################################
-    def _find_attribute_type(self, node, instance_name, attribute_name):
+    def _lookup_attribute_type(self, node, instance_name, attribute_name):
         """Given the node, class name and attribute name, return the type of the attribute."""
-        class_type = self._find_type(node, instance_name)
+        class_type = self.lookup_type(node, instance_name)
         class_name = class_type.__forward_arg__
         class_env = self._closest_frame(node, class_name).locals[class_name][0].type_environment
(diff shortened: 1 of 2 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -124,12 +124,16 @@ def type_inference_transformer(self) -> TransformVisitor:
...
     def _find_attribute_type(self, node, instance_name, attribute_name):
         """Given the node, class name and attribute name, return the type of the attribute."""
-        class_type = self.type_constraints.lookup_concrete(self._closest_frame(node, instance_name)
-                                                           .type_environment.lookup_in_env(instance_name))
+        class_type = self._find_type(node, instance_name)
         class_name = class_type.__forward_arg__
         class_env = self._closest_frame(node, class_name).locals[class_name][0].type_environment
         return self.type_constraints.lookup_concrete(class_env.lookup_in_env(attribute_name))
 
+    def _find_type(self, node, name):
```
david-yz-liu: Mark this as public. It should be used broadly in test helpers.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -124,12 +124,16 @@ def type_inference_transformer(self) -> TransformVisitor:
...
+        """Given a variable name, return it's concrete type."""
```
david-yz-liu: You used the wrong "its". Also explain the role of `node` here.

python_ta/transforms/type_inference_visitor.py
david-yz-liu: Also it's better to be consistent with verb choice - use `lookup` instead of `find`.

PR #297, opened 2017-07-23
Syntax errors / Miscellaneous / Custom errors

ostrokach:
Goes together with pyta-uoft/website#65.

Includes commits from pyta-uoft/pyta#295 and pyta-uoft/pyta#296.

### DIFF
--- a/examples/dynamic_execution_example.py
+++ b/examples/dynamic_execution_example.py
@@ -1,7 +1,9 @@
-def dynamic_execution(name):
-    """Append in front of string 'name'"""
-    eval('hello, my name is ' + name)
+def get_power_function(power):
+    """Return a function which takes its argument to the <power>th power."""
+    func = eval("lambda x: x ** {}".format(power))
+    return func
+
 
 if __name__ == '__main__':
-    dynamic_execution('hayley')
-
+    power_func = get_power_function(10)
+    print(power_func(2))  # Prints 1024
--- a/examples/dynamic_execution_example.py
+++ b/examples/dynamic_execution_example.py
@@ -4,6 +4,5 @@ def get_power_function(power):
     return func
 
 
-if __name__ == '__main__':
-    power_func = get_power_function(10)
-    print(power_func(2))  # Prints 1024
+power_func = get_power_function(10)
+print(power_func(2))  # Prints 1024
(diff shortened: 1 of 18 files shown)


examples/dynamic_execution_example.py
```diff
@@ -1,7 +1,9 @@
...
-    eval('hello, my name is ' + name)
+def get_power_function(power):
+    """Return a function which takes its argument to the <power>th power."""
+    func = eval("lambda x: x ** {}".format(power))
+    return func
 
-if __name__ == '__main__':
-    dynamic_execution('hayley')
 
+if __name__ == '__main__':
```
david-yz-liu: The changed example is good. But in this case, I don't need to see the main block at all - the error should be detected just in the function definition.

examples/pylint/W1401_anomalous_backslash_in_string.py
```diff
@@ -1 +1,2 @@
-print('\d{3}')  # Error on this line
+import re
+re.findall('(\d+)\D*', '123 Main St.')  # Error on this line
```
david-yz-liu: I don't like this example. Regular expressions are not typically covered in CSC108/148. Just do something even simpler than the original example, like `'This is a bad escape: \h'`.

examples/syntax_errors/assignment_to_literal.py
```diff
@@ -0,0 +1,3 @@
+a = 12
+12 = a  # Error on this line: can't assign to literal
+"hello" = a  # Error on this line: can't assign to literal
```
david-yz-liu: Good, but please use single-quotes for strings in general.

examples/syntax_errors/missing_quote.py
```diff
@@ -0,0 +1 @@
+print('Hello!)  # Error on this line: missing closing quote (`'`)
```
david-yz-liu: I think the backticks are actually confusing here. Please remove them.

david-yz-liu:
@ostrokach note that I reviewed all changes here (which is fine) so I'm closing the other two PRs.

david-yz-liu:
Also do merge from master, and you'll get the fix for the failing tests.

PR #300, opened 2017-07-26
Augment type constraints

RyanDJLee:
- Change structure of TypeConstraints; sets are now composed of tuples of TypeVar and the corresponding Astroid Node that was being visited during unification of the TypeVar.
- Reimplement the relevant operations on the sets of the TypeConstraints; _find, fresh_tvar, create_in_env, unify, can_unify
- Update all calls to changed functions.

--------------------------------------------------

- Decided to refactor the code for AnnAssign nodes in this branch as well as there was a lot of refactoring work to be done of the same nature.

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -58,7 +58,7 @@ class TypeInferer:
     def _set_module_environment(self, node):
         """Method to set environment of a Module node."""
         node.type_environment = Environment(
-            globals_={name: self.type_constraints.fresh_tvar() for name in node.globals})
+            globals_={name: self.type_constraints.fresh_tvar(node) for name in node.globals})
         self._populate_local_env(node)
 
     def _set_classdef_environment(self, node):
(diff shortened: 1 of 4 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -189,6 +189,11 @@ def visit_expr(self, node):
     def _closest_frame(self, node, name):
         """Helper method to find the closest ancestor node containing name relative to the given node."""
         closest_scope = node
+        if hasattr(closest_scope, 'type_environment') and (
+                            name in closest_scope.type_environment.locals or
```
david-yz-liu: The vertical alignment here is incorrect.

python_ta/typecheck/base.py
```diff
@@ -140,35 +140,38 @@ def clear_tvars(self):
...
+                        return i, tuple(_tuple)[0]
+                    else:
+                        return i, tuple(_tuple)[1]
+        return -1, None
+
+    def unify(self, t1, t2, node):
         if isinstance(t1, TypeVar) and isinstance(t2, TypeVar):
-            i1 = self._find(t1)
-            i2 = self._find(t2)
+            i1, n1 = self._find(t1)
```
david-yz-liu: Use underscores when you aren't actually going to use the variable.

PR #301, opened 2017-07-28
Restructure type constraints

RyanDJLee:
- Add Disjoint sets operations.
- Restructure sets of TypeConstraints to be Disjoint sets (concrete types as set representative)
- Add dictionary to TypeConstraints for quick lookup.
- Reimplement unification algorithm to work with Disjoint sets structure.
- Update all calls to unify accordingly (Reverted back to original interface)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -321,48 +325,46 @@ def visit_return(self, node):
...
             # Check whether this is a method in a class
             if isinstance(node.parent, astroid.ClassDef) and isinstance(arg_types[0], TypeVar):
-                self.type_constraints.unify(arg_types[0], _ForwardRef(node.parent.name))
+                self.type_constraints.unify(arg_types[0], _ForwardRef(node.parent.name), node)
 
             # check if return nodes exist; there is a return statement in function body.
             if len(list(node.nodes_of_class(astroid.Return))) == 0:
                 func_type = Callable[arg_types, None]
             else:
+                polymorphic_tvars = [arg for arg in arg_types if isinstance(arg, TypeVar)]
```
david-yz-liu: This part should really be done for both branches of the `if`. The `create_Callable` is a good change to make, and should supersede the code in the `if` branch.

python_ta/typecheck/base.py
```diff
@@ -129,42 +133,83 @@ def lookup_method(name, caller_type, *args):
     return TYPE_SIGNATURES[caller_origin][name]
 
 
+def make_set(node):
```
david-yz-liu: These should all be methods in `TypeConstraints`. Also `make_set` should not take in a node, but an item. It should instantiate a `TNode` given the item.

python_ta/typecheck/base.py
```diff
@@ -129,42 +133,83 @@ def lookup_method(name, caller_type, *args):
...
+    node.parent = node
```
david-yz-liu: New nodes should have their `parent` set to `None`. This should also be done in the constructor of `TNode`. Never add new attributes outside of a constructor.

python_ta/typecheck/base.py
```diff
@@ -129,42 +133,83 @@ def lookup_method(name, caller_type, *args):
...
+    rep2 = _find_rep(node2)
+    rep2.parent = rep1
+
+
+class TNode:
+    def __init__(self, node_type, origin_node):
+        self.type = node_type
+        self.origin = origin_node
+
+    def __str__(self):
```
david-yz-liu: This is unnecessary.

python_ta/typecheck/base.py
```diff
@@ -129,42 +133,83 @@ def lookup_method(name, caller_type, *args):
...
+def _find_rep(node):
+    while node.parent != node:
+        node = node.parent
+    return node
+
+
+def _union(node1, node2):
+    rep1 = _find_rep(node1)
+    rep2 = _find_rep(node2)
+    rep2.parent = rep1
```
david-yz-liu: Let's make this a bit better. Pick as parent the node that has a *concrete* type, if it exists.

python_ta/typecheck/base.py
```diff
@@ -129,42 +133,83 @@ def lookup_method(name, caller_type, *args):
...
         self._count += 1
         return tvar
 
-    def _find(self, t: TypeVar) -> int:
-        """Return the index of the set containing t."""
-        for i, type_set in enumerate(self._sets):
-            if t in type_set:
-                return i
-        return -1
+    def add_concrete_to_sets(self, _type):
```
david-yz-liu: While it is appropriate to create a node for a concrete type, you don't need a separate helper fir it. In particular, don't change any of the other instance variables here.

python_ta/typecheck/base.py
```diff
@@ -129,42 +133,83 @@ def lookup_method(name, caller_type, *args):
...
+        """Add a concrete type to the type constraints sets."""
+        tnode = TNode(_type, None)
+        self._sets.append(tnode)
+        make_set(tnode)
+        self._tvar_tnode[_type] = tnode
```
david-yz-liu: Basically your logic for this attribute is incorrect. Its keys should *only* be for  type variables.

python_ta/typecheck/base.py
```diff
@@ -129,42 +133,83 @@ def lookup_method(name, caller_type, *args):
...
+        self._count += 1
+        return tnode
 
     def unify(self, t1, t2):
+        try:
+            node1 = self._tvar_tnode[t1]
```
david-yz-liu: Based on my previous comment, this should only occur when `t1` is a type variable.

python_ta/typecheck/base.py
```diff
@@ -185,17 +230,17 @@ def unify(self, t1, t2):
         elif issubclass(t1, t2) or issubclass(t2, t1):
             pass
         elif t1 != t2:
-            raise Exception(str(t1) + ' ' + str(t2))
+            raise BadUnificationError(str(t1) + ' ' + str(t2))
 
-    def _unify_generic(self, t1: GenericMeta, t2: GenericMeta):
-        """Unify two generic types."""
+    def _unify_generic(self, t1, t2):
```
david-yz-liu: It is regression to remove a type signature without replacing it with an updated one. Don't do this.

python_ta/typecheck/base.py
```diff
@@ -303,24 +348,24 @@ def lookup_concrete(self, tvar):
...
-
-        if isinstance(rep, CallableMeta):
-            return _gorg(rep)[[self.lookup_concrete(t1) for t1 in rep.__args__[:-1]],
-                              self.lookup_concrete(rep.__args__[-1])]
-        elif isinstance(rep, GenericMeta):
-            return _gorg(rep)[tuple(self.lookup_concrete(t1) for t1 in rep.__args__)]
-        return rep or tvar
+        tnode = self._tvar_tnode[tvar]
+        return _find_rep(tnode).type
+        #
```
david-yz-liu: Don't leave commented blocks.

tests/test_type_inference/test_classdef.py
```diff
@@ -41,7 +41,8 @@ def test_classdef_method_call():
     module, inferer = cs._parse_text(program)
     attribute_node = list(module.nodes_of_class(astroid.Attribute))[1]
     expected_rtype = attribute_node.parent.type_constraints.type
-    assert attribute_node.type_constraints.type.__args__[-1] == expected_rtype
+    actual_rtype = inferer.type_constraints.lookup_concrete(attribute_node.type_constraints.type.__args__[-1])
```
david-yz-liu: I don't know why you aren't using `lookup_type` here.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -275,7 +280,7 @@ def visit_compare(self, node):
                                                             left_value.type_constraints.type,
                                                             right_value.type_constraints.type)
                 left_value = right_value
-                return_types.add(self.type_constraints.unify_call(function_type, left_value.type_constraints.type,
+                return_types.add(self.type_constraints.unify_call(node, function_type, left_value.type_constraints.type,
```
RyanDJLee: I cannot do this because I am using positional arguments in the parameters of unify_call.

tests/test_type_inference/test_classdef.py
RyanDJLee: lookup_type returns the type of a variable in the closest enclosing scope given a Node and a name representing the identifier name; in this case we are doing a lookup of a type variable to see if that type variable has been properly unified in the environment.

python_ta/typecheck/base.py
RyanDJLee: Unify is not be guaranteed to take in TypeVar instances as arguments as you mentioned above, so I need to be able to create TNodes representing the concrete types in the sets and have a way to look them up. I agree that these nodes should not be stored in our _tvars_tnode dictionary so I decided to create a new private dictionary attribute in TypeConstraints to store the concrete types and their corresponding TNodes. This requires a helper because I do not want to raise the counts attribute while creating these concrete type Nodes and I also do not want to add them to the tvars dictionary.

PR #307, opened 2017-08-08
Global variables checker dev

nigef:
The `global_variables_checker` would think variables inside a list comprehension are always global -- these variables should never be seen as global, purely from being in a list comp.

### DIFF
--- a/python_ta/checkers/global_variables_checker.py
+++ b/python_ta/checkers/global_variables_checker.py
@@ -76,7 +76,7 @@ def _get_child_disallowed_global_var_nodes(node):
     global, non-constant variable. 
     """
     node_list = []
-    scoped_comps = (astroid.ListComp, astroid.Comprehension, astroid.DictComp, astroid.SetComp)
+    scoped_comps = (astroid.ListComp, astroid.Comprehension, astroid.DictComp, astroid.SetComp, astroid.GeneratorExp)
     if isinstance(node.parent, scoped_comps):
         return node_list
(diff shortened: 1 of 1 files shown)


python_ta/checkers/global_variables_checker.py
```diff
@@ -76,6 +76,8 @@ def _get_child_disallowed_global_var_nodes(node):
     global, non-constant variable. 
     """
     node_list = []
+    if isinstance(node.parent, astroid.ListComp) or isinstance(node.parent, astroid.Comprehension):
```
david-yz-liu: Two comments here:

1. If `ListComp` is necessary, then there should also be a check for `DictComp` and `SetComp` (there's a symmetry here).
2. But I think `node.parent` is the wrong approach (doesn't take into account nested expressions). I'd like to see you use `lookup_scope` (or whatever that method was called) to do this check.

nigef: Thanks, I've added the checks for additional nodes.
Nice point about the nesting, although I can't think of a case where this doesn't work on nested expressions, so maybe the current approach is acceptable?

david-yz-liu: About the nesting: something like `[x + 3 for x in ...]` would have the `x` in `x+3` as a child of a `BinOp` node, right? If I understand this code properly, in this case the `x` in `x + 3` would be reported.

nigef: Fixed

PR #330, opened 2018-03-31
Changed string format style and removed unnecessary format

pomonam:
https://stackoverflow.com/questions/40714555/python-string-formatting-is-more-efficient-than-format-function


david-yz-liu:
Hello @pomonam! Thanks for your contribution, and sorry for the delay in getting to this. Some parts of PythonTA already use f-strings, so if you want to submit a change to string formatting, I recommend using them instead.

PR #336, opened 2018-05-07
Error messages unaryop

RyanDJLee:
- Add section for unary operation error messages.
- Add test case for a built-in unary operator being called on an argument of the wrong type.


david-yz-liu:
@RyanDJLee it seems like the test you've added doesn't contain any asserts. Also the `TODO` you added should go in the README, not in the code, since it's pretty general.

PR #342, opened 2018-05-09
Type inf visitor refactor

justinpark715:
Made changes to transforms/type_inference_visitor.py to correctly use new version of unify in base.py
Made changes to test cases to  correctly use new version of unify
Changed unify_call in base.py to return TypeResult


justinpark715:
Also skipped some tests in test_functions because they were dependent on specific message formatting

PR #343, opened 2018-05-10
Print ast from mod

justinpark715:
Adding functionality for printing Astroid trees in readable format for testing purposes. Based on `sample_usage/print_ast.py`, but intended to be used within python code, as opposed to being run as a program. 

use by including `from tests.print_ast_from_mod import print_ast`, and calling `print_ast()`


david-yz-liu:
It's fine to have this functionality be importable as a function, but this level of code duplication is not appropriate. You can take this idea and make it better by removing this duplication.

PR #344, opened 2018-05-10

simeonkr:


simeonkr:
I think the fact that the tests fail is due to the current signature of __getitem__() in typeshed. I am not sure if in general you can have slices like a["abc":"def"] (where 'a' is some exotic object with an overriden __getitem__), but I don't see why not. If you can only have integer slices, then this method would be simpler.

PR #347, opened 2018-05-10

simeonkr:
There is a failing testcase, but I think this is due to the signature of \_\_getitem\_\_ in typeshed (\_\_getitem\_\_ returns a list in the case of a slice, so has a different signature in this case).
Also, I'm assuming that slice indices are integers only. I think this is the case because a 'slice' object is like a wrapper around a 'range' object.


david-yz-liu:
Two comments:

1. It's better to do a `unify_call` against the *slice constructor* (note that this is in typeshed).
2. The test shouldn't fail. Your explanation for the current problem is not sufficient. Note that typeshed has two `__getitem__` signatures under `list`.

PR #349, opened 2018-05-13
typecheck: Fix call in visit_subscript & add tests

simeonkr:
I have assumed, as before, that inferred type for a Subscript node within an Assign node (ctx=Store) is NoneType (the return type of \_\_setitem\_\_), but I'm not sure if this is the case. It might make more sense if the whole Assign node is handled as a call to \_\_setitem\_\_ so that Assign.value can be passed as the last argument of \_\_setitem\_\_ (and this also fits in with Assign nodes being of type NoneType). But this means treating this case as a special case in visit_assign. The inferred type of the Subscript node itself would be the type of Assign.value (or a list around that type in case of a slice).

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -253,6 +253,10 @@ class TypeInferer:
                         target_tvar = self.lookup_type(subtarget, subtarget.name)
                         self.type_constraints.unify(target_tvar, contained_type)
                 return iter_type_result
+        elif isinstance(target, astroid.Subscript):
+            # TODO: previous case must recursively handle this one
+            return self._handle_call(target, '__setitem__', target.value.inf_type.getValue(),
+                                     target.slice.inf_type.getValue(), expr_type)
 
     def _lookup_attribute_type(self, node, instance_name, attribute_name):
         """Given the node, class name and attribute name, return the type of the attribute."""
@@ -376,13 +380,12 @@ class TypeInferer:
     def visit_subscript(self, node: astroid.Subscript) -> None:
         if node.ctx == astroid.Load:
             node.inf_type = self._handle_call(node, '__getitem__', node.value.inf_type.getValue(),
-                                                      node.slice.inf_type.getValue())
+                                              node.slice.inf_type.getValue())
         elif node.ctx == astroid.Store:
-            node.inf_type = self._handle_call(node, '__setitem__', node.value.inf_type.getValue(),
-                                                      node.slice.inf_type.getValue(), Any)
+            node.inf_type = TypeInfo(NoType) # type assigned when parent Assign node is visited
         elif node.ctx == astroid.Del:
             node.inf_type = self._handle_call(node, '__delitem__', node.value.inf_type.getValue(),
-                                                      node.slice.inf_type.getValue())
+                                              node.slice.inf_type.getValue())
 
     ##############################################################################
     # Loops
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -383,6 +383,7 @@ class TypeInferer:
         elif node.ctx == astroid.Del:
             node.inf_type = self._handle_call(node, '__delitem__', node.value.inf_type.getValue(),
                                                       node.slice.inf_type.getValue())
+
     ##############################################################################
     # Loops
     ##############################################################################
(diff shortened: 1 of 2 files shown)


simeonkr:
The test for the Del context is currently failing for some reason.

david-yz-liu:
@simeonkr sorry I should have caught this in the previous one. Set the type of a `Subscript` node in a `Store` context to `NoType` (or wrapped in a TypeInfo, I guess). Instead, add another case to `_assign_type`, the helper for `visit_assign`. I think @justinpark715 just worked on that, so you'll need to do a pull/rebase to get the updated code.

Unfortunately I'm not sure why the Delete test is failing. I can look at it more later if you aren't able to make progress.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -248,6 +248,14 @@ def _assign_type(self, target: NodeNG, expr_type: type) -> None:
...
                         target_tvar = self.lookup_type(subtarget, subtarget.name)
                         self.type_constraints.unify(target_tvar, contained_type)
+        elif isinstance(target, astroid.Subscript):
+            # TODO: previous case must recursively handle this one
+            set_call_type = self._handle_call(target, '__setitem__', target.value.inf_type.getValue(),
+                              target.slice.inf_type.getValue(), expr_type)
+            if isinstance(set_call_type, TypeFail):
+                target.inf_type = set_call_type
+            else:
+                target.inf_type = TypeInfo(expr_type)
```
david-yz-liu: This behaviour isn't consistent with the other branches, in which `target.inf_type` isn't changed (from `NoType`). It's okay if you want to record the failure since we've made those changes above in `visit_assign`, but don't change the "success" case here.

tests/test_type_inference/test_subscript.py
```diff
@@ -26,19 +26,53 @@ def test_subscript_homogeneous_list_slice(node):
...
+@given(cs.subscript_node(cs.list_node(min_size=1), cs.index_node()), cs.const_node())
+@settings(suppress_health_check=[HealthCheck.too_slow])
+def test_subscript_store_ctx(node, val):
+    """Test visitor of Subscript node within an assignment."""
+    asgn_node = astroid.Assign()
+    asgn_node.postinit([node], val)
+    module, _ = cs._parse_text(asgn_node)
+    subscript_node = module.body[0].targets[0]
+    val_node = module.body[0].value
+    assert subscript_node.inf_type.getValue() == val_node.inf_type.getValue()
```
david-yz-liu: Right, so this is harder to test. Rather than looking at `subscript_node.inf_type`, I want you to write some tests that will type-check or fail based on whether the `__setitem__` call type-checks or not. E.g.,

```python
# Okay, test passes
[1, 2, 3][0] = 4

# Error
[1, 2, 3][0] = 'david is cool'
```

tests/test_type_inference/test_subscript.py
```diff
@@ -26,19 +26,53 @@ def test_subscript_homogeneous_list_slice(node):
...
+    """Note that this test only takes in a dictionary because the subscript index
+    must be the same type as the dictionary's keys in order to type check.
+    """
+    for key, _ in node.items:
+        new_node = astroid.Subscript()
+        new_node.postinit(node, key)
+        module, _ = cs._parse_text(new_node)
+        for subscript_node in module.nodes_of_class(astroid.Subscript):
+            dict_node = subscript_node.value
+            assert subscript_node.inf_type.getValue() == list(dict_node.items)[0][1].inf_type. getValue()
```
david-yz-liu: Similar comment as above. Also (if you reuse some of this line) there's an extraneous space after the final `.`

python_ta/transforms/type_inference_visitor.py
simeonkr: But if we have an assignment like `a[3] = 0` (say `a` is not homogeneous), where would we store the fact that the expression `a[3]` has type `int` if not in the subscript node (which being the target of the assignment would still have `NoType` if not explicitly unified with `expr_type`)?
Perhaps I should have a lookup function like in the other cases of `_assign_type` and then unify with the corresponding type variable?

david-yz-liu: @simeonkr I've merged that in, so please go ahead. You'll need to fix a merge conflict, which hopefully will be pretty minor.

PR #353, opened 2018-05-15
Typecheck: Call overload function

justinpark715:
Adding "optional_params" attribute to created Callables, contains a list of lists of types, where each list of types is a list of argument types that the Callable can take
Adding test cases for functions with optional arguments

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -195,7 +195,11 @@ class TypeInferer:
     ##############################################################################
     def visit_name(self, node: astroid.Name) -> None:
         try:
-            node.inf_type = TypeInfo(self.lookup_type(node, node.name))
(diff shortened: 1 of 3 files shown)


david-yz-liu:
Indeed, adding attributes dynamically is a risky approach, and not ideal. Here is my suggested approach:

1. Represent alternatives derived from optional parameters with the built-in `Union` type. E.g., the type of `def foo(x: int, y: int = 0) -> str` should be `Union[Callable[[int], str], Callable[[int, int], str]]`.
2. Modify `handle_call` (or related) to handle either a single callable or a union of callables.
3. Modify `type_store.py` to handle this representation as well. Note that they use a decorator to signify different signatures. This will take more work, but if all goes well, you can modify the type of the `functions` and `methods` dictionary attributes to have just single types, rather than lists of types, which is very nice.

PR #354, opened 2018-05-16
typecheck: Fix unify returning unresolved types on higher-order calls

simeonkr:
Also added two tests for polymorphic function calls.

### DIFF
--- a/tests/test_type_constraints/test_unify.py
+++ b/tests/test_type_constraints/test_unify.py
@@ -1,6 +1,6 @@
 from typing import *
 from typing import TupleMeta, CallableMeta, _ForwardRef
-from python_ta.typecheck.base import TypeResult, TypeInfo, TypeFail, TypeConstraints
+from python_ta.typecheck.base import TypeResult, TypeInfo, TypeFail, TypeConstraints, create_Callable
 from nose import SkipTest
 from nose.tools import eq_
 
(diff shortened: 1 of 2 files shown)


tests/test_type_constraints/test_unify.py
```diff
@@ -9,7 +9,7 @@
 
 
 # Helper functions
-def unify_helper(arg1: type, arg2: type, exp_result: type):
+def unify_helper(arg1: type, arg2: type, exp_result: Union[type, TypeFail]):
```
david-yz-liu: This change shouldn't be here. I've already seen it in a separate pull request.

tests/test_type_constraints/test_unify.py
```diff
@@ -190,3 +190,29 @@ def test_diff_callable():
...
     unify_helper(c1, c2, TypeFail(
         f'Incompatible Types {c1} and {c2}'))
+
+
+# Polymorphic types
+def test_simple_polymorphic_call():
+    tc.reset()
+    tv1 = tc.fresh_tvar()
+    tv2 = tc.fresh_tvar()
+    fn1 = Callable[[tv1, tv2], bool]
```
david-yz-liu: This test is okay, but use the helper `create_Callable`. There's something about setting `polymorphic_tvars` which will be important later.

PR #356, opened 2018-05-16
typecheck: Amend unify tests to reflect current expected behavior

simeonkr:
Also check for actual contents of error messages in corresponding tests. This results in a failing test case. The reason is line 299 in base.py: the order of the types is flipped. We could remove this line and duplicate lines 291-297; this won't be as elegant but will match expected behaviour.

### DIFF
--- a/tests/test_type_constraints/test_unify.py
+++ b/tests/test_type_constraints/test_unify.py
@@ -9,14 +9,14 @@ tc = TypeConstraints()
 
 
 # Helper functions
-def unify_helper(arg1: type, arg2: type, exp_result: type):
+def unify_helper(arg1: type, arg2: type, exp_result: Union[type, TypeFail]):
     unify_result = TypeInfo(arg1) >> (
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@simeonkr I like your initiative in looking into this; however, the error messages are too much in flux for me to care about updating the tests to match them. I would merge in the type signature change, and if you really want the updates to the strings in the tests, but *not* in `unify_helper`.

simeonkr:
OK, but do you think the failing test case should be fixed? I.e., if unify(t1, t2) fails because t1 resolves to int and t2 resolves to str, then we want the error message to output 'Incompatible types <class int> <class str>', and not the other way around.

david-yz-liu:
I'd rather leave it for now and later when we get to detailed error messages (with "reasons") we should revisit how we can test for error contents in a way that *doesn't* depend on the order of inputs to `unify`.

PR #357, opened 2018-05-16
Typecheck: handling classes without initalizer

justinpark715:
`visit_call` checks for an `__init__` function, and if there is none, creates a callable with the same type signature that an initialization function would have, to be passed to `unify_call`

### DIFF
--- a/tests/test_type_inference/test_initializer.py
+++ b/tests/test_type_inference/test_initializer.py
@@ -1,10 +1,7 @@
 import astroid
 from typing import _ForwardRef
-from python_ta.transforms.type_inference_visitor import TypeInferer
 import tests.custom_hypothesis_support as cs
-from python_ta.typecheck.base import TypeInfo, TypeFail
 from nose.tools import eq_
-from nose import SkipTest
 
 
 def test_class_with_init():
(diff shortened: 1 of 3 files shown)


tests/test_type_inference/test_initializer.py
```diff
@@ -0,0 +1,62 @@
...
+from typing import _ForwardRef
+from python_ta.transforms.type_inference_visitor import TypeInferer
+from python_ta.typecheck.base import TypeInfo, TypeFail
+from nose.tools import eq_
+
+
+def test_class_with_init():
+    program = """
+    class Foo:
+        a
```
david-yz-liu: This is a less typical case, and should be avoided. These tests should use `pass` when possible.

tests/test_type_inference/test_initializer.py
```diff
@@ -0,0 +1,62 @@
...
+        
+        def __init__(self):
+            a = 5
+    
+    foo = Foo()
+    """
+    ti = TypeInferer()
```
david-yz-liu: These tests should all use the standard test helper functions. See other test files for examples.

tests/test_type_inference/test_initializer.py
```diff
@@ -0,0 +1,62 @@
...
+    ast_mod = astroid.parse(program)
+    ti.environment_transformer().visit(ast_mod)
+    ti.type_inference_transformer().visit(ast_mod)
+    for call_node in ast_mod.nodes_of_class(astroid.Call):
+        assert isinstance(call_node.inf_type.getValue(), _ForwardRef)
+        eq_(call_node.inf_type.getValue(), _ForwardRef('Foo'))
+
+
+def test_class_with_multi_init():
```
david-yz-liu: This isn't a good test. You can't define a method with the same name twice.

justinpark715: Thanks for pointing this out, in the process of trying to fix it, I discovered another bug where `__init__` functions with additional arguments aren't properly checking the number of arguments when they are called!

tests/test_type_inference/test_initializer.py
justinpark715: sorry just to clarify, what makes the case less typical? I intended for this to be a base case of a class definition that had an explicitly declared `__init__` function, as opposed to the later tests that didn't explicitly declare them. Or did you mean the declaration of the `a` variable?

david-yz-liu: Yes, it is indeed the `a` which is not typical. Think about making this test as minimal as possible (hence the `pass` comment).

justinpark715:
As of now, there is not yet support for checking the number of arguments when instantiating a class with an initializer that takes multiple arguments

ie., the following code does not return a "Wrong number of arguments" TypeFail as it should 
```python
class Foo:
    def __init__(self, x):
        self.attr = x

foo = Foo()
```

However, this pull request does solve the issue of classes that are declared without an explicit initializer, and I think it is still worth including just for that small fix, and then we can solve the other problems as a separate issue?

david-yz-liu:
@justinpark715 yeah that's fine, but you should add a TODO somewhere to record this.

PR #358, opened 2018-05-17

simeonkr:


python_ta/transforms/type_inference_visitor.py
```diff
@@ -334,8 +334,19 @@ def visit_index(self, node: astroid.Index) -> None:
...
+        lower_type = node.lower.inf_type.getValue() if node.lower else type(None)
+        upper_type = node.upper.inf_type.getValue() if node.upper else type(None)
+        step_type = node.step.inf_type.getValue() if node.step else type(None)
+        node.inf_type = self._handle_call(node, '__init__', slice, lower_type,
+                                          upper_type, step_type)
+        if node.inf_type.getValue() == type(None):
+            node.inf_type = TypeInfo(slice)
+
+    def visit_ext_slice(self, node: astroid.ExtSlice):
+        for dim in node.dims:
```
david-yz-liu: This check isn't necessary; it's actually checking the structure of the AST, which is beyond the scope of our checker (which is all about types).

The `ExtSlice` node can be set to `NoType`; the more interesting algorithm part is in the `Subscript` node type-checking, where the "slice" attribute is an `ExtSlice`. That should be handled in a separate pull request, though.

tests/test_type_inference/test_subscript.py
```diff
@@ -41,6 +42,31 @@ def test_subscript_homogeneous_list_slice(node):
...
+    slice = astroid.Slice()
+    slice.postinit(astroid.Const(0), astroid.Const('a'))
+    sub_node.postinit(node, slice)
+    module, _ = cs._parse_text(sub_node)
+    for subscript_node in module.nodes_of_class(astroid.Subscript):
+        assert isinstance(subscript_node.inf_type, TypeFail)
+
+
+@given(cs.simple_homogeneous_list_node(min_size=1))
+def test_inference_invalid_ext_slice(node):
```
david-yz-liu: Remove this test. If you want to test the `ExtSlice` node here, give a valid one (e.g., `x[1, 2]`) and check that it has `type_inf` `NoType`.

simeonkr:
I think the failing test is due to _handle_call not properly looking up functions with optional arguments (if I change the type signature of slice.\_\_init\_\_ in builtins.py to take `int`s as arguments instead of `Optional[int]`, then the lookup works on calls where none of the arguments are `None` types). This probably boils down to unify not handling Optional types yet.

david-yz-liu:
@simeonkr you should skip the failing test in this case.

PR #367, opened 2018-05-22
typecheck: Make sure visit_compare works for 'in'

simeonkr:


tests/test_type_inference/test_compare.py
```diff
@@ -23,16 +22,10 @@ def test_compare_equality(left_value, operator_value_tuples):
     assert compare_node.inf_type.getValue() == bool
 
 
-@nottest
-@given(hs.lists(cs.comparator_operator, min_size=3), cs.homogeneous_list(min_size=4))
+@given(hs.lists(cs.comparator_operator, min_size=3), cs.numeric_list(min_size=4))
```
simeonkr: Test previously failed due to attempt to apply '<' and '>' to booleans.

python_ta/typecheck/typeshed/builtins.pyi
```diff
@@ -297,7 +297,7 @@ class str(Sequence[str]):
     def __ge__(self, x: str) -> bool: ...
 
     def __len__(self) -> int: ...
-    def __contains__(self, s: object) -> bool: ...
+    def __contains__(self, s: str) -> bool: ...
```
david-yz-liu: Hmm this is interesting. In principle this file should be an exact copy of `https://github.com/python/typeshed/blob/master/stdlib/3/builtins.pyi`, so I'm hesitant to change it. On the other hand, it does seem like any non-string type is an error. Can you please create a `README.md` under typeshed to make a note of the changes to this file?

python_ta/typecheck/typeshed/builtins.pyi
```diff
@@ -391,7 +391,7 @@ class bytes(ByteString):
     def __rmul__(self, n: int) -> bytes: ...
     if sys.version_info >= (3, 5):
         def __mod__(self, value: Any) -> bytes: ...
-    def __contains__(self, o: object) -> bool: ...
+    def __contains__(self, o: bytes) -> bool: ...
```
david-yz-liu: On the other hand, this (and likely the others down below) is too strict:

```python
>>> 3 in b'123'
False
```
I don't mind making the type-checker more strict here, but you should articulate this in the README as well.

python_ta/typecheck/typeshed/builtins.pyi
```diff
@@ -549,7 +549,7 @@ class slice:
 class tuple(Sequence[_T_co], Generic[_T_co]):
     def __init__(self, iterable: Iterable[_T_co] = ...) -> None: ...
     def __len__(self) -> int: ...
-    def __contains__(self, x: object) -> bool: ...
+    def __contains__(self, x: _T_co) -> bool: ...
```
david-yz-liu: Please don't change this one. I want to treat tuples differently, later.

tests/test_type_inference/test_compare.py
david-yz-liu: This is a nice observation, thanks. Although I guess `bool` inherits from `int`, so eventually we'll allow this. :)

PR #369, opened 2018-05-23
Typecheck: __init__ with multiple arguments

justinpark715:
Added proper error checking in `visit_call`
Edited `visit_attribute` to account for TypeFail objects, which do not have a `__name__` attribute
Added new test to `test_initializer`
Added `reset` function to `type_inference_visitor` to reset TypeStore, edited `_parse_text` to include a flag for resetting before parsing text
Edited tests in `test_classdef` to make use of `reset` (information from previous tests was being carried over)

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -33,6 +33,9 @@ class TypeInferer:
     def __init__(self):
         self.type_constraints.reset()
 
+    def reset(self):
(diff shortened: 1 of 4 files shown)


justinpark715:
Admittedly this is a fairly large pull request, should I break this up into separate parts? My only concern was that adding the error checking for multi-arg init functions was causing some tests to fail without the ability to reset typestore

PR #378, opened 2018-05-28
Typecheck: TNode lookup in visit functions

justinpark715:
Modifying various visit functions to assign TypeVar to inf_type instead of resolving to concrete type right away
`visit_name` and `visit_assign` now look up and assign `TypeVar`
`visit_attribute` and `visit_call` now check to see if the `inf_type` of the `func` or `expr` nodes are `TypeVar`, and resolve if they are
`_merge_sets` now sets parent as the `TypeVar` `_TNode`, rather than the concrete type `_TNode`
Editing test case in `test_assign` to account for changes

* Unsure if this is the best approach in the long run. Strongly considering changing TypeConstraints to use proper disjoint sets, so that looking up connections between TNodes can happen in more than one direction

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -199,7 +199,7 @@ class TypeInferer:
     ##############################################################################
     def visit_name(self, node: astroid.Name) -> None:
         try:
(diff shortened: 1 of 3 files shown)


justinpark715:
Does not yet work when variables are declared or assigned within function definition

PR #384, opened 2018-06-06
typecheck: Handle ExtSlice in Subscript nodes

simeonkr:
I remember that we talked about there not being a need to visit `ExtSlice` nodes. I could move the code in `visit_extslice` to be handled as a case in `visit_subscript` if it's a problem that `ExtSlice` nodes are assigned types.

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -5,7 +5,8 @@ from typing import *
 import typing
 from typing import CallableMeta, TupleMeta, Union, _ForwardRef
 from astroid.transforms import TransformVisitor
-from ..typecheck.base import Environment, TypeConstraints, parse_annotations, create_Callable,_node_to_type, TypeResult, TypeInfo, TypeFail
+from ..typecheck.base import Environment, TypeConstraints, parse_annotations, \
+    create_Callable,_node_to_type, TypeResult, TypeInfo, TypeFail, failable_collect
 from ..typecheck.errors import BINOP_TO_METHOD, UNARY_TO_METHOD, binop_error_message, unaryop_error_message
 from ..typecheck.type_store import TypeStore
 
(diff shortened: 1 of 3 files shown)


simeonkr:
I'm looking into the failing test. It's very odd because it passes for me locally.

david-yz-liu:
I reran the tests and they pass now. However, this implementation is too specific. A few comments:

1. `ExtSlice` is a general node that should support *arbitrary*, *multiple* expressions in the square brackets in a subscript node, e.g. `x[1, 2, 'hi', [int, str]]`. (E.g., think about/look up the implementation of the `List` of `Tuple` types from `typing`.
2. Looking at the implementation of these, you should typecheck `ExtSlice` by doing a `handle_call` against the `__getitem__` method. These methods take just a single argument other than `self`, and this argument is a *tuple* of values.
3. This means you can actually set the type of the `ExtSlice` node to be a `Tuple` (I didn't realize this before). Makes the `handle_call` unification easier.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -388,12 +388,17 @@ def visit_slice(self, node: astroid.Slice) -> None:
         if node.inf_type.getValue() == type(None):
             node.inf_type = TypeInfo(slice)
 
+    def visit_extslice(self, node: astroid.ExtSlice):
+        node.inf_type = TypeInfo(Tuple)
```
david-yz-liu: This is better, but I actually want you to get the individual elements as well (so the example below would have an `ExtSlice` of type `Tuple[int, str]`.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -388,12 +388,23 @@ def visit_slice(self, node: astroid.Slice) -> None:
         if node.inf_type.getValue() == type(None):
             node.inf_type = TypeInfo(slice)
 
+    def visit_extslice(self, node: astroid.ExtSlice):
+        dim_types = []
+        for dim in node.dims:
```
david-yz-liu: Seems like this is a perfect place for `collect_failables`

PR #388, opened 2018-06-08
typecheck: Reimplement can_unify to be as general as possible

simeonkr:


david-yz-liu:
This change is good, but please add to the docstring of `unify` to document this new parameter. Also I guess you should rebase to resolve the merge conflict.

PR #389, opened 2018-06-08
typecheck: Add extra tests for tnode structure

simeonkr:


simeonkr:
The tests are failing because of a graphviz import. The tests are not using graphviz, but for debugging convenience there is an option to draw the tnodes if a test is run manually. Is there a way to keep this behavior?

PR #390, opened 2018-06-08
typecheck: Support for functions with same argument and return type

justinpark715:


david-yz-liu:
It isn't appropriate to make the change in `_set_function_def_environment`, whose job is to initialize a type variable for the return value. If the type variables aren't being unified correctly, this is a problem in the visitor, not here.

Also, one of the extra pieces of logic that this code kind of added is when there are no `Return` nodes in the function body. This is good, but really in this case the return type of the function should be `None`. You can check this in `visit_function_def` directly, and also you should add a test case for it.

PR #402, opened 2018-06-14
typecheck: Modifying failable_collect to accept any iterable as argument

justinpark715:


python_ta/util/monad.py
```diff
@@ -29,14 +29,17 @@ def __str__(self):
         
     def bind(self, fn):
         return fn(self.value)
+
         
-        
-def failable_map(fn : Callable[[TypeVar('T')], Failable], lst : List[Failable]): 
-    # TODO: allow arbitrary containers like tuples
+def failable_map(fn : Callable[[TypeVar('T')], Failable], itr: Iterable[Failable]):
+    if isinstance(itr, list):
```
david-yz-liu: Don't worry about making this check, just call `list(itr)` regardless of what its type is.

python_ta/util/monad.py
```diff
@@ -29,14 +29,17 @@ def __str__(self):
         
     def bind(self, fn):
         return fn(self.value)
+
         
-        
-def failable_map(fn : Callable[[TypeVar('T')], Failable], lst : List[Failable]): 
-    # TODO: allow arbitrary containers like tuples
+def failable_map(fn : Callable[[TypeVar('T')], Failable], itr: Iterable[Failable]):
```
david-yz-liu: Let's tidy up the type signature here: remove the space before the first colon, and add a return type. (Similar to `failable_collect` below.

PR #404, opened 2018-06-15
typecheck: Reimplement can_unify to fix incorrect behavior

simeonkr:
`can_unify` has been reimplemented so that it makes a copy of the current TypeConstraints object and calls unify on the copy so as to avoid modifying the original TypeConstraints. This is inefficient, but my previous implementation was short-sighted and caused problems throughout. The added test case (which previously failed at the assert line) demonstrates the necessity of this change.

To make things more efficient, I could use a hybrid approach: the previous method always ~~fails~~ passes when it should, so a copy would be necessary only to verify a success (most calls to can_unify result in failures anyway). Alternatively, I could hard-code some common cases that allow us to bypass the general check.

### DIFF
--- a/python_ta/typecheck/base.py
+++ b/python_ta/typecheck/base.py
@@ -256,6 +256,24 @@ class TypeConstraints:
     def __init__(self):
         self.reset()
 
+    def __deepcopy__(self, memodict={}):
+        tc = TypeConstraints()
+        tc._count = self._count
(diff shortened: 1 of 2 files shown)


david-yz-liu:
Thanks, @simeonkr. This is okay, at least for now. Please resolve the merge conflicts and I'll pull this in.

PR #405, opened 2018-06-15
typecheck: Modify TypeStore to allow proper method type differentiation in visit_attribute

simeonkr:
The dictionaries in TypeStore now have tuples as their values: the first item is the type signature of the function, as before, and the second item is the the method type ('method', 'classmethod', 'staticmethod') in case of a method or 'attribute' otherwise. This allows `visit_attribute` to not trim off the first argument in static methods.

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -64,7 +64,7 @@ class TypeInferer:
         node.type_environment = Environment()
         for name in node.instance_attrs:
             node.type_environment.locals[name] = self.type_constraints.fresh_tvar(node.instance_attrs[name][0])
-            self.type_store.classes[node.name][name] = [node.type_environment.locals[name]]
+            self.type_store.classes[node.name][name] = [(node.type_environment.locals[name], 'attribute')]
         for name in node.locals:
             node.type_environment.locals[name] = self.type_constraints.fresh_tvar(node.locals[name][0])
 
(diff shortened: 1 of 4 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -536,15 +536,23 @@ def visit_classdef(self, node: astroid.ClassDef) -> None:
...
+                self.type_store.classes[node.name][attr].append((attr_type, 'attribute'))
 
     ##############################################################################
     # Statements
     ##############################################################################
     def visit_attribute(self, node: astroid.Attribute) -> None:
         expr_type = node.expr.inf_type.getValue()
+
+        # case when expr_type is the type of a class
+        if hasattr(expr_type, '__name__') and expr_type.__name__ == 'Type':
```
david-yz-liu: Seems like this happens too early. If we have an instance method `foo` in class`A`, `A.foo()` shouldn't autobind `self`, but you won't have a way of distinguishing this case if you make this update.

P.S. Would be good to add tests to check this.

python_ta/typecheck/base.py
```diff
@@ -597,9 +597,9 @@ def parse_annotations(node, class_tvars=None):
                 arg_types.append(_node_to_type(annotation))
 
         rtype = _node_to_type(node.returns)
-        return create_Callable(arg_types, rtype, class_tvars)
+        return (create_Callable(arg_types, rtype, class_tvars), node.type)
```
david-yz-liu: Don't put parentheses around a return tuple expression

PR #407, opened 2018-06-15
typecheck: Updating inferred type of initializer function

justinpark715:
Updating inferred type of initializer functions to be ForwardRef. Updating relevant test cases
Issue discovered while doing preliminary work in updating visit_call

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -473,6 +473,8 @@ class TypeInferer:
 
         if any(node.nodes_of_class(astroid.Return)):
             inferred_return = self.type_constraints.resolve(node.type_environment.lookup_in_env('return')).getValue()
+        elif node.name == '__init__':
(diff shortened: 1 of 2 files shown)


justinpark715:
Unsure if there was a specific reason you wanted the return type of initializers to be None @david-yz-liu ?

david-yz-liu:
@justinpark715 the return type *is* `None`. Object creation in Python involves more than just calling `__init__`, although this is the simplification we normally use. You can read more about it in the Python Language Reference.

justinpark715:
Right, in actual Python the return type is None, but within the TypeConstraints, is it okay to set the inferred type signature of the init function as Callable[[Foo], Foo] (or whatever class)?

david-yz-liu:
I don't want to change the `__init__` signature to something that's not true in Python. There's room to emulate the correct behaviour with `__new__`, but I don't want you to spend time really looking into that right now.

PR #408, opened 2018-06-18
typecheck: Updating more visit functions in type_inference_visitor

justinpark715:
Updating more visit functions in type_inference_visitor to use unify with new decorator. Updating readme


david-yz-liu:
Thanks, @justinpark715!

PR #409, opened 2018-06-18
typecheck: Updating visit_call, visit_attribute

justinpark715:


justinpark715:
Modified some tests that were indirectly raising a SkipTest via an invalid call to `callable_t.__args__[0].__name__`. Now directly raises SkipTest with explanation
Modified unify_call to add ForwardReference to Class when called as an instance or class method

tests/test_type_inference/test_classdef.py
```diff
@@ -110,6 +111,7 @@ def test_builtin_method_call_bad_argument():
     """
     program = f'x = 1\n' \
               f'x.extend(1)\n'
+    raise SkipTest("TypeStore does not contain extend method")
```
david-yz-liu: On a second look, it appears that this test isn't checking what it wants to. Please modify the test code to `x = [1, 2, 3]`. There should still be a `TypeFail`, but it should be a different one that the previous test. Maybe this test will actually pass.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -267,40 +267,27 @@ def lookup_typevar(self, node: NodeNG, name: str):
...
+    @accept_failable
+    def get_call_signature(self, c) -> TypeResult:
+        if isinstance(c, Type) and isinstance(c.__args__[0], _ForwardRef):
+            c = c.__args__[0]
+        if isinstance(c, _ForwardRef):
+            class_type = c.__forward_arg__
+            if '__init__' in self.type_store.classes[class_type]:
+                init_func = self.type_store.classes[class_type]['__init__'][0]
+            else:
+                # Classes declared without initializer
```
david-yz-liu: This is very good. You should also add a test for this case.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -267,40 +267,27 @@ def lookup_typevar(self, node: NodeNG, name: str):
     ##############################################################################
     # Operation nodes
     ##############################################################################
+    @accept_failable
+    def get_call_signature(self, c) -> TypeResult:
+        if isinstance(c, Type) and isinstance(c.__args__[0], _ForwardRef):
+            c = c.__args__[0]
```
david-yz-liu: In general I'm not a fan of this variable rebinding. Don't use sequenced `if`s here, instead just extract what you want.

Also, since you're changing this anyway, you can add in support for built-in type constructors here (+ tests) fairly easily. I.e., something like `int('3')` should type-check to `int`.

python_ta/typecheck/base.py
```diff
@@ -473,12 +473,16 @@ def _unify_generic(self, tnode1: _TNode, tnode2: _TNode,
...
         return isinstance(self.unify(t1, t2, None, False), TypeInfo)
 
+    @accept_failable
     def unify_call(self, func_type, *arg_types, node=None) -> TypeResult:
         """Unify a function call with the given function type and argument types.
 
         Return a result type.
         """
         # Check that the number of parameters matches the number of arguments.
+        if hasattr(func_type, '__args__') and isinstance(func_type.__args__[0], _ForwardRef) and isinstance(func_type.__args__[-1], _ForwardRef):
```
david-yz-liu: Don't do this here. Instead, put logic inside `get_call_signature` to strip off the `self` argument for `__init__` specifically.

tests/test_type_inference/test_classdef.py
justinpark715: Decided to deal with these in a separate PR #410

justinpark715:
Fixing case-sensitive class name lookup in visit_attribute
Updating relevant test cases in test_classdef

python_ta/transforms/type_inference_visitor.py
justinpark715: This should be tested for in `test_initializer.py`

david-yz-liu:
@justinpark715 very nice work here.

PR #411, opened 2018-06-18
typecheck: Fix lookup_method not using can_unify correctly

simeonkr:


python_ta/typecheck/type_store.py
```diff
@@ -64,39 +65,27 @@ def _parse_functions(self, module: astroid.Module) -> None:
     def lookup_function(self, operator, *args):
         """Helper method to lookup a function type given the operator and types of arguments."""
```
simeonkr: I can fix the code duplication here, but this function isn't being used anywhere.

python_ta/typecheck/type_store.py
```diff
@@ -64,39 +65,27 @@ def _parse_functions(self, module: astroid.Module) -> None:
...
                 if len(args) != len(func_type.__args__) - 1:
                     continue
-                unified = True
-                for t1, t2 in zip(func_type.__args__[:-1], args):
-                    # TODO: replace call to can_unify
-                    if not self.type_constraints.can_unify(t1, t2):
-                        unified = False
-                        break
-                if unified:
+                if(self.type_constraints.can_unify(Callable[list(func_type.__args__[:-1]), Any],
```
david-yz-liu: @simeonkr it's fine to leave the duplication, but you shouldn't need the outer parentheses here. Plus, leave a space after `if`.

PR #415, opened 2018-06-21
Bad loop indexing checker + example file

dhslim:
I made a new PyTA checker that looks for bad loop indexing in for-loops and I added an examples file for the checker


python_ta/checkers/bad_loop_indexing_checker.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # pass in message symbol as a parameter of check_messages
+    @check_messages("bad-loop-indexing")
+    def visit_for(self, node):
+        if self._check_bad_loop_indexing(node):
+            self.add_message('bad-loop-indexing', node=node)
+
+
+# Helper functions
+def _check_iterable_is_range(node: astroid.For) -> List[str]:
```
david-yz-liu: @RyanDJLee suggested returning `Optional[str]`

python_ta/checkers/bad_loop_indexing_checker.py
```diff
@@ -0,0 +1,108 @@
...
+        if iterable_lst:
+            iterable = iterable_lst[0]
+        else:
+            return False
+
+        # Use node_of_class to find all Name nodes.
+        if any(_check_correct_usage(name_node, index, iterable) for name_node in node.nodes_of_class(astroid.Name)):
+            return False
+        # There is at least one 'index' Name node and they are all used badly.
+        elif any(name_node.name == index for name_node in node.nodes_of_class(astroid.Name)):
```
david-yz-liu: Merge this branch with the following one. Note that these two branches have the form

```
if cond:
  return True
else:
  return False
```

python_ta/checkers/bad_loop_indexing_checker.py
```diff
@@ -0,0 +1,108 @@
...
+        return [_iter.args[0].args[0].name]
+    else:
+        return []
+
+
+def _check_correct_usage(node: astroid.Name, index: str, iterable: str) -> bool:
+    """Return whether or not this Name node 'index' was used correctly in the for loop.
+    Return True if correctly used, False if badly used or node is not 'index'.
+    """
+    # If Name node is 'index'
```
david-yz-liu: Avoid the deep nesting here by using an [early return](http://blog.timoxley.com/post/47041269194/avoid-else-return-early)

python_ta/checkers/bad_loop_indexing_checker.py
```diff
@@ -0,0 +1,108 @@
...
+    if node.name == index:
+        # If Name node is inside Subscript node iterable[index]
+        if isinstance(node.parent, astroid.Index) and isinstance(node.parent.parent, astroid.Subscript) \
+                and node.parent.parent.value.name == iterable:
+            subscript_node = node.parent.parent
+            # iterable[index] (eg. lst[i]) used in the left side of an assignment
+            # Assign node of form "a = b = 1"
+            if isinstance(subscript_node.parent, astroid.Assign):
```
david-yz-liu: The rest of this part should be simplified if you access the `subscript_node.ctx` attribute. You'll have to do some reading about that, but it should greatly help.

python_ta/checkers/bad_loop_indexing_checker.py
```diff
@@ -0,0 +1,109 @@
...
+
+import astroid
+from pylint.interfaces import IAstroidChecker
+from pylint.checkers import BaseChecker
+from pylint.checkers.utils import check_messages
+from typing import List, Optional
+from astroid.node_classes import NodeNG
+
+
+class BadLoopIndexingChecker(BaseChecker):
```
david-yz-liu: The terms "good" and "bad" are vague and should generally be avoided when naming things. Here, you should rename this checker (and the corresponding files/methods/etc.) in the following ways:

1. Replace references to "bad loop indexing" with "unnecessary indexing".
2. Replace references to "bad usage" with "redundant index use". (The term "redundant" here means that the index name could have been replaced with a simpler reference to the list element. This should be added to the docstring of a function that checks whether the use is 'redundant' or not.)

python_ta/checkers/bad_loop_indexing_checker.py
```diff
@@ -0,0 +1,109 @@
...
+    __implements__ = IAstroidChecker
+    # name is the same as file name but without _checker part
+    name = 'bad_loop_indexing'
+    # use dashes for connecting words in message symbol
+    msgs = {'E9994': (f'Bad use of loop variable %s in for loop',
```
david-yz-liu: Put the quotation marks around the variable here (i.e., `...loop variable "%s" in for loop`) so that you can remove them below.

python_ta/checkers/bad_loop_indexing_checker.py
```diff
@@ -0,0 +1,109 @@
...
+        # i) For all index_node in augassign_index_nodes, .lookup != node.target, so they are all rebound
+        # or ii) there are no AugAssign nodes.
+        return not any(index_node.lookup(index)[1][0] == node.target
+                       for index_node in _good_use_index_nodes(node, index)) \
+            and (_index_assign_name_nodes(node, index) or _index_name_nodes(node, index))
+
+
+# Helper functions
+def _iterable_if_range(node: Optional[NodeNG]) -> Optional[str]:
+    """Return the string of the iterable if this node is of the form: range(len(iterable)).
```
david-yz-liu: The phrase "string of the iterable" isn't clear. Refer to the iterable's name.

python_ta/checkers/bad_loop_indexing_checker.py
```diff
@@ -0,0 +1,109 @@
...
+    Return None otherwise.
+    """
+    # Order of operations is important.
+    if isinstance(node, astroid.Call):
```
david-yz-liu: There's no reason to nest the `if`s here. Just leave one with a long string of conditions. Also, use parentheses to enclose the different clauses, rather than using backslashes:

```
if (condition1 and
        condition2 and
        condition3 and
        ...
        conditionN):
```

Also, you're missing a final check for whether the inner `args[0]` is a `Name` node.

python_ta/checkers/bad_loop_indexing_checker.py
```diff
@@ -0,0 +1,109 @@
...
+        index = node.target.name
+
+        # Check if the iterable of the for loop is of the form "range(len(lst))".
+        iterable = _iterable_if_range(node.iter)
+        if iterable is None:
+            return False
+
+        # AugAssign is special (i += 2 is actually i = i + 2).
+        # Check if any AssignName nodes in an AugAssign node refers to the iteration variable.
+        if any(index_node.lookup(index)[1][0] == node.target for index_node in _augassign_index_nodes(node, index)):
```
david-yz-liu: The code is certainly improved from last time. However, the central algorithm can still be simplified, basically by merging the checks for `AugAssign` and `Name` nodes. Restructure your code as follows:

1. Obtain a list of *all* `AssignName` and `Name` nodes that have the same name as the loop target, and whose `.lookup` method call returns that target. Note that `nodes_of_class` can take in multiple node classes.
2. Check whether *all* of the nodes from the previous step are "redundant". This can be done using just a single helper function that takes a `AssignName` or `Name` node and determines whether it's "redundant"; also, this flips the boolean output from what you had before, so be careful here.
3. Return the correct thing. Remember to check for whether there's at least one usage of the loop target at all.

Note that this strategy will reduce the overall number of helper functions.

python_ta/checkers/bad_loop_indexing_checker.py
dhslim: would this be acceptable? 
Return the iterable's name if this node is of the form: range(len(iterable)).

david-yz-liu: Sure, that's better. You've been assuming that the argument to `len` is an identifier like `lst`, but something like `range(len([1, 2, 3]))` is technically possible (and this checker should *ignore* lists with such an iterable).

PR #417, opened 2018-06-23
typecheck: Improve logic in unify

simeonkr:


python_ta/typecheck/base.py
```diff
@@ -450,15 +451,7 @@ def unify(self, t1: type, t2: type,
         elif conc_tnode2 is not None:
             return self.unify(t2, t1, ast_node)
```
simeonkr: The two cases below are covered by the previous cases (in particular the `elif ct1 == ct2` and the `else` cases).

PR #419, opened 2018-06-23
typecheck: Add better set comparison in test_tnode_structure

simeonkr:
Otherwise, we can't test sets containing _ForwardRefs, etc.

### DIFF
--- a/tests/test_type_constraints/test_tnode_structure.py
+++ b/tests/test_type_constraints/test_tnode_structure.py
@@ -31,8 +31,8 @@ def tc_to_disjoint(tc: TypeConstraints) -> List[Set[Union[type, str]]]:
 
 
 def compare_list_sets(s1: List[Set[Union[type, str]]], s2: List[Set[Union[type, str]]]):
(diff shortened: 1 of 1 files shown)


simeonkr:
We've noticed that checks for pull requests are failing randomly, and it's always due to various health check errors. I think this may be due to the slower implementation of can_unify. Perhaps we should look into the Hypothesis settings?

tests/test_type_constraints/test_tnode_structure.py
```diff
@@ -31,8 +31,8 @@ def tc_to_disjoint(tc: TypeConstraints) -> List[Set[Union[type, str]]]:
 
 
 def compare_list_sets(s1: List[Set[Union[type, str]]], s2: List[Set[Union[type, str]]]):
-    list_sets_1 = s1.copy()
-    list_sets_2 = s2.copy()
+    list_sets_1 = [set([str(t) for t in s]) for s in s1]
```
david-yz-liu: Use a set comprehension here

PR #421, opened 2018-06-26
typecheck: Initiate support for inheritance

simeonkr:
I am currently looking into the astroid source code for mro() which seems to be discarding subscripted classes like `SupportsAbs[int]`. Ideally, we shouldn't have to store both `__bases` and `__mro` in the classes dictionary.

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -525,7 +525,7 @@ class TypeInferer:
         self.type_constraints.unify(self.lookup_inf_type(node.parent, node.name),
                                     Type[_ForwardRef(node.name)], node)
 
-        self.type_store.classes[node.name]['__bases'] = [self.type_store.node_to_type(base)
+        self.type_store.classes[node.name]['__bases'] = [_node_to_type(base)
                                                          for base in node.bases]
         self.type_store.classes[node.name]['__mro'] = [cls.name for cls in node.mro()]
 
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -27,12 +27,15 @@ class TypeInferer:
     """
     type_constraints = TypeConstraints()
     type_store = TypeStore(type_constraints)
+    type_constraints.type_store = type_store
 
     def __init__(self):
         self.type_constraints.reset()
 
     def reset(self):
+        self.type_constraints.reset()
         self.type_store = TypeStore(self.type_constraints)
+        self.type_constraints.type_store = self.type_store
 
     ###########################################################################
     # Setting up the environment
@@ -277,7 +280,10 @@ class TypeInferer:
     def get_call_signature(self, c, node: NodeNG) -> TypeResult:
         if hasattr(c, '__name__') and c.__name__ == 'Type':
             class_type = c.__args__[0]
-            class_name = c.__args__[0].__forward_arg__
+            if isinstance(class_type, _ForwardRef):
+                class_name = c.__args__[0].__forward_arg__
+            else:
+                class_name = class_type.__name__
         elif isinstance(c, _ForwardRef):
             class_type = c
             class_name = c.__forward_arg__
@@ -519,6 +525,10 @@ class TypeInferer:
         self.type_constraints.unify(self.lookup_inf_type(node.parent, node.name),
                                     Type[_ForwardRef(node.name)], node)
 
+        self.type_store.classes[node.name]['__bases'] = [self.type_store.node_to_type(base)
+                                                         for base in node.bases]
+        self.type_store.classes[node.name]['__mro'] = [cls.name for cls in node.mro()]
+
         # Update type_store for this class.
         # TODO: include node.instance_attrs as well?
         for attr in node.locals:
@@ -557,7 +567,11 @@ class TypeInferer:
         class_name, class_type, inst_expr = expr_inf_type >> self.get_attribute_class
 
         if class_name in self.type_store.classes:
-            attribute_type = self.type_store.classes[class_name].get(node.attrname)
+            attribute_type = None
+            for par_class_type in self.type_store.classes[class_name]['__mro']:
+                attribute_type = self.type_store.classes[par_class_type].get(node.attrname)
+                if attribute_type:
+                    break
             if attribute_type is None:
                 class_tnode = self.type_constraints.get_tnode(class_type)
                 node.inf_type = TypeFailLookup(class_tnode, node, node.parent)
--- a/python_ta/typecheck/base.py
+++ b/python_ta/typecheck/base.py
@@ -319,6 +319,7 @@ class TypeConstraints:
     type_to_tnode: Dict[str, _TNode]
 
     def __init__(self):
+        self.type_store = None
         self.reset()
 
     def __deepcopy__(self, memodict={}):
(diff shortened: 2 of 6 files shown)


python_ta/typecheck/base.py
```diff
@@ -281,6 +281,8 @@ def __deepcopy__(self, memodict={}):
         tc._count = self._count
         tc._nodes = []
         tc.type_to_tnode = {}
+        if hasattr(self, 'type_store'):
```
david-yz-liu: Don't use `hasattr`. See what I said in #423.

python_ta/typecheck/type_store.py
```diff
@@ -88,6 +95,39 @@ def lookup_method(self, operator, *args):
...
+                inner_var = tuple([self.node_to_type(elt) for elt in node.slice.value.elts])
+            else:
+                inner_var = self.node_to_type(node.slice.value)
+            try:
+                outer_var = globals()[node.value.name]
+            except KeyError:
+                outer_var = _ForwardRef(node.value.name)
+            return outer_var[inner_var]
+
+    def is_descendant(self, child: type, ancestor: type):
```
david-yz-liu: Why aren't you using `mro` here?

python_ta/typecheck/type_store.py
```diff
@@ -88,6 +95,39 @@ def lookup_method(self, operator, *args):
                     return func_type
             raise KeyError
 
+    def node_to_type(self, node):
```
david-yz-liu: Add type signatures here and for all other functions you're touching.

python_ta/typecheck/type_store.py
```diff
@@ -88,6 +95,39 @@ def lookup_method(self, operator, *args):
...
+        if isinstance(node, astroid.Name):
+            try:
```
david-yz-liu: Don't do it like this. Use a simple `if ... elif ... else ...` structure.

david-yz-liu: This also helps avoid the use of the identifier `inner_var`.

python_ta/typecheck/type_store.py
```diff
@@ -88,6 +95,39 @@ def lookup_method(self, operator, *args):
...
+                else:
+                    inner_var = _ForwardRef(node.name)
+            return inner_var
+        elif isinstance(node, astroid.Subscript):
+            if hasattr(node.slice.value, 'name') and \
+                    isinstance(node.slice.value, astroid.Tuple):
+                inner_var = tuple([self.node_to_type(elt) for elt in node.slice.value.elts])
+            else:
+                inner_var = self.node_to_type(node.slice.value)
+            try:
```
david-yz-liu: Recurse on `node.value`

python_ta/typecheck/type_store.py
```diff
@@ -88,6 +95,39 @@ def lookup_method(self, operator, *args):
...
+                inner_var = globals()[node.name]
+            except KeyError:
+                if node.name in _TYPESHED_TVARS:
+                    inner_var = _TYPESHED_TVARS[node.name]
+                else:
+                    inner_var = _ForwardRef(node.name)
+            return inner_var
+        elif isinstance(node, astroid.Subscript):
+            if hasattr(node.slice.value, 'name') and \
```
david-yz-liu: This is funny. Can `Tuple` nodes really have a `name` attribute?

python_ta/typecheck/type_store.py
```diff
@@ -88,6 +95,39 @@ def lookup_method(self, operator, *args):
...
+                    isinstance(node.slice.value, astroid.Tuple):
+                inner_var = tuple([self.node_to_type(elt) for elt in node.slice.value.elts])
```
david-yz-liu: Remove the square brackets.

python_ta/typecheck/type_store.py
```diff
@@ -88,6 +95,39 @@ def lookup_method(self, operator, *args):
                     return func_type
             raise KeyError
 
+    def node_to_type(self, node):
+        if isinstance(node, astroid.Name):
+            try:
+                inner_var = globals()[node.name]
```
david-yz-liu: I'm not sure about using `globals` here -- why not `self.classes`?

python_ta/transforms/type_inference_visitor.py
```diff
@@ -523,7 +523,7 @@ def visit_classdef(self, node: astroid.ClassDef) -> None:
         self.type_constraints.unify(self.lookup_inf_type(node.parent, node.name),
                                     Type[_ForwardRef(node.name)], node)
 
-        self.type_store.classes[node.name]['__bases'] = [self.type_store.node_to_type(base)
+        self.type_store.classes[node.name]['__bases'] = [_node_to_type(base)
```
simeonkr: It looks like this function has already been implemented in base.py. Can't believe I missed it.

python_ta/typecheck/type_store.py
simeonkr: For some reason mro does not return base classes like Sequence[~T] or SupportsAbs[int], that's the reason why I'm storing `__bases` separately and using this function (for regular non-mro inheritance). There still remains the issue of how to deal with method resolution that depends on classes like these.

python_ta/typecheck/base.py
```diff
@@ -149,10 +149,50 @@ class TupleSubscript(TypeVar, _root=True):
...
+
+
+def create_Callable(args: Iterable[type], rtype, class_poly_vars=None):
+    poly_vars = class_poly_vars or []
+    poly_vars = set(poly_vars)
     c = Callable[list(args), rtype]
-    c.polymorphic_tvars = poly_vars
+    poly_vars.update(_get_poly_vars(c))
+    c.polymorphic_tvars = set()
+    c.polymorphic_tvars.update(poly_vars)
```
simeonkr: This is an odd way to set `c.polymorphic_tvars` to `poly_vars`, but it is a workaround for a general problem with Callable attributes that we've been bumping into.
For example,
```
c = Callable[[int],int]
c.x = 3
c = Callable[[int],str]
print(c.x)  # prints 3
c.x = 2
print(c.x)  # prints 3
```

python_ta/typecheck/base.py
```diff
@@ -149,10 +149,50 @@ class TupleSubscript(TypeVar, _root=True):
...
+
+_BUILTIN_TO_TYPING = {
+    'list': 'List',
+    'dict': 'Dict',
+    'tuple': 'Tuple',
+    'set': 'Set',
+    'frozenset': 'FrozenSet',
+    'function': 'Callable'
+}
+
```
simeonkr: I've moved these dictionaries here because they're used in `get_poly_vars`, which in turn is called in the `create_Callable` calls in `TYPE_SIGNATURES`

python_ta/typecheck/type_store.py
```diff
@@ -83,11 +88,24 @@ def lookup_method(self, operator, *args):
...
                 if len(args) != len(func_type.__args__) - 1:
                     continue
-                if self.type_constraints.can_unify(Callable[list(func_type.__args__[:-1]), Any],
-                                                   Callable[list(args), Any]):
+                if self.type_constraints.can_unify(Callable[list(args), Any],
+                                                   Callable[list(func_type.__args__[:-1]), Any]):
                     return func_type
             raise KeyError
 
+    def is_descendant(self, child: type, ancestor: type):
```
david-yz-liu: Put in a return type here

python_ta/typecheck/type_store.py
```diff
@@ -1,10 +1,14 @@
 import astroid
 from astroid.builder import AstroidBuilder
 from collections import defaultdict
-from python_ta.typecheck.base import parse_annotations, class_callable, accept_failable
+from python_ta.typecheck.base import parse_annotations, \
+    class_callable, accept_failable, _node_to_type
 from typing import Callable
 import os
+from builtins import *
```
david-yz-liu: Are these two imports still necessary? I think you removed the `globals()` call that had previously required it, if I understand correctly.

PR #422, opened 2018-06-27
typecheck: Adding some TypeFail subclasses

justinpark715:
Adding TypeFail subclasses `TypeFailUnify` and `TypeFailFunction`, with `get_reasons()` function, to be used by checkers and reporters to generate comprehensive error messages
Adding `find_path` function to `_TNode`, to be used by `TypeFailUnify` to find a chain of type inferences
Modifying `unify_generic` to return `TypeFail` objects with `_TNode`s that have not been fully resolved to concrete types, so that error checking has more information available
Also modifying `unify_generic` to create `TypeFailUnify` based on generic, rather than propagating `TypeFail` between individual elements
Modifying `_find_type_fail` in test_tnode_structure.py to be used by other test files
Adding relevant test cases 

*Still a work in progress, planning on adding other subclasses such as TypeFailLookup, which have been implemented, but are not included in this pull request as they have not been fully tested

### DIFF
--- a/python_ta/typecheck/base.py
+++ b/python_ta/typecheck/base.py
@@ -21,11 +21,58 @@ class _TNode:
         self.adj_list = []
         self.ast_node = ast_node
 
+    def __hash__(self):
+        return hash(repr(self))
+
     def __eq__(self, other: '_TNode'):
-        if str(self.type) == str(other.type):
-            return True
+        return str(self.type) == str(other.type)
+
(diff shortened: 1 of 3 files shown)


david-yz-liu:
@justinpark715 this PR is a bit too complicated for me to look at. Please simplify it by making only the changes required to support `TypeFailUnify` and nothing else.

PR #426, opened 2018-06-28
typecheck: Replacing TypeInfo(NoType)

justinpark715:
Replacing TypeInfo(NoType) with NoType, now a subclass of TypeResult, moved to base.py


david-yz-liu:
This is a bit subtle, but you should use `NoType()`, not `NoType`. Otherwise `.inf_type` won't always be an instance of `TypeResult`

PR #427, opened 2018-06-29
typecheck: Small fix to logic in get_attribute_class

justinpark715:
Removing accidental case attempting to call `lower()` on a `None` object

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -547,7 +547,7 @@ class TypeInferer:
         else:
             class_name = None
 
-        if class_name not in self.type_store.classes:
+        if class_name is not None and class_name not in self.type_store.classes:
             class_name = class_name.lower()
 
         return class_name, is_inst_expr


python_ta/transforms/type_inference_visitor.py
```diff
@@ -545,14 +545,13 @@ def get_attribute_class(self, t: type) -> Tuple[str, bool]:
...
         if isinstance(class_type, _ForwardRef):
             class_name = class_type.__forward_arg__
+            if class_name not in self.type_store.classes:
+                class_name = class_name.lower()
         elif hasattr(t, '__name__'):
             class_name = t.__name__
         else:
             class_name = None
 
-        if class_name not in self.type_store.classes:
```
david-yz-liu: Hmm this change is interesting. The above `if` has three cases; I'm worried that you missed out on the middle case with this change. You could just change this condition here to `class_name is not None and ..`

justinpark715: Sure thing! That was actually my original solution as well, I just felt like the if statement felt a bit long

PR #432, opened 2018-07-03
typecheck: Add support for binary expressions with reflected operands

simeonkr:


tests/test_type_inference/test_binops.py
```diff
@@ -1,4 +1,5 @@
 import astroid
+from typing import *
```
david-yz-liu: Let's not import `*` here

david-yz-liu:
@simeonkr this is really good start (so I can merge in), but it's more complicated than just this. The good reference is https://docs.python.org/3/reference/datamodel.html#object.__radd__.

PR #435, opened 2018-07-04
typecheck: Adding type annotation checker

justinpark715:
Adding checker for type annotations. Currently support type signatures for functions and methods, class variables, and instance variables

### DIFF
--- a/python_ta/checkers/type_annotation_checker.py
+++ b/python_ta/checkers/type_annotation_checker.py
@@ -1,0 +1,53 @@
+"""checker for type annotation.
+"""
+
+import astroid
+from pylint.interfaces import IAstroidChecker
+from pylint.checkers import BaseChecker
+
+
+class TypeAnnotationChecker(BaseChecker):
+
+    __implements__ = IAstroidChecker
+
+    name = 'TypeAnnotationChecker'
+    msgs = {'E9970': ('Function parameter is missing type annotation',
+                      'type-annotation-param',
+                      'Presented when a type annotation is missing.'),
+            'E9971': ('Function is missing return type annotation',
+                      'type-annotation-return',
+                      'Presented when a type annotation is missing.'),
+            'E9972': ('Class variable is missing type annotation',
+                      'type-annotation-class-var',
+                      'Presented when a type annotation is missing.'),
+            'E9973': ('Instance variable is missing type annotation',
+                      'type-annotation-inst-var',
+                      'Presented when a type annotation is missing.'),
+           }
+
+    # this is important so that your checker is executed before others
+    priority = -1
+
+    def visit_functiondef(self, node):
+        for i in range(len(node.args.annotations)):
+            if node.args.annotations[i] is None:
+                # Check if function is a non-static instance method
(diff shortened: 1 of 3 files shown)


python_ta/checkers/type_annotation_checker.py
```diff
@@ -0,0 +1,57 @@
...
+                      'Presented when a type annotation is missing.'),
+           }
+
+    # this is important so that your checker is executed before others
+    priority = -1
+
+    def visit_functiondef(self, node):
+        if isinstance(node.parent, astroid.ClassDef):
+            # Function is a method. Type annotation not required for first argument 'self'
+            if any(arg is None for arg in node.args.annotations[1:]):
```
david-yz-liu: This is good, but since we have the expertise at this point, let's also very here that this function isn't a *staticmethod* (since these should have an annotation for the first parameter).

python_ta/checkers/type_annotation_checker.py
```diff
@@ -0,0 +1,57 @@
...
+                self.add_message('type-annotation-arg', node=node)
+            # __init__ has return type None
+            if node.returns is None and node.name != '__init__':
```
david-yz-liu: This is a bit scary. Are you saying that astroid `returns` attribute can't distinguish between the following cases:

```
def f():

def f() -> None
```

If so, there's a larger problem here.

python_ta/checkers/type_annotation_checker.py
```diff
@@ -0,0 +1,57 @@
...
+                self.add_message('type-annotation-return', node=node)
+        else:
+            if any(arg is None for arg in node.args.annotations):
+                self.add_message('type-annotation-arg', node=node)
```
david-yz-liu: For parameters, I'd like to highlight the individual parameter, not the entire function.

python_ta/checkers/type_annotation_checker.py
```diff
@@ -0,0 +1,57 @@
...
+            if node.returns is None:
+                self.add_message('type-annotation-return', node=node)
```
david-yz-liu: Similarly, here highlight just the function arguments. This is a bit ugly, but better than highlighting the whole function.

python_ta/checkers/type_annotation_checker.py
```diff
@@ -0,0 +1,57 @@
...
+from pylint.interfaces import IAstroidChecker
+from pylint.checkers import BaseChecker
+
+
+class TypeAnnotationChecker(BaseChecker):
+
+    __implements__ = IAstroidChecker
+
+    name = 'TypeAnnotationChecker'
+    msgs = {'E9970': ('Function argument is missing type annotation',
```
david-yz-liu: Use "parameter" rather than "argument" (and in the short identifier for this error as well).

python_ta/checkers/type_annotation_checker.py
justinpark715: In the second case, the `node.returns` attribute would contain a `Const` node with the value None, so it should be able to distinguish between the two cases

david-yz-liu: Okay, then this is not a special case. Students are expected to annotate the return type of `__init__`.

PR #437, opened 2018-07-05
typecheck: Add support for automatic type conversion in arithmetic expressions

simeonkr:
This approach may be a little too simplistic, because ideally we would have to account for each of the "common type" conversion cases in [https://docs.python.org/3.6/reference/expressions.html?highlight=common%20type%20arithmetic](url). However, I think that it is always the case that unification against a float (or complex) happens in the context of one of these cases (which means that any subsequent unifications that involve the to-be converted SupportsFloat/SupportsComplex types do as well). Would there be any situation where would strictly expect, say, a float type and unification against an int (in that direction) would fail?

### DIFF
--- a/python_ta/typecheck/base.py
+++ b/python_ta/typecheck/base.py
@@ -536,11 +536,15 @@ class TypeConstraints:
             elif ct1 == Any or ct2 == Any:
                 return TypeInfo(ct1)
             # Handle inheritance
-            elif self.type_store and \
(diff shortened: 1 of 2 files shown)


david-yz-liu:
@simeonkr so I guess in principle the problem is when users define their own classes implementing these methods. While this isn't a huge concern in CSC108/148, I don't think it should take that much more effort to implement correctly over the suggestion here.

PR #441, opened 2018-07-06
typecheck: Updating and implementing type_inference_checker

justinpark715:
@david-yz-liu  I'm unsure if the `hasattr` in `visit_default` is the best choice... However, the only alternative I can think of is changing the behaviour of `patch_type_inference_transform` in `python_ta/patches/type.py` so that it removes `type-inference-checker` from the list of custom checkers if `type_transformer.visit(ast)` raises an error. I'm worried that might involve too much changing of the attributes of the linter itself, so I left it with the `hasattr` for now

### DIFF
--- a/examples/type_inference_example.py
+++ b/examples/type_inference_example.py
@@ -1,0 +1,6 @@
+"""Module docstring."""
+
+
+A = 1
+
+A = "Hello"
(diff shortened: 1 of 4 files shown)


david-yz-liu:
@justinpark715 it's fine to experiment with this but I won't be enabling this checker until I'm confident it's ready. Also, don't add error messages without checking with me.

PR #443, opened 2018-07-06
typecheck: Adding support for isnot comparator

justinpark715:


david-yz-liu:
Thanks, @justinpark715 -- please add a test case for this as well.

PR #444, opened 2018-07-06
typecheck: Fixing usage of wrap_container in visit_extslice

justinpark715:


python_ta/transforms/type_inference_visitor.py
```diff
@@ -379,7 +379,8 @@ def visit_extslice(self, node: astroid.ExtSlice):
         if isinstance(unif_res, TypeFail):
```
david-yz-liu: What I means that this check is no longer necessary if you're using `>>` below.

PR #446, opened 2018-07-09
typecheck: Adding missing type annotations, docstrings

justinpark715:


david-yz-liu:
@justinpark715 a good start. Two general guidelines for writing docstrings:

1. Use the imperative tense to describe a function's behaviour (i.e., "Return x" instead of "Returns x")
2. Use proper grammar, and in particular good punctuation.

PR #448, opened 2018-07-09
typecheck: Adding TypeFailFunction support to _handle_call

justinpark715:
Also removing obsolete TypeErrorInfo and related code


david-yz-liu:
It's good to remove TypeErrorInfo, but the other changes go too far and remove code that you should not touch. In particular, don't remove the special cases here.

justinpark715:
Just to clarify, do you mean the `error_func=binop_error_message` cases?

david-yz-liu:
Yes that's right

PR #449, opened 2018-07-10
typecheck: Adding case in _node_to_type for astroid.Const node

justinpark715:
Fixing crash in nodes/Arguments.py, when an argument is annotated with a string literal, as opposed to the str type

### DIFF
--- a/python_ta/typecheck/base.py
+++ b/python_ta/typecheck/base.py
@@ -855,6 +855,8 @@ def _node_to_type(node, locals=None):
         return [_node_to_type(t) for t in node.elts if not isinstance(t, astroid.Ellipsis)]
     elif isinstance(node, astroid.Const) and node.value is None:
         return None
+    elif isinstance(node, astroid.Const) and isinstance(node.value, str):
+        return _node_to_type(node.value)
     else:
         return node
 


david-yz-liu:
This change is indeed good, but your PR here should include a description of what prompted you to make this change.

david-yz-liu:
Oh, I see. This is why giving context is so valuable. There's already a case on line 812 that handles string literals. So you can recurse on that if you detect a `Const` node whose value is a `str`.

justinpark715:
So in this code:
```
def f(a: 'annotation', b=1, c=2, *d, e, f=3, **g):
    pass
```
The node passed to `_node_to_type` is a `Const` node with a value of `'annotation'`. If we recurse on the value node, that would result in the type of `a` being set to `_ForwardRef('annotation')`, is that the behaviour we should expect?

david-yz-liu:
Yes, that's right.

python_ta/typecheck/base.py
```diff
@@ -831,6 +831,8 @@ def _node_to_type(node, locals=None):
         return [_node_to_type(t) for t in node.elts if not isinstance(t, astroid.Ellipsis)]
     elif isinstance(node, astroid.Const) and node.value is None:
         return None
+    elif isinstance(node, astroid.Const):
```
david-yz-liu: You should only do this if the Const node stores a string. Other `Const` values should proceed to the `else`

PR #450, opened 2018-07-10
typecheck: Fixing crash in visit_assign 

justinpark715:
Preventing crashing of visit_assign when assigning an unbound indentifier


david-yz-liu:
Actually (thinking about the other "unbound identifier" PR), I think it's better to modify `lookup_typevar` to be monadic (i.e., return a `TypeInfo` or `TypeFail` subclass). That involves more changes, but hopefully reduces other problems as well.

PR #453, opened 2018-07-10
typecheck: Adding support for invalid return nodes

justinpark715:
Fixing crashes in nodes/Return.py


david-yz-liu:
For this one, there are really two changes to make:

1. If `node.value is None`, then the return type should be unified against `type(None)`  (or just `None`, I can never keep track of how the type of `None` is represented). I.e., `return` and `return None` should have the same type-checking effect.
2. Catch the exception if `return` is not in the type environment. But use the try-except block to *just* isolate that part; don't put it inside a larger expression.

PR #454, opened 2018-07-10
typecheck: Adding support for assignment to Starred variables

justinpark715:
First noticed while fixing crash in nodes/Starred.py
Adding support for assigning tuples, lists and iterators to a tuple with a starred variable

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -7,7 +7,7 @@ from typing import CallableMeta, TupleMeta, Union, _ForwardRef
 from astroid.transforms import TransformVisitor
 from ..typecheck.base import Environment, TypeConstraints, parse_annotations, \
     _node_to_type, TypeResult, TypeInfo, TypeFail, failable_collect, accept_failable, create_Callable_TypeResult, \
-    wrap_container, NoType, TypeFailLookup, TypeFailFunction, TypeFailReturn
+    wrap_container, NoType, TypeFailLookup, TypeFailFunction, TypeFailReturn, TypeFailStarred
 from ..typecheck.errors import BINOP_TO_METHOD, BINOP_TO_REV_METHOD, UNARY_TO_METHOD, \
     binop_error_message, unaryop_error_message
 from ..typecheck.type_store import TypeStore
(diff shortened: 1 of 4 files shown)


david-yz-liu:
For this one, you should use an explicit `isinstance` check, because the type inference needs to change. In `*a, b = range(5)`, the type of `a` is `List[int]`, not `int`. This is worht making an explicit test case for.

david-yz-liu:
This is much improved. However, the right-hand side being a `TupleMeta` has become complex enough that the logic is worth pulling out into a separate helper.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -224,21 +224,54 @@ def _assign_type(self, target: NodeNG, expr_type: type, node: astroid.Assign) ->
...
+                            lambda t: self.type_constraints.unify(target_tvar, t.__args__[0], node))
+
+                    if isinstance(unif_result, TypeFail):
+                        return unif_result
+            return assign_result
         elif isinstance(target, astroid.Subscript):
             # TODO: previous case must recursively handle this one
             return self._handle_call(target, '__setitem__', target.value.inf_type, target.slice.inf_type, expr_type)
 
+    def _assign_tuple(self, target: astroid.Tuple, value: TupleMeta, node: astroid.Assign):
```
david-yz-liu: The return type is missing here.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -224,21 +224,54 @@ def _assign_type(self, target: NodeNG, expr_type: type, node: astroid.Assign) ->
...
+        if any(isinstance(subtarget, astroid.Starred) for subtarget in target.elts):
```
david-yz-liu: This method implementation should be simplified. Use the following approach:

1. Find the index of the `Starred` subtarget, if any. If there's more than one, stop. You can make another TypeFail subclass for this.
2. Using the index, partition the tuple into the types for the starred subtarget, and all other types. Similarly, extract a list of all non-starred subtargets.
3. Unify the non-starred subtargets with the "other types" from (2). This is currently lines 269-271. The same logic applies in all cases.
4. If there's a starred subtarget, unify it using the same approach you're currently using.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -224,21 +224,61 @@ def _assign_type(self, target: NodeNG, expr_type: type, node: astroid.Assign) ->
...
+
+        if starred_index is not None:
+            starred_length = len(value.__args__) - len(target.elts) + 1
+            starred_values = node.value.elts[starred_index:starred_index+starred_length]
+            new_starred_value = wrap_container(List, self.unify_elements(starred_values, node))
+            subvalue_types = list(value.__args__)
+            new_subvalues = subvalue_types[:starred_index] + [new_starred_value] + subvalue_types[starred_index + starred_length:]
+
+            for subtarget, subvalue in zip(target.elts, new_subvalues):
+                if isinstance(subtarget, astroid.Starred):
```
david-yz-liu: Ah, so the fact that you still need to differentiate `Starred` from the other nodes makes this approach not as elegant as the one I suggested. Please switch it.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -224,21 +224,61 @@ def _assign_type(self, target: NodeNG, expr_type: type, node: astroid.Assign) ->
...
+                    target_tvar = self.lookup_typevar(subtarget, subtarget.name)
+
+                unif_result = self.type_constraints.unify(target_tvar, subvalue, node)
+
+                if isinstance(unif_result, TypeFail):
+                    return unif_result
+
+            assign_result = TypeInfo(value)
+        else:
+            elt_inf_types = [self.lookup_inf_type(subtarget, subtarget.name) for subtarget in target.elts]
```
david-yz-liu: The other reason to merge this case into the previous one is to ensure the logic is the same. This line uses `lookup_inf_type`, while above uses `lookup_typevar`.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -224,21 +224,61 @@ def _assign_type(self, target: NodeNG, expr_type: type, node: astroid.Assign) ->
...
+                if starred_index is None:
+                    starred_index = i
+                else:
+                    return TypeFailStarred(node)
+
+        if starred_index is not None:
+            starred_length = len(value.__args__) - len(target.elts) + 1
+            starred_values = node.value.elts[starred_index:starred_index+starred_length]
+            new_starred_value = wrap_container(List, self.unify_elements(starred_values, node))
+            subvalue_types = list(value.__args__)
```
david-yz-liu: Is `value.__args__` a tuple? You should be able to slice it directly.

PR #455, opened 2018-07-10
typecheck: Adding support for Unions in binop_error_message

justinpark715:
Fixing crash in examples/pylint/C0121_singleton_comparison.py


david-yz-liu:
@simeonkr had the same idea in #451. Why don't you wait until that one is merged in, and then add your change on top of it.

PR #456, opened 2018-07-10
typecheck: Improve inference for class methods

simeonkr:


tests/test_type_inference/test_attribute.py
```diff
@@ -31,7 +31,7 @@ def foo(cls, x):
         '''
     module, _ = cs._parse_text(program, reset=True)
     for attribute_node in module.nodes_of_class(astroid.Attribute):
-        assert attribute_node.inf_type.getValue() == Callable[[int], int]
+        assert attribute_node.inf_type.getValue() == Callable[[Type[_ForwardRef('A')], int], int]
```
david-yz-liu: This test isn't right, so if it's passing there's a problem with the implementation. Class methods should still autobind when accessed through an instance: `A().foo(0)` should work, while `A().foo(A, 0)` should be an error.

david-yz-liu:
@simeonkr this is good, but there's a merge conflict to fix.

PR #457, opened 2018-07-10
typecheck: Modifying lookup_typevar to be monadic

justinpark715:
Changing lookup_typevar to return a TypeInfo or TypeFail, thereby fixing crashes caused by nodes/Call.py and nodes/Assign.py
Updating test case in test_binops to use failable_collect

This replaces pull request #450 and #452

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -196,22 +196,20 @@ class TypeInferer:
     # Name lookup and assignment
     ##############################################################################
     def visit_name(self, node: astroid.Name) -> None:
-        try:
-            node.inf_type = self.lookup_inf_type(node, node.name)
-        except KeyError:
-            if node.name in self.type_store.classes:
-                node.inf_type = TypeInfo(Type[__builtins__[node.name]])
-            elif node.name in self.type_store.functions:
-                node.inf_type = TypeInfo(self.type_store.functions[node.name][0][0])
-            else:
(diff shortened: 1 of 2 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -196,22 +196,18 @@ def _closest_frame(self, node, name):
     # Name lookup and assignment
     ##############################################################################
     def visit_name(self, node: astroid.Name) -> None:
-        try:
-            node.inf_type = self.lookup_inf_type(node, node.name)
-        except KeyError:
+        node.inf_type = self.lookup_inf_type(node, node.name)
```
david-yz-liu: Store this in a temporary variable. Prefer symmetry in branches for initialization (i.e., add an extra branch below for `node.inf_type = ___` if the value is _not_ a `TypeFail`.

python_ta/transforms/type_inference_visitor.py
```diff
@@ -258,9 +254,12 @@ def _lookup_attribute_type(self, node, instance_name, attribute_name):
         class_env = self._closest_frame(node, class_name).locals[class_name][0].type_environment
         return self.type_constraints.resolve(class_env.lookup_in_env(attribute_name))
 
-    def lookup_typevar(self, node: NodeNG, name: str) -> TypeVar:
+    def lookup_typevar(self, node: NodeNG, name: str) -> TypeInfo:
```
david-yz-liu: The return type should be a `TypeResult`

PR #463, opened 2018-07-13
typecheck: Implementing TypeFailFunction for non-callable objects

justinpark715:
Add check for non-callable objects in `get_call_signature`, and add relevant test case
Change prompted by `examples/pylint/E1102_not_callable.py`

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -274,7 +274,7 @@ class TypeInferer:
     # Operation nodes
     ##############################################################################
     @accept_failable
-    def get_call_signature(self, c, node: NodeNG) -> TypeResult:
+    def get_call_signature(self, c: type, node: NodeNG) -> TypeResult:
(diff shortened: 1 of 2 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -284,8 +284,10 @@ def get_call_signature(self, c, node: NodeNG) -> TypeResult:
         elif isinstance(c, _ForwardRef):
             class_type = c
             class_name = c.__forward_arg__
-        else:
+        elif isinstance(c, CallableMeta) or (hasattr(c, '__origin__') and c.__origin__ is Union):
```
david-yz-liu: Why is `Union` permitted here?

justinpark715: Overloaded functions pass in a Union of different possible function signatures. However, I'm now realizing this is too general, as it would then accept any Union, even if the elements were not function signatures, so I will adjust accordingly

PR #465, opened 2018-07-13
typecheck: Resolving crash caused by duplicate bases

justinpark715:
Prompted by crash in `examples/pylint/E0241_duplicate_bases.py`


david-yz-liu:
In general use None sparingly. Here, you could set the attribute to be a list with just the class itself, I think. But check with Simeon

PR #473, opened 2018-07-17
typecheck: Ignore irrelevant operators when doing arithmetic conversion...

simeonkr:
...and initiate support for auto-conversion within assign statements


python_ta/transforms/type_inference_visitor.py
```diff
@@ -408,12 +413,13 @@ def visit_binop(self, node: astroid.BinOp) -> None:
...
 
     @accept_failable
-    def _arithm_convert(self, t1_: type, t2_: type) -> Optional[TypeInfo]:
-        for t1, t2 in [(t1_, t2_), (t2_, t1_)]:
-            if t1 is complex and self.type_store.is_descendant(t2, SupportsComplex):
-                return TypeInfo(complex)
-            if t1 is float and self.type_store.is_descendant(t2, SupportsFloat):
-                return TypeInfo(float)
+    def _arithm_convert(self, op : str, t1_: type, t2_: type) -> Optional[TypeInfo]:
+        if op in ['**', '*', '/', '//', '%', '+', '-']:
```
david-yz-liu: Eh, this is pretty hacky. You should be able to use typeshed to check this in a principled manner:

1. Do the arithmetic conversion, resulting in the general numeric type, say `num`.
2. Do a `_handle_call` on (say) `__add__` with argument types `[num, num]`.

This should result in just a slightly more complex implementation of `visit_binop` and the like---but if you treat `_arithm_convert` as a preprocessing step only, it shouldn't get much more complicated.

PR #477, opened 2018-07-17
typecheck: Move (most) TypeStore initialization to environment visitors

simeonkr:
I don't know if it's possible to move the remaining code (involving `TypeStore`) in `visit_classdef` to `_set_classdef_environment` since it it involves updating `TypeStore` with the inferred types of attribute nodes, but inference happens after environment setup.


david-yz-liu:
@simeonkr I merged this in, however looking at the next block of code, I think that it should _not_ be necessary to update at least the *attributes* in `TypeStore` in the way currently done. Instead:

1. In `_set_classdef_environment`, make sure every name in `node.locals` gets an entry in the type store.
2. In `visit_classdef`, just do a `unify` on the type variable.

But maybe I don't remember what the data format in the typestore is. But basically I'm doubtful of using the fact that the dictionaries store *lists* here, and would rather use `unify`.

simeonkr:
I think the current behaviour is there for consistency between the way `TypeStore` is used in builtins and user-defined classes. For builtins we store the inferred types (all variants) directly in `TypeStore`. Storing attributes as `TypeVars` in `TypeStore` would mean that we would have to make additional calls to `resolve` during every access. Is the problem you are referring to the fact that attributes are stored as list values within the `type_store.classes` dictionary?

PR #479, opened 2018-07-18
typecheck: Infer empty list/set/dict types as polymorphic types

simeonkr:


david-yz-liu:
This is good, but please update the README. Also I'm interested in seeing some tests added here. The old behaviour was *more* permissive than now, so look at code that should now have a type error, whereas before it didn't. E.g.,

```
x = [] # List[TypeVar(T)]
x.append(1)  # T == int
x.append('hi')  # unification error
```

simeonkr:
Do you mean to say that we should no longer support heterogeneous lists (before the inferred type in that example would be `List[Any]`)? Also, we don't have support for list inference based on calls to `append` so I suppose I should test this in another way?

simeonkr:
Calls to `append` are currently not resulting in correct unification. We would need a way to unify the TypeVar that appears in the list argument (say `_T1` in `List[_T1]`) with the fresh TypeVar generated for the call to `append` (say `_T2` in `Callable[[_T2], None]`), via the generic TypeVar in builtins (`_T` in `def append(self, object: _T)`) which rightly so is not unifed against anything. Would you like me to look into this now or after the other issues have been resolved?

david-yz-liu:
You can do that after the other issues, but it's certainly a high priority.

PR #482, opened 2018-07-19
typecheck: Moving lookup_type and types_in_callable to test helper file

justinpark715:


david-yz-liu:
These are definitely the right changes, but I don't think these belong in a separate file. Let's put them with some of the existing (type-checking) test helpers.

justinpark715:
Should we just put then into the custom hypothesis support file?

david-yz-liu:
Yeah, that's alright for now. In the future I'll take a look at the test organization more globally.

PR #489, opened 2018-07-23
typecheck: Adding support for generic type annotations with subscripts

justinpark715:
Prompted by errors raised when running type inference checker on student code for CSC 148, specifically
```
_cars: Dict[str, 'Car']
```

Does not yet support unspecified multi-parameter Tuple annotation, ie. if a variable is annotations as `Tuple`, then it will only unify with single-parameter tuples

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -7,7 +7,7 @@ from typing import CallableMeta, TupleMeta, Union, _ForwardRef
 from astroid.transforms import TransformVisitor
 from ..typecheck.base import Environment, TypeConstraints, parse_annotations, \
     _node_to_type, TypeResult, TypeInfo, TypeFail, failable_collect, accept_failable, create_Callable_TypeResult, \
-    wrap_container, NoType, TypeFailLookup, TypeFailFunction, TypeFailReturn, TypeFailStarred
+    wrap_container, NoType, TypeFailLookup, TypeFailFunction, TypeFailReturn, TypeFailStarred, _gorg
 from ..typecheck.errors import BINOP_TO_METHOD, BINOP_TO_REV_METHOD, UNARY_TO_METHOD, \
     INPLACE_TO_BINOP, binop_error_message, unaryop_error_message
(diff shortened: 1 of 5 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -195,9 +195,27 @@ def visit_assign(self, node: astroid.Assign) -> None:
...
     def visit_annassign(self, node: astroid.AnnAssign) -> None:
         var_inf_type = self.lookup_typevar(node.target, node.target.name)
-        self.type_constraints.unify(var_inf_type, _node_to_type(node.annotation.name), node)
+        ann_type = self._ann_node_to_type(node.annotation)
+        self.type_constraints.unify(var_inf_type, ann_type, node)
         node.inf_type = NoType()
 
+    def _ann_node_to_type(self, node: astroid.Name) -> type:
+        ann_node_type = _node_to_type(node)
+        if isinstance(ann_node_type, GenericMeta) and ann_node_type.__args__ is None:
```
david-yz-liu: This part should be changed. We tend so use both `Dict` and `dict` as aliases for `Dict[Any, Any]`, so this is how I'd like you to interpret all generic types with missing arguments.

justinpark715: So I did initially have all the generic types set to `Any`, but then I saw #479 and thought that we were now inferring them as polymorphic? Otherwise we could have code such as
```
x: Dict

x = {'Hello', 0}
x = {1, 'Goodbye'}
```

Which would be permitted, even though `x` should be inferred as `Dict[str, int]` after the first assignment?

david-yz-liu:
The difference is that this is an annotation.

justinpark715:
So just to clarify, in the above example, should we allow both assignments to `x`? Or should we make the inferred type of `x` more specific after the first assignment, and then indicate a unification error in the second assignment?

david-yz-liu:
User annotations override all type inference. The example you gave should not result in a type error; rather, the type of `x` should remain as `Dict[Any, Any]` throughout the course of the type-checking.

PR #495, opened 2018-07-24
typecheck: Handle inheritance for generic types properly

simeonkr:
This change allows unification of types like `List[int]` and `Sequence[~_T]`.


david-yz-liu:
@simeonkr I merged this in, but you can submit another PR to unskip that test you added.

PR #496, opened 2018-07-25
typecheck: Fixing type_annotation_checker

justinpark715:
Fixing logic in `for attr_node in node.locals`, was previously raising false positives. Previous type annotation example only checked that missing annotations were pointed out, and variables with correct annotations were also classified as missing annotations. For example, `class_var: int = 0` and `class_var: int` would previously be highlighted as missing a type annotation. 
Also now requires instances variables to be annotated in class body. type_annotation_example edited to provide example


david-yz-liu:
Please give more detail bout the "false positives" that were being raised.

PR #501, opened 2018-07-26
typecheck: Adding TypeFail when type annotation is list of types

justinpark715:
Prompted by errors caused by student example code, where student incorrectly annotates a variable as `[str, 'Car']` rather than `Dict[str, 'Car']`


python_ta/transforms/type_inference_visitor.py
```diff
@@ -217,6 +217,8 @@ def _ann_node_to_type(self, node: astroid.Name) -> type:
                 ann_type = wrap_container(ann_node_type, Any)
             else:
                 ann_type = wrap_container(ann_node_type, Any)
+        elif isinstance(ann_node_type, List):
```
david-yz-liu: We should really generalize this. Not sure exactly where, but there should be a check like `isinstance(annotation, type)` that, if false, immediately leads to a type fail. One tricky case is that a string like `'Car'` is okay (interpreted as forward reference), but everything else should be a type.

PR #505, opened 2018-07-26
typecheck: Adding type_environment to Lambda nodes

justinpark715:


david-yz-liu:
This one is interesting. My main comment is that this is fine, but I'd really like to see a `visit_lambda` implemented. Note that in lambda, the `return` keyword isn't used; the return type of the Callable is the type of the body of the lambda.

PR #507, opened 2018-07-27
typecheck: Modifying visit_return to use get_attr

justinpark715:
Prompted by crashes when return statement is not in function, thus `node.scope()` returns a node that does not have a `returns` attribute


david-yz-liu:
You have actually made two changes in this PR, only one of which is the one you described. Please explain the other one, ideally with a test case.

PR #509, opened 2018-07-27
typecheck: Changing type signature of initializers for classes not yet defined

justinpark715:
This behaviour may be too permissive, as it leads to initializers (for classes that have not yet been defined) accepting any number of arguments, as noted in a skipped test case.
Alternative solution may be to move some of the functionality of `visit_functiondef` into `_set_functiondef_environment`?

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -417,7 +417,7 @@ class TypeInferer:
                 init_func = Callable[init_args[1:-1], init_args[0]]
             else:
                 # Classes declared without initializer
-                init_func = Callable[[], class_type]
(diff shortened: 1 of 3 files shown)


david-yz-liu:
The current approach isn't ideal. The introduction of ellipsis in `Callable` signatures should happen in a separate PR, independent of this issue.

As you are hinting at, this is more about populating `type_store` properly before type-checking occurs. Indeed, in both test cases `A` does have an `__init__`, so it's not good if `__init__` can't be found for the class in type-store current error-checking. If this is indeed the case, that's the problem you should fix. However, this doesn't mean you have to put in a good callable signature then; you can instead just make sure there's a fresh type variable associated with the method.

PR #512, opened 2018-07-27
typecheck: Fixing logic in visit_for and node_to_type for dictionaries

justinpark715:
Prompted by crashes in student code when students try to iterate through dictionary

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -559,15 +559,30 @@ class TypeInferer:
     # Loops
     ##############################################################################
     def visit_for(self, node: Union[astroid.For, astroid.Comprehension]) -> None:
-        iter_type_result = self._handle_call(node, '__iter__', node.iter.inf_type)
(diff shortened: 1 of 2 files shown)


david-yz-liu:
While it's bad if PyTA crashed, the dictionary looping example you gave is not valid Python code. You can't do `for k, v in dict` to bind both keys and values. `dict.__iter__` should return an iterator over just the keys.

Revert the changes you made to type_inference_visitor. You can keep your tests, and I'll look at what the failures are.

justinpark715:
The code that was causing crashes that I meant to fix was  actually `a, b in some_dict.items()`, looping over a list of tuples similar to the example in `test_for_list_tuple_multi_arg`. Do you still want me to revert the changes? Or can I just fix the incorrect dictionary example?

david-yz-liu:
While you should fix the example, I am still unconvinced about these changes, so yes, you should revert them. In the case of the list of tuples, this should already be correctly handled by calling `__iter__` on the list. If this is not the case, there is a bigger problem that cannot be solved in `visit_for`.

PR #522, opened 2018-08-02
typecheck: Creating new callables in visit_attribute to avoid accidental modification of type store

justinpark715:
Previous implementation would inadvertently modifying the type store at line 757 `func_type.__args__ = func_type.__args__[1:]`, causing the type signature in type store to be reduced by one parameter each time the function was called. This was noticed while testing the type inference checker on student code. 
Some test cases are failing, this is due to `create_Callable` mistakenly treating some regular type variables as polymorphic. This issue is resolved in #521 and once both pull requests are merged, all tests pass


david-yz-liu:
@justinpark715 you can rebase this one

PR #524, opened 2018-08-03
typecheck: Adding support for subscript and attribute targets in visit_for

justinpark715:
Prompted by crashes when students attempted similar code
Although the test cases listed below are valid python code, it may be worth raising an error that this is not good practice

### DIFF
--- a/python_ta/transforms/type_inference_visitor.py
+++ b/python_ta/transforms/type_inference_visitor.py
@@ -604,7 +604,13 @@ class TypeInferer:
         iter_type_result = self._handle_call(node, '__iter__', node.iter.inf_type)
         if isinstance(node.target, astroid.AssignName):
             target_inf_type = self.lookup_inf_type(node.target, node.target.name)
-        else:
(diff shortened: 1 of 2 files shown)


python_ta/transforms/type_inference_visitor.py
```diff
@@ -562,7 +562,14 @@ def visit_for(self, node: Union[astroid.For, astroid.Comprehension]) -> None:
         iter_type_result = self._handle_call(node, '__iter__', node.iter.inf_type)
         if isinstance(node.target, astroid.AssignName):
             target_inf_type = self.lookup_inf_type(node.target, node.target.name)
-        else:
+        elif isinstance(node.target, astroid.AssignAttr):
+            target_inf_type = self._lookup_attribute_type(node.target, node.target.expr.inf_type, node.target.attrname)
+        elif isinstance(node.target, astroid.Subscript):
+            # bassemph e6
```
david-yz-liu: Remove this line

PR #525, opened 2018-08-03
typecheck: Resolving type variables

justinpark715:
Resolving type variables in `_lookup_attribute_type`, `visit_attribute` and `visit_subscript` before they are passed to helper functions `_gorg` and `get_attribute_class` which require concrete variables


david-yz-liu:
@justinpark715 I think this change is okay, but it causes a test to fail now.

PR #527, opened 2018-08-04
typecheck: Add missing case in _node_to_type

simeonkr:


david-yz-liu:
@simeonkr I'd like to have more context for this one. I assume this is about module imports, but can you point to something concrete?

simeonkr:
Oh, this is to allow module names in type annotations; e.g., in `base.py`, we have `def find_function_def(self, tn: _TNode) -> Optional[astroid.FunctionDef]` so `node_to_type` will be called on the node corresponding to `Optional[astroid.FunctionDef]` and, in turn, on the one corresponding to `astroid.FunctionDef`, which is an `Attribute` node.

PR #528, opened 2018-08-04
typecheck: Fix unhandled TypeFail case in visit_binop

simeonkr:


python_ta/transforms/type_inference_visitor.py
```diff
@@ -442,6 +442,13 @@ def visit_call(self, node: astroid.Call) -> None:
     def visit_binop(self, node: astroid.BinOp) -> None:
         left_inf, right_inf = node.left.inf_type, node.right.inf_type
 
+        if isinstance(left_inf, TypeFail):
```
david-yz-liu: `_arithm_convert` already has the accept failable; why don't you branch on `isinstance(arithm_type, TypeFail)` below

simeonkr:
You're right. The code was right in the first place because even if `arithm_type` if a `TypeFail` we still enter the `if arithm_type` case and do not proceed. I guess I got confused by the way `@accept_failable` works when I reviewed this code the second time, but it makes sense to me again.

PR #529, opened 2018-08-04
typecheck: Fix literal_substitute returning reference to original Callable

simeonkr:
This ensures that `literal_substitute` returns a copy of the original Callable when no substitutions are made.


tests/test_type_inference/test_classdef.py
```diff
@@ -44,7 +44,7 @@ def test_classdef_method_call():
     attribute_node = list(module.nodes_of_class(astroid.Attribute))[1]
     expected_rtype = attribute_node.parent.inf_type.getValue()
     actual_rtype = inferer.type_constraints.resolve(attribute_node.inf_type.getValue().__args__[-1]).getValue()
-    assert actual_rtype == expected_rtype
```
simeonkr: This is due to the way `TypeVar` comparisons are intended to behave ([https://github.com/python/typing/issues/512](https://github.com/python/typing/issues/512)). Since the new code makes a copy of any previous type variables, direct comparison won't work.

PR #531, opened 2018-08-04
typecheck: Add support for overloaded constructors

simeonkr:


tests/test_type_inference/test_function_overload.py
```diff
@@ -117,3 +119,51 @@ def foo(x: int, y: int=None):
...
+    ast_mod, ti = cs._parse_text(program)
+    for assgn_node in ast_mod.nodes_of_class(astroid.AssignName):
+        if assgn_node.name == 'x':
+            x = ti.lookup_typevar(assgn_node, assgn_node.name)
+            assert ti.type_constraints.resolve(x).getValue() == int
+        if assgn_node.name == 'y':
+            y = ti.lookup_typevar(assgn_node, assgn_node.name)
+            assert ti.type_constraints.resolve(y).getValue() == float
+
+
```
simeonkr: These tests highlight the more general case that we are still trying to figure out. There are several variants for `__add__` but the type of `f` will be inferred to be the first one that successfully unifies (the `int` variant). The obvious approach is to modify `can_unify` to return a `Union` of all unifiable types instead of just the first one. But to make this fully work, we would need to modify the underlying graph logic. For say we have a type variable corresponding to the first argument of a call to a function whose type is a `Union` of `Callables`. If that variable is unified against a `str`, the `Union` type would have to collapse to contain only those `Callables` whose first argument is a `str`.

PR #532, opened 2018-08-05
 typecheck: Implement attribute inference for unresolved class names

simeonkr:
This change allows us to handle attribute access of function arguments (which are unresolved classes or class instances). This is done by making an entry in `TypeStore` for the `TypeVar` corresponding to the argument and storing the attribute types within.

### DIFF
--- a/tests/test_type_inference/test_attribute.py
+++ b/tests/test_type_inference/test_attribute.py
@@ -106,3 +106,22 @@ def test_attribute_self_bind():
     x = [ti.lookup_typevar(node, node.name) for node
          in module.nodes_of_class(astroid.AssignName)][0]
     assert str(ti.type_constraints.resolve(x).getValue()) == "typing.List[int]"
+
+
+def test_unknown_class_attribute():
+    program = \
+        '''
+        def foo(a, x):
+            a = 4
+            x.name1 + 2
+            x.name2 = 3
+            y = x.name1
+        '''
+    module, ti = cs._parse_text(program, reset=True)
+    x = [ti.lookup_typevar(node, node.name) for node
+         in module.nodes_of_class(astroid.AssignName)][0]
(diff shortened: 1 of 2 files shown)


tests/test_type_inference/test_attribute.py
```diff
@@ -106,3 +107,137 @@ def test_attribute_self_bind():
...
+    for binop_node in module.nodes_of_class(astroid.BinOp):
+        assert binop_node.inf_type.getValue() == int
+    for assgn_node in module.nodes_of_class(astroid.AssignName):
+        if assgn_node.name == 'z':
+            z = ti.lookup_typevar(assgn_node, assgn_node.name)
+            assert ti.type_constraints.resolve(z).getValue() == int
+
+
+def test_unknown_class_subscript_attribute():
+    raise SkipTest("Currently, the inferred type of the z variable is str")
```
simeonkr: This is related to the issue described in #531 (because the first of all the `__getitem__` signatures apparently returns a `str`).

tests/test_type_inference/test_attribute.py
```diff
@@ -106,3 +107,137 @@ def test_attribute_self_bind():
...
+        x = 3
+        y = x.name
+        '''
+    module, ti = cs._parse_text(program, reset=True)
+    assgn_node = list(module.nodes_of_class(astroid.Assign))[1]
+    assert isinstance(assgn_node.inf_type, TypeFail)
+
+
+def test_attribute_unification_fail():
+    raise SkipTest('This case is not supported yet')
```
simeonkr: This case requires that when we unify a `TypeVar` with a class instance type or class type, we unify each common attribute for both of the types (although that won't be enough to make this pass due to the fact that the type variable in `unify_call` will be a fresh one). This will be a separate pull request.

PR #533, opened 2018-08-05
typecheck: Add support for imported names

simeonkr:
This is based on #532 (I'll rebase and fix the commit history as soon as it's merged) and is mainly tests (because #532 sort of handles the case automatically). 

Imported names are treated as unresolved `TypeVars`: that way we depend only on the current code file to infer anything about them.


david-yz-liu:
@simeonkr you can rebase this one now.

david-yz-liu:
@simeonkr Okay I've thought about this approach and am not comfortable doing type inference on imported names. I'd rather use `Any` as I originally asked for.

PR #534, opened 2018-08-07
typecheck: Implementing support for unifying compatible generics with different numbers of parameters

justinpark715:
Prompted by crash caused by student code where student called `sorted()` on a Dict object. Type signature for `sorted()` is `Callable[[Iterable[~T]], List[~T]]`, and unifying `Dict` with `Iterable` caused a crash when attempting to create a `Dict` object with just one argument

### DIFF
--- a/python_ta/typecheck/base.py
+++ b/python_ta/typecheck/base.py
@@ -678,9 +678,14 @@ class TypeConstraints:
             else:
                 arg_inf_types.append(result)
 
+        if len(g1.__parameters__) > len(arg_inf_types):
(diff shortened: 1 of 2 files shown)


david-yz-liu:
This change strikes me as far too permissive (allows lots of different gorgs to be unified, and ignores typefails in the loop above the change). Instead, right now I want you to just treat `Iterable[T]` as a special case, in fact generalizing how we do for loops and unpacking assignment.

Modify `unify` as follows: _If one of the arguments is an `Iterator`, check the signature of the other argument for its `__iter__` method._

PR #535, opened 2018-08-07
typecheck: Adding check for type resolving to generic containing same type

justinpark715:
Prompted by recursion errors raised by student code when creating nested lists with append

### DIFF
--- a/python_ta/typecheck/base.py
+++ b/python_ta/typecheck/base.py
@@ -531,7 +531,10 @@ class TypeConstraints:
             try:
                 repr = self.find_repr(self.type_to_tnode[str(t)])
                 if repr and repr.type is not t:
-                    return self.resolve(repr.type)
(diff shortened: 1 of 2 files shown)


python_ta/typecheck/base.py
```diff
@@ -505,7 +505,10 @@ def resolve(self, t: type) -> TypeResult:
             try:
                 repr = self.find_repr(self.type_to_tnode[str(t)])
                 if repr and repr.type is not t:
-                    return self.resolve(repr.type)
+                    if getattr(repr.type, '__args__', None) is not None and any(elt is t for elt in repr.type.__args__):
```
david-yz-liu: This can be simplified by using `getattr` directly in the comprehension in the any: `any(elt is t for elt in getattr(...))`, of course choosing a good default value.

PR #538, opened 2018-08-08
Implementing support for unifying Iterable with subclasses of Iterable

justinpark715:
Prompted by crash caused by student code where student called sorted() on a Dict object. Type signature for sorted() is Callable[[Iterable[~T]], List[~T]], and unifying Dict with Iterable caused a crash when attempting to create a Dict object with just one argument


justinpark715:
Also had to make a change to a for look in `is_descendant`, as it ignored the first argument of `__orig_bases__`. Does not cause any change in functionality for pre-existing test cases as far as I can tell, as the first argument will not have any impact anyway if it isn't a GenericMeta, as checked in the next line.

PR #540, opened 2018-08-09
typecheck: Adding support for checking unions of function type signatures, while checking for iterable parameter

justinpark715:
Prompted by crashes when running check on `list()` initializer, which can take 0 or 1 arguments and thus has a union of different possible function signatures


david-yz-liu:
I ended up resolving this in a different way in #543 (by moving the iterable auto-conversion check to `unify_call`).

PR #545, opened 2018-08-14
typecheck: Updating failing test in test_subscript

justinpark715:
Test case now fails due to `_handle_call` returning a `TypeFail` instance with a message set by `error_func`, instead of `TypeFailFunction`

Also, unsure what this test case is actually testing? `List` is a generic type

### DIFF
--- a/tests/test_type_inference/test_subscript.py
+++ b/tests/test_type_inference/test_subscript.py
@@ -125,7 +125,7 @@ def test_subscript_slice():
         '''
     module, _ = cs._parse_text(program)
     assign_node = next(module.nodes_of_class(astroid.Assign))
-    assert isinstance(assign_node.inf_type, TypeFailFunction)
+    assert isinstance(assign_node.inf_type, TypeFail)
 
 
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@justinpark715 thanks! Not sure how I missed this one earlier.

This test (should be) checking for the case when a value can take a subscript on one kind of value but not a slice. So `List[int]` is okay but `List[:]` isn't. Not sure if there's a better type of value that we could use for this test instead of `List`.

justinpark715:
While I was looking for the source of this particular test, I noticed that `List[int]` actually also results in a TypeFail, because `_handle_call` tries to call `__get_item__` with `int` as an argument. It doesn't result in any direct problems right now, because we usually use `List[int]` in type annotations and `_ann_node_to_type` has its own way of interpreting subscripts, and it ignores the `inf_type` set by `visit_subscript`. But it does mean that `List[int]` will show up in the final error report as a TypeFail.

david-yz-liu:
Thanks for pointing that out. I think the current behaviour is okay for our purposes, but it's good to note for the future.

PR #546, opened 2018-08-14
typecheck: Resolving type in error functions and checking for subscripts of type None

justinpark715:
Prompted by crashes when calling `subscript_error_message` while the subscript value `inf_type` is TypeVar or type(None)


david-yz-liu:
Thanks @justinpark715!

PR #562, opened 2018-12-25
Fix for function spanning multiple lines of code

dhslim:


dhslim:
Previously, when checking functions where the parameter is on a different line than the rest of the function definition (as in the example below), it doesn't highlight the entire function defintion because msg.node is the Arguments node. 
screen shot 2019-01-08 at 2 27 43 pm
With this change, the checker will highlight the entire function definition since msg.node.parent is the FunctionDef node. 
screen shot 2019-01-08 at 2 33 19 pm
Note: there is still a small issue where the last line

PR #565, opened 2019-01-23
Adjust what gets highlighted

dhslim:
This is what it looks like now:
screen shot 2019-01-22 at 9 35 29 pm


dhslim:
Actually, I think this is better:
screen shot 2019-01-22 at 9 56 22 pm

dhslim:
Actually, I think this is better:
screen shot 2019-01-22 at 9 56 22 pm

PR #566, opened 2019-01-23
Transform Arguments node to contain parentheses

dhslim:
This change with commit 2f4fc24 (Add custom renderer for R0913_too_many_arguments), you will see this:
screen shot 2019-01-22 at 9 44 58 pm
This change with commit 3381e6a (Adjust what gets highlighted), you will see this:
screen shot 2019-01-22 at 9 48 21 pm


david-yz-liu:
@dhslim this is good, but can you also look at `tests/test_setendings.py` for the commented-out `test_arguments` method, and try to uncomment them and make sure they pass?

PR #571, opened 2019-03-12
Fix for empty Arguments node case

dhslim:
Addressed these two things
- if the parent is a `FunctionDef`, the Arguments node spans from the open parenthesis to the closing parenthesis, but not including either.
- if the parent is a `Lambda`, just set it to be the location of the colon after the `lambda` keyword


dhslim:
note: change "left bracket" to "right bracket"

PR #579, opened 2019-05-31
CFG - While Node Visitor

kavins14:
Added `while` visitor method as well as the corresponding tests.

### DIFF
--- a/python_ta/cfg/graph.py
+++ b/python_ta/cfg/graph.py
@@ -31,9 +31,12 @@ class ControlFlowGraph:
         An "empty" node for this purpose is when source has no statements.
         """
         if source.statements == []:
-            for edge in source.predecessors:
-                edge.target = target
-                target.predecessors.append(edge)
+            if source.predecessors == []:   # implies source == empty start node
+                self.start = target
+            else:
+                for edge in source.predecessors:
+                    edge.target = target
+                    target.predecessors.append(edge)
         else:
             CFGEdge(source, target)
 
@@ -51,6 +54,21 @@ class ControlFlowGraph:
 
         for edge in block.successors:
             yield from self._get_blocks(edge.target, visited)
+
+    def get_edges(self) -> Generator[CFGEdge, None, None]:
+        """Generate a sequence of all edges in this graph."""
+        yield from self._get_edges(self.start, set())
+
+    def _get_edges(self, block: CFGBlock,
+                   visited: Set[int]) -> Generator[CFGEdge, None, None]:
+        if block.id in visited:
+            return
+
+        visited.add(block.id)
+
+        for edge in block.successors:
+            yield edge
+            yield from self._get_edges(edge.target, visited)
 
 
 class CFGBlock:
--- a/python_ta/cfg/graph.py
+++ b/python_ta/cfg/graph.py
@@ -31,7 +31,7 @@ class ControlFlowGraph:
         An "empty" node for this purpose is when source has no statements.
         """
         if source.statements == []:
-            if source.predecessors == []:   # implies source == empty start node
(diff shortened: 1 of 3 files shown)


kavins14:
I feel that checking the `expected edges` in the `while` visitor tests can be done with fewer lines of code by comparing the block ids (instead of the statements). However, I decided to write them this way is because it is much easier to understand. Also, checking for `expected blocks` is somewhat implied when checking `expected edges` but I still decided to include them because it would allow us to debug block-related errors more easily.

python_ta/cfg/graph.py
```diff
@@ -31,9 +31,12 @@ def link_or_merge(self, source: CFGBlock, target: CFGBlock) -> None:
         An "empty" node for this purpose is when source has no statements.
         """
         if source.statements == []:
-            for edge in source.predecessors:
-                edge.target = target
-                target.predecessors.append(edge)
+            if source.predecessors == []:   # implies source == empty start node
```
david-yz-liu: Use `if source is self.start` here.

tests/test_cfg/test_while.py
```diff
@@ -0,0 +1,236 @@
...
+        [["print(n + 1)"], ["print(n)"]],
+        [["print(n)"], []]
+    ]
+    assert expected_edges == _extract_edges(cfg)
+
+
+def test_while_with_if() -> None:
+    src="""
+    while n > 10:
+        if (n > 20):
```
david-yz-liu: Remove the parentheses around the if condition (throughout).

tests/test_cfg/test_while.py
```diff
@@ -0,0 +1,236 @@
...
+        [["n > 10"], ["print(n)"]],
+        [["print(n)"], ["n > 10"]],
+        [["n > 10"], ["print(n + 1)"]],
+        [["print(n + 1)"], []]
+    ]
+    assert expected_edges == _extract_edges(cfg)
+
+
+def test_simple_while_with_surrounding_blocks() -> None:
+    src="""
```
david-yz-liu: Put spaces around the `=` (throughout this file).

python_ta/cfg/graph.py
```diff
@@ -31,9 +31,12 @@ def link_or_merge(self, source: CFGBlock, target: CFGBlock) -> None:
         An "empty" node for this purpose is when source has no statements.
         """
         if source.statements == []:
-            for edge in source.predecessors:
-                edge.target = target
-                target.predecessors.append(edge)
+            if source is self.start:
+                self.start = target
```
dhslim: add self.block_count = 1  here?

dhslim: add self.block_count = 1  here?

dhslim: self.block_count = 1 included here?

kavins14: Good catch, the issue with including `self.block_count = 1` is that we might have multiple instances of `CFGBlock` with the same id. This also applies if we include `self.block_count -= 1` instead (why I think would be more appropriate). A possible fix is that we use the built-in `id()` function to set the id of the blocks. Then we would be able to adjust the `block_count` attribute in the `link_or_merge` method.

david-yz-liu: Yes, this is a good question. Right now `block_count` is used as an "autoincrement" to ensure the block ids are unique, and we'll take the policy that this number never decreases. I would be open to suggestions for a different name for the attribute, or adding a comment explaining its purpose.

dhslim: I see. I thought `block_count` meant the total number of blocks in the graph. I will make a PR with the comment explaining its purpose for now.

PR #582, opened 2019-06-06
CFG Break Statement

kavins14:
Added `break` statement visitor method. I implemented this on top of the `cfg-while-node` branch.

I added a `jump` attribute to the `CFGBlock` which flags a block that includes non-compound statements that branch the control flow (ex. `break`, `continue`, `return`). Blocks with these statements (`jump` statements) cannot have a statement added to it and cannot have any out-going edges other than the one specified in the respective visitor method.

If the source code contains statement/s after a `jump` statement, they won't be added onto 
the control flow graph. Ex:
 ```
while x > 10: 
    break 
    print(x)
```
`print(x)` here won't be found in the control flow graph. This is because PyTA already reports unreachable code after `jump` statements.

Also added tests for the `break` statement in `test_while.py`.

### CODE REVIEW

File: python_ta/cfg/visitor.py [hunk 1]
```diff
@@ -1,18 +1,25 @@
...
 class CFGVisitor:
     """An astroid visitor that creates a control flow graph for a given Python module.
+
+    Private Attributes:
+    _scope: A stack of the scopes the visitor is currently in. The top element refers
+            to the current scope. (scope_type [for, while],  start_block, end_block)
     """
     cfg: ControlFlowGraph
     _current_block: CFGBlock
+    _scope: List[Tuple[str, CFGBlock, CFGBlock]]
```
david-yz-liu: Names are hard. Use the name `_control_boundaries` for now.

The idea of using a stack is excellent, but I want you to change the inner type as follows:
1. The first element of the tuple should be a `NodeNG`---that is, set it to the be node itself, not a name for the node.
2. The second element should be a dictionary mapping *node class names* to `CFGBlock`s, which obeys the intuitive rule "when the current statement has name `s`, the current block is connected to the dictionary's value corresponding to `s`. For example, you'd store a `While` node's "scope" as something like `(node, {'Break': after_while_block, 'Continue': test_block})`. 

Note that you can access the name of an object's class by doing `type(x).__name__`.

python_ta/cfg/visitor.py
```diff
@@ -58,3 +65,49 @@ def visit_if(self, node: astroid.If) -> None:
...
+        end_else = self._current_block
+
+        self.cfg.link_or_merge(end_else, after_while_block)
+        self._current_block = after_while_block
+
+    def visit_break(self, node: astroid.Break) -> None:
+        old_curr = self._current_block
+        old_curr.add_statement(node)
+        loop_scope = None
+        for scope in self._scope:
```
david-yz-liu: Set the loop in reverse order (you can do `reversed(self._scope)` here. This way you can `break` (or early `return`) the first time a loop node is encountered. You can even use an `else` attached to the loop!

Also you can use tuple unpacking right in the for loop: `for node, exits in ...`.

python_ta/cfg/visitor.py
```diff
@@ -58,3 +65,49 @@ def visit_if(self, node: astroid.If) -> None:
...
+            if scope[0] in ['while', 'for']:
+                loop_scope = scope
+        if not loop_scope:
+            raise SyntaxError('\'break\' outside loop')
+        self.cfg.link_or_merge(old_curr, loop_scope[2])
+        self._current_block.jump = node
```
david-yz-liu: This line should be removed (see above comments about `jump`).

Also, I actually want to preserve any statements after the `break` in a separate (unreachable) node. This is actually useful to capture for later to represent exactly which statements are unreachable. You can support this by initializing a new empty node right at the end of this method, and setting `self._current_block` to that new empty node.

python_ta/cfg/visitor.py
```diff
@@ -58,3 +65,49 @@ def visit_if(self, node: astroid.If) -> None:
...
+
+        self.cfg.link_or_merge(end_else, after_while_block)
+        self._current_block = after_while_block
+
+    def visit_break(self, node: astroid.Break) -> None:
+        old_curr = self._current_block
+        old_curr.add_statement(node)
+        loop_scope = None
+        for scope in self._scope:
+            if scope[0] in ['while', 'for']:
```
david-yz-liu: Similar to a previous comment, don't do anything with `for` loops for now; only check for `While` nodes here. When adding future support for `For` nodes, this part can be adapted accordingly.

python_ta/cfg/visitor.py
```diff
@@ -102,12 +104,12 @@ def visit_while(self, node: astroid.While) -> None:
...
-        for scope in self._scope:
-            if scope[0] in ['while', 'for']:
-                loop_scope = scope
-        if not loop_scope:
-            raise SyntaxError('\'break\' outside loop')
-        self.cfg.link_or_merge(old_curr, loop_scope[2])
-        self._current_block.jump = node
+        for boundary, exits in reversed(self._control_boundaries):
+            if isinstance(boundary, astroid.While):
+                self.cfg.link_or_merge(old_curr, exits[type(node).__name__])
```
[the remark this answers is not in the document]
kavins14: I've decided to go with (1) since it isn't as clunky as (2) and also because it could possibly become handy in the future.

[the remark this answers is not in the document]
david-yz-liu: @kavins14 yes that sounds good.

PR #585, opened 2019-06-12
CFG Function Definition Visitor

kavins14:
Note: This visitor does not handle function calls only function definitions. 

Functions (or Methods) by themselves are essentially mini-programs, and because of that, I decided to create a control flow graph for each individual function (or method). Conventionally, a control flow graph is built from the `main` function of the program which results in a huge control flow graph that encapsulates all the functions used in the program. I have opted against that implementation because of the following reasons:

1. A program can include code from multiple source files but `pyta` only checks individual source files. 
2. A lot of source files (based on  CSC148) primarily have class definitions that are used in other files, and as a result, don't have a `main` function.

### DIFF
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -12,14 +12,16 @@ class CFGVisitor:
         The top of the stack corresponds to the end of the list.
         (compound statement [while], {'Break'/'Continue': CFGBlock to link to})
     """
-    cfg: ControlFlowGraph
+    cfgs: List[ControlFlowGraph]
+    _current_cfg: ControlFlowGraph
     _current_block: CFGBlock
     _control_boundaries: List[Tuple[NodeNG, Dict[str, CFGBlock]]]
 
     def __init__(self) -> None:
         super().__init__()
-        self.cfg = ControlFlowGraph()
-        self._current_block = self.cfg.start
+        self.cfgs = []
+        self._current_cfg = None
+        self._current_block = None
         self._control_boundaries = []
 
     def __getattr__(self, attr: str):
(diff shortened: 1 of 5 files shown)


python_ta/cfg/visitor.py
```diff
@@ -12,14 +12,16 @@ class CFGVisitor:
         The top of the stack corresponds to the end of the list.
         (compound statement [while], {'Break'/'Continue': CFGBlock to link to})
     """
-    cfg: ControlFlowGraph
+    cfgs: List[ControlFlowGraph]
```
david-yz-liu: This is a really good start. However, instead of a list, I'd like you to use a dictionary `Dict[Node, ControlFlowGraph]`, where the key is a `Module` or `FunctionDef` node, and the corresponding value is the CFG for that node. This will be helpful in identifying these different graphs later.

python_ta/cfg/visitor.py
```diff
@@ -12,14 +12,16 @@ class CFGVisitor:
...
+    _current_cfg: ControlFlowGraph
```
david-yz-liu: The type here should be `Optional[ControlFlowGraph]`. Similarly, the `_current_block` attribute should also be `Optional`.

python_ta/cfg/visitor.py
```diff
@@ -33,19 +35,42 @@ def visit_generic(self, node: NodeNG) -> None:
...
         for child in module.body:
             child.accept(self)
-        self.cfg.link_or_merge(self._current_block, self.cfg.end)
+
+        self._current_cfg.link_or_merge(self._current_block, self._current_cfg.end)
+
+    def visit_functiondef(self, func: astroid.FunctionDef) -> None:
+        previous_cfg = self._current_cfg
+        previous_block = self._current_block
+        previous_block.add_statement(func)
```
david-yz-liu: Do this line at the very top of the function (`self._current_block.add_statement`) to distinguish it from the rest of the function.

python_ta/cfg/visitor.py
```diff
@@ -33,19 +35,42 @@ def visit_generic(self, node: NodeNG) -> None:
...
+
+        self.cfgs.append(ControlFlowGraph())
+        self._current_cfg = self.cfgs[-1]
+
+        self._current_block = self._current_cfg.start
```
david-yz-liu: Don't update `self._current_block` here. Just do `self._current_cfg.start.add_statement(...)`.

python_ta/cfg/visitor.py
```diff
@@ -33,19 +35,42 @@ def visit_generic(self, node: NodeNG) -> None:
...
+        self._current_block.add_statement(func)
```
david-yz-liu: But also, don't add `func`. You can add the `Arguments` node specifically.

python_ta/cfg/visitor.py
```diff
@@ -33,19 +35,42 @@ def visit_generic(self, node: NodeNG) -> None:
...
+
+        body_block = self._current_cfg.create_block(self._current_block)
```
david-yz-liu: Don't use a temporary variable here.

sample_usage/draw_cfg.py
```diff
@@ -41,7 +41,7 @@ def main(filepath: str) -> None:
     visitor = CFGVisitor()
     mod.accept(visitor)
 
-    display(visitor.cfg, filename)
+    display(visitor.cfgs[0], filename)
```
david-yz-liu: @kavins14 this should be `cfgs[mod]` now

PR #587, opened 2019-06-13
CFG For Visitor

dhslim:
Pretty much the same as the while visitor. Thanks @kavins14!

### DIFF
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -104,11 +104,11 @@ class CFGVisitor:
 
     def visit_for(self, node: astroid.For) -> None:
         old_curr = self._current_block
+        old_curr.add_statement(node.iter)
 
         # Handle "test" block
         test_block = self.cfg.create_block()
         test_block.add_statement(node.target)
-        test_block.add_statement(node.iter)
         self.cfg.link_or_merge(old_curr, test_block)
 
         after_for_block = self.cfg.create_block()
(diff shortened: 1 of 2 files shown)


tests/test_cfg/test_for.py
```diff
@@ -0,0 +1,604 @@
...
+    for i, j in zip(range(10), range(20)):
+        print(i, j)
+    """
+    cfg = build_cfg(src)
+    expected_blocks = [
+        ["(i, j)", "zip(range(10), range(20))"],
+        ["print(i, j)"],
+        []  # end block
+    ]
+    print(_extract_blocks(cfg))
```
david-yz-liu: Delete this line.

python_ta/cfg/visitor.py
```diff
@@ -102,6 +102,42 @@ def visit_while(self, node: astroid.While) -> None:
         self.cfg.link_or_merge(end_else, after_while_block)
         self._current_block = after_while_block
 
+    def visit_for(self, node: astroid.For) -> None:
+        old_curr = self._current_block
+
+        # Handle "test" block
+        test_block = self.cfg.create_block()
+        test_block.add_statement(node.target)
+        test_block.add_statement(node.iter)
```
david-yz-liu: Everything is good, except this should be merged with the block above it. The iterable is only evaluated once (before the loop), not at every iteration. You can keep the `node.target` where it currently is. Note that this will cause all the tests to be updated too.

PR #588, opened 2019-06-13
CFG Return Statement

kavins14:
The `return` statement is a jump statement that when used in a function links the current block to the end block of the function CFG. A `SyntaxError` will be raised if it is used outside of a function.

### DIFF
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -130,7 +130,7 @@ class CFGVisitor:
     def visit_break(self, node: astroid.Break) -> None:
         self._visit_continue_or_break(node)
 
-    def visit_continue(self, node:astroid.Continue) -> None:
+    def visit_continue(self, node: astroid.Continue) -> None:
         self._visit_continue_or_break(node)
 
     def _visit_continue_or_break(self, node: Union[astroid.Break, astroid.Continue]) -> None:
@@ -145,3 +145,11 @@ class CFGVisitor:
         unreachable_block = self._current_cfg.create_block()
         self._current_block = unreachable_block
 
+    def visit_return(self, node: astroid.Return) -> None:
+        if type(self._current_cfg.start.statements[0]).__name__ != astroid.FunctionDef.__name__:
+            raise SyntaxError(f'\'{type(node).__name__}\' outside function')
+        old_curr = self._current_block
+        self._current_cfg.link(old_curr, self._current_cfg.end)
+        old_curr.add_statement(node)
+        unreachable_block = self._current_cfg.create_block()
+        self._current_block = unreachable_block
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -12,14 +12,16 @@ class CFGVisitor:
         The top of the stack corresponds to the end of the list.
         (compound statement [while], {'Break'/'Continue': CFGBlock to link to})
     """
-    cfg: ControlFlowGraph
+    cfgs: List[ControlFlowGraph]
+    _current_cfg: ControlFlowGraph
     _current_block: CFGBlock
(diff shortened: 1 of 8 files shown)


tests/test_cfg/test_functions.py
```diff
@@ -3,7 +3,7 @@
 from python_ta.cfg import CFGVisitor, ControlFlowGraph
 
 
-def build_cfgs(src: str) -> ControlFlowGraph:
+def build_cfgs(src: str) -> List[ControlFlowGraph]:
```
david-yz-liu: Don't change this here. You can change it in your other PR.

python_ta/cfg/visitor.py
```diff
@@ -130,7 +130,7 @@ def visit_while(self, node: astroid.While) -> None:
     def visit_break(self, node: astroid.Break) -> None:
         self._visit_continue_or_break(node)
 
-    def visit_continue(self, node:astroid.Continue) -> None:
+    def visit_continue(self, node: astroid.Continue) -> None:
```
david-yz-liu: Don't make this change here. You can submit a separate pull request for this.

python_ta/cfg/visitor.py
```diff
@@ -145,3 +145,11 @@ def _visit_continue_or_break(self, node: Union[astroid.Break, astroid.Continue])
         unreachable_block = self._current_cfg.create_block()
         self._current_block = unreachable_block
 
+    def visit_return(self, node: astroid.Return) -> None:
+        if type(self._current_cfg.start.statements[0]).__name__ != astroid.FunctionDef.__name__:
```
david-yz-liu: Overall I'd say it's better not to use this approach, but instead treat a `FunctionDef` as a new *control boundary* and add an exit point for `'Return'` to jump to the end of the function.

python_ta/cfg/visitor.py
```diff
@@ -55,67 +84,72 @@ def visit_if(self, node: astroid.If) -> None:
...
-        self._visit_continue_or_break(node)
+        self._visit_jump(node)
+
+    def visit_return(self, node: astroid.Return) -> None:
+        self._visit_jump(node)
 
-    def _visit_continue_or_break(self, node: Union[astroid.Break, astroid.Continue]) -> None:
+    def _visit_jump(self, node: Union[astroid.Break, astroid.Continue, astroid.Return]) -> None:
         old_curr = self._current_block
+        req_boundary = astroid.FunctionDef if isinstance(node, astroid.Return) else astroid.While
```
david-yz-liu: This refactoring is a great start, but awkward as the cases grow larger (e.g., with @dhslim's addition of support for `For` nodes).

But it turns out you don't need to actually check what the `req_boundary` needs to be at all. Instead, rely on the fact that the *second* element of the tuple is a dictionary mapping names of node classes to which block to connect to. So you just need to check whether `type(node).__name__ in exits` in the loop below.

(This begs the question of whether the first tuple element is necessary at all. I'm not sure, so we can leave it in for now.)

python_ta/cfg/visitor.py
```diff
@@ -163,20 +167,26 @@ def visit_for(self, node: astroid.For) -> None:
...
+        self._visit_jump(node)
 
-    def _visit_continue_or_break(self, node: Union[astroid.Break, astroid.Continue]) -> None:
+    def _visit_jump(self, node: Union[astroid.Break, astroid.Continue, astroid.Return]) -> None:
         old_curr = self._current_block
         for boundary, exits in reversed(self._control_boundaries):
-            if isinstance(boundary, (astroid.While, astroid.For)):
+            if type(node).__name__ not in exits:
+                continue
+            if isinstance(boundary, (astroid.For, astroid.While, astroid.FunctionDef)):
```
david-yz-liu: You can remove this check (it should be implied by the previous check being false. This means that you can combine this part and the previous check together.

PR #590, opened 2019-06-24
CFG Class Definition Visitor

kavins14:
The visitor traverses through the methods of the class and constructs a control flow graph for each method. Unlike for the module scope, the class scope does not have a control flow graph of its own. This is because a class's functionality is almost always within a method (at least for CSC108/148). The only case that I find of code not being in a method is when defining class attributes (ex. if expressions).

### DIFF
--- a/python_ta/cfg/graph.py
+++ b/python_ta/cfg/graph.py
@@ -107,12 +107,15 @@ class CFGBlock:
         if not self.is_jump():
             self.statements.append(statement)
 
+    @property        
+    def jump(self) -> Optional[NodeNG]:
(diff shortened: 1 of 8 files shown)


david-yz-liu:
@kavins14 this is okay, but again please rebase from master.

kavins14:
This is my first time rebasing (every other time I would merge master) so I'm not quite sure if this is right. Nonetheless, I have resolved every conflict (when rebasing) and it should be the most updated version.

david-yz-liu:
@kavins14 oh sorry I thought you rebased last time, I forgot. It's okay to keep doing merges!

PR #595, opened 2019-07-02
Adding Flags to CFG Blocks

kavins14:
While working on rewriting the checker for the [always_returning_checker](https://github.com/pyta-uoft/pyta/blob/master/python_ta/checkers/always_returning_checker.py) using the control flow graph implementation, I realized how complicated looking for the starting node of a for/while node could get.

Looking for the starting node (of a loop) without flags could go like this:

Check every block in the cfg to see if the last statement `s` has an ancestor that's a loop. If yes, check if `s` is a descendant of the test node of the loop. If yes, then you have the starting node. However, this is not enough since test nodes that are expanded (like `BoolOp` and potentially `If expressions`   ) would require further checks.

I find this to be unnecessarily complicated (nested loops would complicate things even further) and so decided to add flags to blocks indicating whether the block is a start/end of a loop. This will simplify checks we do specifically on loops. 

I opted for flagging blocks over having the `control_boundaries` attribute similar to the one in the `CFGVisitor` class because it would also help with data flow analysis. One example would be when performing unreachable code analysis, we would be able to identify the starting/ending node of branching statements (loops, if, try/catch) while traversing through the graph itself.

### DIFF
--- a/python_ta/cfg/graph.py
+++ b/python_ta/cfg/graph.py
@@ -145,5 +145,3 @@ class CFGEdge:
 class Flag(enum.Enum):
     LOOP_START = 'LOOP_START'
     LOOP_END = 'LOOP_END'
-    IF_START = 'IF_START'
-    IF_END = 'IF_END'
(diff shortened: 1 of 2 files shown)


david-yz-liu:
I'll discuss this with you further on Slack! See #596.

PR #597, opened 2019-07-03
One Iteration Checker

kavins14:
The method used to deduce if a loop only executes one iteration is outlined in the docstring of the `_check_one_iteration` function. 

Note that test cases for this implementation have not yet been written.


python_ta/checkers/one_iteration_checker.py
```diff
@@ -25,20 +26,59 @@ class OneIterationChecker(BaseChecker):
     # pass in message symbol as a parameter of check_messages
     @check_messages('one-iteration')
     def visit_for(self, node):
-        if self._check_one_iteration(node):
+        if hasattr(node, 'cfg_block') and self._check_one_iteration(node):
```
kavins14: This check is added because `While` and `For` nodes defined in the class scope don't have a control flow graph. To avoid this altogether (and for all other checks using the CFG), I think it would be good to have a control flow graph for class definitions. If you (@david-yz-liu) think so too, then I'll work on implementing it.

python_ta/checkers/one_iteration_checker.py
```diff
@@ -25,20 +26,63 @@ class OneIterationChecker(BaseChecker):
...
+        <_pred> is the predecessor of <block>.
+
+        """
+        if block is _pred:
+            return True
+        elif block.predecessors == []:
+            return False
+        else:
+            # The left-most predecessor ensures we won't step into an
+            # unreachable block of a nested loop.
```
kavins14: This holds because the start block of a while/for subgraph has to already have a predecessor: https://github.com/pyta-uoft/pyta/blob/9980e44f051722abc45d0afead2ca87539c8f27d/python_ta/cfg/visitor.py#L116

This implies that the edge linking to an unreachable block will not be the first element in the `predecessors` list.

python_ta/checkers/one_iteration_checker.py
david-yz-liu: @kavins14 Sorry I missed this earlier. The `ClassDef` node shouldn't have its own CFG, but the statements in it should be part of the CFG of the "main" module. (This is actually how the control flow works in Python; statements in the class body are indeed evaluated when the ClassDef is reached.)

So in fact, nothing special should be done; in the `ClassDef` visitor, you can just remove the check that looks for `FunctionDef` nodes, and call `accept` with all the nodes in the body.

david-yz-liu: (And therefore revert the change on this line)

python_ta/checkers/one_iteration_checker.py
```diff
@@ -25,20 +26,63 @@ class OneIterationChecker(BaseChecker):
...
+        start = node.target if isinstance(node, astroid.For) else node
+        preds = start.cfg_block.predecessors
+
+        if preds == []:
+            return False
+
+        for pred in preds:
+            pred = pred.source
+            if node.parent_of(pred.statements[0]) and \
+                    self.is_predecessor(pred, start.cfg_block):
```
david-yz-liu: I think I'm forgetting something here; why is the `self.is_predecessor` check necessary? I.e., why isn't it enough that a predecessor contains a statement that's inside the given loop?

kavins14: The reason behind this is because of how unreachable blocks (statements after `jump` statements) within the loops are handled. If there is a statement `s` after a jump statement then it will be in a separate block that 'points' to the start block of the loop. This occurs because of these two lines:
https://github.com/pyta-uoft/pyta/blob/a36e750b130c4ab23294f24273016e9e96ea2687/python_ta/cfg/visitor.py#L201-L202

This means that a predecessor to the loop start block could be one of the unreachable blocks that won't be executed.

python_ta/checkers/one_iteration_checker.py
kavins14: Correction: This does not hold for all cases. For instance when having statements after jump statements within a nested `If` statement. I've made the necessary changes.

python_ta/checkers/one_iteration_checker.py
kavins14: https://github.com/pyta-uoft/pyta/pull/600

kavins14:
Note: I will update this implementation once #602 is merged.

david-yz-liu:
@kavins14 I merged in #602!

PR #598, opened 2019-07-04
Visit tryexcept

dhslim:
A visitor for a try-except node.

There are three cases in the execution of the try clause:
1. No exception occurs
2. An exception occurs but is handled.
3. An exception occurs but can't be handled.

So every statement in the try clause will have 3 control flow edges 1 for each of the cases.

Nesting:
When there are nested try-except nodes, there is recursion. Here is an example:
```print('before try')
try:
    outer()
    try:
        inner()
        try:
            innermost()
        except ValueError:
            print('innermost error')
    except TypeError:
        print('inner error')
except NameError:
    print('outer error')
else:
    print('no error')
print('after try')
```

image

Fixed duplicate edges and blank box issue.


python_ta/cfg/visitor.py
```diff
@@ -201,3 +201,50 @@ def _visit_jump(self, node: Union[astroid.Break, astroid.Continue, astroid.Retur
...
+
+        # Create blocks for handlers and store them
+        handlers = []
+        for handler in node.handlers:
+            self._current_block = self._current_cfg.create_block()
+            handler.accept(self)
+            handlers.append(self._current_block)
+            self._current_cfg.link_or_merge(self._current_block, after_tryexcept_block)
+
+        for child in node.body:
```
david-yz-liu: @dhslim This part is interesting. You are trying to capture the possibility that *every* statement in the body of the `try` might raise an error, and so dividing up each statement into its own block so that you can connect it to each of the handlers, or the `orelse`.

This currently is inaccurate in two ways: first, some statements might _never_ raise an exception (e.g., `x = 1`), and second, compound statements (e.g., with an `if` block or loop) will be treated as a whole block, rather than being split up as individual statements.

So for now I prefer that you simplify this implementation by _not_ creating a separate block for each statement in the `node.body`, but simply call `child.accept` for every statement in the body, and then connect the resulting block to the different exception handlers and the `orelse`. This is also a simplification, but the benefit is it'll lead to a simpler implementation that we can build on later.

PR #599, opened 2019-07-19
Possibly Undefined Checker

kavins14:
This checker raises an error if a statement `s` in a program uses a variable `v` that might not be defined. That is if there exists a path from the start block to `s` such that `v` might not be defined. In functions, `v` would pertain to only local variables while in the module-level scope it would pertain to only global variables.

Data flow algorithm: https://www.seas.harvard.edu/courses/cs252/2011sp/slides/Lec02-Dataflow.pdf#page=31

Because of the complexity of the implementation (which could result in false positives/negatives), I will be writing tests for the checker using the guidelines here (under `testing checkers`): http://pylint.pycqa.org/en/latest/how_tos/custom_checkers.html

Note: I named the checker `possibly_undefined` but I think there would be a more appropriate name. Also, the checker `id` I put was arbitrary.


kavins14:
Here's an edge case that the checker does not identify ('x' might not be defined):
Screen Shot 2019-07-16 at 5 53 28 PM
I realize that we can't possibly cover all false negatives but as long as we don't have false positives I think we should be fine.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,157 @@
...
+    # use dashes for connecting words in message symbol
+    msgs = {'E9969': ('This variable might not be defined when this statement is executed.',
+                      'possibly-undefined',
+                      'Reported when a statement uses a variable that might not be assigned.'),
+           }
+
+    # this is important so that your checker is executed before others
+    priority = -1
+
+    possibly_undefined = set()
```
david-yz-liu: Don't make this a class variable.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,157 @@
...
+
+    @check_messages('possibly-undefined')
+    def visit_name(self, node):
+        if self._check_possibly_undefined(node):
```
david-yz-liu: Don't create a separate helper for this, just do the check inline.

python_ta/checkers/possibly_undefined_checker.py
david-yz-liu: (And also document the type of this attribute.)

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        facts = {}
+        blocks = self._get_blocks_po(node)
+
+        all_assigns = self._get_assigns(blocks)
+        for block in blocks:
+            facts[block] = {}
+            facts[block]['out'] = all_assigns.copy()
+
+        worklist = blocks[:]
```
david-yz-liu: As far as I can tell `blocks` isn't used after this, which means you shouldn't need to create a copy here.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,157 @@
...
+            b = worklist.pop()
+            outs = [facts[p.source]['out'] for p in b.predecessors if p.source in facts]
+            if outs == []:
+                facts[b]['in'] = set()
+            else:
+                facts[b]['in'] = set.intersection(*outs)
+            temp = self._transfer(b, facts[b]['in'], all_assigns)
+            if temp != facts[b]['out']:
+                facts[b]['out'] = temp
+                successors = set([succ.target for succ in b.successors])
```
david-yz-liu: You should be able to remove the square brackets here.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,157 @@
...
+        gen = set()
+        kill = set()
+        for statement in block.statements:
+            if isinstance(statement, astroid.Assign):
+                gen.update(self._get_targets(statement))
+            elif isinstance(statement, astroid.AnnAssign) and hasattr(statement.target, 'name'):
+                gen.add(statement.target)
+            elif isinstance(statement, astroid.Delete):
+                kill.update(self._get_targets(statement))
+            self._analyze_statement(statement, gen.union(in_facts.difference(kill)), all_assigns)
```
david-yz-liu: This part can be simplified; it shouldn't be necessary to do a full recursion, but instead do a `for name in statements.nodes_of_class(astroid.Name): ...`.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,157 @@
...
+        return gen.union(in_facts.difference(kill))
+
+    def _analyze_statement(self, node: NodeNG, facts: Set[str], local_vars: Set[str]) -> None:
+        if isinstance(node, astroid.Name):
+            name = node.name
+            if name in local_vars and name not in facts and not self._is_function_name(node):
```
david-yz-liu: There shouldn't need to be a special case for names of functions. It seems like `_transfer` above isn't handling the possibility that a statement in a block is a `FunctionDef`, which is another way of introducing a local name.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,157 @@
...
+                self.possibly_undefined.add(node)
+            elif node in self.possibly_undefined:
+                self.possibly_undefined.remove(node)
+        else:
+            for child in node.get_children():
+                self._analyze_statement(child, facts, local_vars)
+
+    def _get_targets(self, statement: Union[astroid.Assign, astroid.Delete]) -> Set[str]:
```
david-yz-liu: There's probably a more natural astroid way of doing this. At the very least, using `nodes_of_class(astroid.AssignName)`, but could be even better.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,158 @@
...
+from typing import Set, List
+
+
+class PossiblyUndefinedChecker(BaseChecker):
+
+    __implements__ = IAstroidChecker
+    # name is the same as file name but without _checker part
+    name = 'possibly_undefined'
+    # use dashes for connecting words in message symbol
+    msgs = {'E9969': ('This variable might not be defined when this statement is executed.',
```
kavins14: What message-id should it be? @david-yz-liu

python_ta/checkers/possibly_undefined_checker.py
kavins14: The `_transfer` function loops over all the statements in the block, so it is possible that you come across a `FunctionDef` node (nested function). However, The reason behind having this extra check is that a `Name` node could be the name of a function in a function call. For example the ast representation of  `print(x)` would be `Call(n=Name('print'),args=Name('x'))`. This means that the `name` variable in the code above could be storing `'print'`, a function name. It gets tricky when the function name is equivalent to a local variable name. In that case, there would be no difference (in the `Name` node) between a function name and a variable name.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,157 @@
...
+        visited.add(block.id)
+        blocks = []
+
+        for succ in block.successors:
+            blocks.extend(self._get_blocks(succ.target, visited))
+        blocks.append(block)
+
+        return blocks
+
+    def _get_child_nodes(self, node: NodeNG, klasses, skip_klasses=None):
```
kavins14: This method has the absolute same implementation as the `NodeNG.nodes_of_class` method. The reason I chose to rewrite the method here is because its public interface didn't support multiple classes to select (or ignore) even though the implementation apparently does. I think it shouldn't be a problem to actually use the `nodes_of_class` method since it's highly improbable that the creators of astroid will change the implementation, however, I feel that with an explicit implementation we are more certain that our code will work the way we intend it to. What do you think? @david-yz-liu

kavins14:
I wrote a test suite for this checker because I felt that even a tiny change to the implementation could cause some edge case to occur. I also plan to continue writing test suites for custom checkers with fairly complex implementations.

P.S. I think Travis CI is having issues with how the test cases for the custom checkers are written.

python_ta/checkers/possibly_undefined_checker.py
david-yz-liu: Great point! I think it's fine to use the `nodes_of_class` method. Pylint itself has a couple of instances of using it with multiple classes. If you want to, you could submit an issue or pull request in the astroid repo to update the documentation. (Note that `skip_klasses` can also be a tuple.)

python_ta/checkers/possibly_undefined_checker.py
david-yz-liu: `E9969` is okay with me, we don't currently have a systematic way of naming new errors.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,140 @@
...
+                else:
+                    kill.add(node.name)
+            self._analyze_statement(statement, gen.union(in_facts.difference(kill)), all_assigns)
+        return gen.union(in_facts.difference(kill))
+
+    def _analyze_statement(self, statement: NodeNG, facts: Set[str], local_vars: Set[str]) -> None:
+        nodes = statement.nodes_of_class(astroid.Name)
+        for node in nodes:
+            name = node.name
+            if not self._is_function_name(node) and name in local_vars and name not in facts:
```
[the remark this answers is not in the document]
kavins14: I was able to look into this issue and it seems like their implementation is buggy when it comes to handling `global` and `nonlocal` statements. There is an issue to remove support for them (https://github.com/PyCQA/astroid/issues/250) but nothing has been done about it yet. It seems like an interesting issue to look into.

Either way, I think our handling of `nonlocal` and `global` is sufficient.

P.S. This is for future reference.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,121 @@
...
+    def _analyze(self, node: Union[astroid.Module, astroid.FunctionDef]) -> None:
+        """Runs the data flow algorithm on a `Module` or `Function` CFG, which in turn
+        appends `Name` nodes to `self.possibly_undefined` if it might not be defined.
+
+        Data flow algorithms retrieved from:
+        https://www.seas.harvard.edu/courses/cs252/2011sp/slides/Lec02-Dataflow.pdf#page=31
+        """
+        out_facts = {}
+        cfg = ControlFlowGraph()
+        cfg.start = node.cfg_block
```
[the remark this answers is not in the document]
david-yz-liu: Please revert this change. At this point you should be limiting your changes to only comments I make.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,119 @@
...
+                continue
+            for node in statement.nodes_of_class((astroid.AssignName, astroid.DelName, astroid.Name),
+                                              astroid.FunctionDef):
+                if isinstance(node, astroid.AssignName):
+                    gen.add(node.name)
+                elif isinstance(node, astroid.DelName):
+                    kill.add(node.name)
+                else:
+                    name = node.name
+                    if not self._is_function_name(node) and name in local_vars \
```
david-yz-liu: I did some more digging about this builtin function issue to see what pylint does. The short version is that I want you to replace `_is_function_name` with the three checks done [here](https://github.com/PyCQA/pylint/blob/master/pylint/checkers/variables.py#L1660). It's also worth your time to read through that `visit_name` function to look at the different cases pylint already handles. In particular, good to think about recursive function calls and using the class name in a type annotation.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,119 @@
...
+        https://www.seas.harvard.edu/courses/cs252/2011sp/slides/Lec02-Dataflow.pdf#page=31
+        """
+        out_facts = {}
+        blocks = list(node.cfg.get_blocks_postorder())
+        blocks.reverse()
+
+        all_assigns = self._get_assigns(node)
+        for block in blocks:
+            out_facts[block] = {}
+            out_facts[block] = all_assigns.copy()
```
david-yz-liu: I'd like to double-check this---isn't it sufficient to initialize `out_facts[block]` to be empty? It seems like the main work should be done in the `_transfer` to accumulate the actual `AssignName` nodes that populate the out facts.

Either way, lines 56 and 57 conflict, so one of them should be removed.

kavins14: According to the reference, the `outs` must be set to the top of the lattice (i.e all the local variables).

A case where `outs` being empty is not sufficient is when there's a back-edge.
 
```
y = 0
while y > 10:
    x = 0
```

cfg_misc

Since the `join` operation is set `intersection`, If `outs` wasn't set to all the local variables then `y` won't be definitely assigned at the loop condition block.

python_ta/checkers/possibly_undefined_checker.py
kavins14: I wasn't able to add one [check](https://github.com/PyCQA/pylint/blob/6b3afd4f6d75debd7f286f0d3c760ed10ab1e79f/pylint/checkers/variables.py#L1662) from the three you mentioned. I wasn't able to find the `additional_builtins` attribute definition in the entire pylint project. According to the documentation, this attribute stores additional user-defined builtins and by default it an empty tuple. 

Either way, there's still an issue with user-defined function names that conflict with variable names.

There is currently a failing test: 
https://github.com/pyta-uoft/pyta/blob/759455aa030288e470b5167a0ef64a39df452e5e/tests/test_custom_checkers/test_possibly_undefined_checker.py#L162-L176

python_ta/checkers/possibly_undefined_checker.py
david-yz-liu: Alright, that makes sense to me. Thanks!

python_ta/checkers/possibly_undefined_checker.py
david-yz-liu: The `additional_builtins` is a pylint configuration option (see [here](http://pylint.pycqa.org/en/latest/technical_reference/features.html#variables-checker-options); these are actually added dynamically from config files (e.g., `.pylintrc`), which is why you can't find it in the pylint project. You should add it in.

I'm not sure exactly what you're pointing out with the test there. Is the problem that the test as written is _passing_ but it should actually be failing?

kavins14: This test as written is not passing but if the implementation were correct (implying it currently isn't) it should be passing. The reason is that `y` in `y()` is a `Name(name=y)` node but there isn't a check to see if it's a function name so the algorithm thinks that a variable (that is not definitely assigned) is being used. 

The checks you have mentioned would be enough to filter out builtin function call name nodes however when it comes to user-defined functions, these kinds of cases can occur.

david-yz-liu: I see. You should keep in mind that functions aren't that special---variables can refer to functions, and this is perfectly okay. So for example you could change the test to `y = lambda: 20` and the function type-checks properly, except that `y` might not be defined.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -0,0 +1,119 @@
...
+                                              astroid.FunctionDef):
+                if isinstance(node, astroid.AssignName):
+                    gen.add(node.name)
+                elif isinstance(node, astroid.DelName):
+                    kill.add(node.name)
+                else:
+                    name = node.name
+                    # comment out 'self.config....' check when running tests
+                    if not (name in astroid.Module.scope_attrs or utils.is_builtin(name) or
+                            name in self.config.additional_builtins ) \
```
kavins14: This specific check throws an attribute error:
Screen Shot 2019-09-22 at 3 22 01 PM

david-yz-liu: @kavin Got it, I realized what was missing. If you look at the [pylint code](https://github.com/PyCQA/pylint/blob/master/pylint/checkers/variables.py#L600), it includes a class variable `options` that contains a tuple specifying that the option "additional-builtins" is used. This is what initializes `self.config.additional_builtins`. Please add the same code to this checker too.

david-yz-liu:
I also (finally!) looked into why the tests were failing. I think you're using `pytest` to run the tests, but "officially" PyTA uses `nosetest`, which is a little different. Pylint uses [`setup_method`](https://github.com/PyCQA/pylint/blob/b72c0dd9a3ee5c992f94937b58de3b2153f4624d/pylint/testutils.py#L222) to specify the code to run before each test, and this is what `pytest` uses; but `nosetest` uses `setUp`, same as the built-in `unittest` module. So you can fix this by adding to your class something like

```python
def setUp(self):
    self.setup_method()
```

python_ta/checkers/possibly_undefined_checker.py
kavins14: I added the code pylint code but this caused another error to occur. I was able to try many different ideas to solve this issue but I'm currently at a dead end. 

Screen Shot 2019-09-28 at 7 17 33 PM

I'll look into it again in a couple of days.

david-yz-liu: Okay I looked into this a bit more; looks like there's no easy way to reuse an option from an existing checker. You can remove the part that uses `additional_builtins` and the options!

PR #601, opened 2019-07-31
Unreachable Block Identification

kavins14:
Definition of `unreachable`:

> A block `b` is `unreachable` if and only if there does not exist a path from the start block to `b`.

This PR introduces a new attribute `is_unreachable` to the CFGBlock class which evaluates to _True_ if the block is unreachable and _False_ otherwise.

In our current CFG implementation, the only times we can expect an unreachable block is when there is code that comes after a jump statement (`break`, `continue` or `return`).
Ex:
```
while x > 10:
    break
    print(x)  #  this expression becomes an unreachable block
```

### DIFF
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -46,15 +46,11 @@ class CFGVisitor:
         self._current_cfg.link_or_merge(self._current_block, self._current_cfg.end)
 
     def visit_classdef(self, node: astroid.ClassDef) -> None:
-        self._current_block.add_statement(node)
-
         for child in node.body:
-            if isinstance(child, astroid.FunctionDef):
-                child.accept(self)
+            child.accept(self)
 
     def visit_functiondef(self, func: astroid.FunctionDef) -> None:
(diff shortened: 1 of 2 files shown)


david-yz-liu:
@kavins14 this is a good attempt, but determining whether a block is unreachable dynamically as the CFG is being constructed is a really hard problem, as a block can gain new predecessors which are a mix of reachable/unreachable. For example, your current implementation fails on
```python
while n > 10:
        print(n)
        if n > 20:
            print('hi')
        else:
            break
            print('bye')
        print('after')
```
because it marks `print('after')` as unreachable.

A better approach is to have a default state where all blocks are unreachable, and then when the entire module is parsed and all CFGs created, you start at the `start` of each CFG and traverse the graph to determine which nodes are reachable, by doing the same kind of traversal as in `ControlFlowGraph.get_blocks`. You can store the unreachable nodes in an instance attribute so that they aren't lost forever. (Indeed, it might be interesting to display them in the visual output of `draw_cfg`, but grayed out or something.

PR #602, opened 2019-08-01
Unreachable Block Identification

kavins14:
Continuation from #601

### DIFF
--- a/python_ta/cfg/graph.py
+++ b/python_ta/cfg/graph.py
@@ -85,6 +85,11 @@ class ControlFlowGraph:
             yield edge
             yield from self._get_edges(edge.target, visited)
 
+    def update_block_reachability(self) -> None:
+        for block in self.get_blocks():
+            block.is_unreachable = False
+            self.unreachable_blocks.remove(block)
+
 
 class CFGBlock:
     """A node in a control flow graph.
--- a/python_ta/cfg/graph.py
+++ b/python_ta/cfg/graph.py
@@ -9,11 +9,11 @@ class ControlFlowGraph:
     end: CFGBlock
     # block_count is used as an "autoincrement" to ensure the block ids are unique.
     block_count: int
-    unreachable_blocks: List[CFGBlock]
+    unreachable_blocks: Set[CFGBlock]
 
     def __init__(self) -> None:
         self.block_count = 0
-        self.unreachable_blocks = []
+        self.unreachable_blocks = set()
         self.start = self.create_block()
         self.end = self.create_block()
 
@@ -23,7 +23,7 @@ class ControlFlowGraph:
         If pred is specified, set that block as a predecessor of the new block.
         """
         new_block = CFGBlock(self.block_count)
-        self.unreachable_blocks.append(new_block)
+        self.unreachable_blocks.add(new_block)
 
         self.block_count += 1
         if pred:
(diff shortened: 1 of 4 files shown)


sample_usage/draw_cfg.py
```diff
@@ -28,13 +28,13 @@ def display(cfgs: Dict[NodeNG, ControlFlowGraph],
         else:
             continue
         with graph.subgraph(name=f'cluster_{id(node)}') as c:
-            _visit(cfg, cfg.start, c, set())
+            _visit(cfg.end, c, set())
```
david-yz-liu: Why did you switch this from start to end?

python_ta/cfg/graph.py
```diff
@@ -9,18 +9,22 @@ class ControlFlowGraph:
     end: CFGBlock
     # block_count is used as an "autoincrement" to ensure the block ids are unique.
     block_count: int
+    unreachable_blocks: List[CFGBlock]
```
david-yz-liu: Use a set instead of a list to store these. Small performance improvement.

python_ta/cfg/graph.py
```diff
@@ -95,13 +104,16 @@ class CFGBlock:
     predecessors: List[CFGEdge]
     # This block's out-edges (to blocks that can execute immediately after this one).
     successors: List[CFGEdge]
+    # Whether there exists a path from the start block to this block.
+    is_unreachable: bool
```
david-yz-liu: Generally prefer positive instead of negative versions of attributes. So here, use "reachable" instead of "unreachable" as the attribute name (this is also consistent with the description written above). Also you can drop the "is" prefix, and just use "reachable".

sample_usage/draw_cfg.py
kavins14: This is because I wanted to capture every block in the control flow graph (even the unreachable ones). My understanding is that using this traversal method starting from `start` won’t capture unreachable blocks, while starting the traversal from `end` will ensure that.

david-yz-liu: We can't assume there's a path from every block to the end node. For example,

```python
for i in range(10):
  print('reachable')
  break
  print('unreachable and doesn't reach end')
  break
  print('also unreachable, but reaches end')
```

kavins14: This is an interesting example, I ran this code on `draw_cfg.py` and this is what it returned:
Screen Shot 2019-08-08 at 3 40 55 PM
 
At first, I got totally confused but then it made sense and this is how the CFG actually should look like. 

The reason I think that there is a path from the `end` block to every block in the CFG is because of how unreachable blocks are stored. These blocks are added as predecessors to blocks that are reachable and they are not successors to the `start` block (this makes them unreachable). However, if we traverse the CFG from the `end` to the `start` we will be doing so by visiting every ancestor (i.e predecessor) block of `end`. This means that reachable blocks will be visited as well as unreachable blocks.

david-yz-liu: Ah yes, I forgot that `break` and `return` do a link to the CFG end block. It does seem true that all blocks, at least for now, have a successor that is reachable.

However, this might change in the future (e.g., if we do some of the removal of edges based on constant propagation). I'd rather not change the algorithm more than once. The reason I said to use an attribute to store the unreachable nodes is so that we don't have to rely on these nodes being reachable from any other node in the graph.

PR #606, opened 2019-08-27
Possible Infinite Loop Checker

kavins14:
This checker will report a message if a loop variable does not get mutated in the loop body. 

If a loop variable `x` is inferred as immutable, then only `AssignName(name='x')` nodes in the loop body will satisfy as a 'mutation' of the loop variable.

If a loop variable `x` is inferred as immutable, then both `Name(name='x')` & `AssignName(name='x')` nodes in the loop body will satisfy as a 'mutation' of the loop variable.

Edit: This P.R. is __not__ a work in progress anymore


kavins14:
There is an issue where the type inference reports a mutable type to be an immutable one. 
For example: 
Screen Shot 2019-08-29 at 6 22 19 PM

PR #607, opened 2019-09-03
New checker: if-return-bool

kavins14:
This PR presents a new checker that emits a message if there is only one statement `s` in both branches of an `If` such that `s` is a return statement with values either `True` or `False`.

The idea is that this:
```
if condition:
    return True
else:
    return False
```

can be simplified to:
`return condition`

Note that technically, the simplified statement does not function exactly like the previous block (ex: condition being an object instance). If the returned expression is treated specifically like a boolean then the simplification is justified. However, it is possible that the returned expression is treated something like `var = func_returning_bool()` and so the simplification can change the behavior of the program. I see this as the only problem with this particular check.

### DIFF
--- a/python_ta/__init__.py
+++ b/python_ta/__init__.py
@@ -180,7 +180,8 @@ def reset_linter(config=None, file_linted=None):
         'python_ta/checkers/structure_test_checker',
         'python_ta/checkers/type_annotation_checker',
         'python_ta/checkers/unnecessary_indexing_checker',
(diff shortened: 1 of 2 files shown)


kavins14:
I just found out that pylint already has this check [here](https://github.com/PyCQA/pylint/blob/6b3afd4f6d75debd7f286f0d3c760ed10ab1e79f/pylint/checkers/refactoring.py#L349):

PR #608, opened 2019-09-20
Visitor: Try-Except

kavins14:
Added CFG building functionality for `try/except` compound statements. 

Note that only the final block(s) of the `try` body will be linked to the `except` blocks. This is basically a subset of all the possible edges in the `try/except` CFG representation.

### DIFF
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -214,6 +214,7 @@ class CFGVisitor:
         for handler in node.handlers:
             h = self._current_cfg.create_block()
             self._current_block = h
+            handler.cfg_block = h
             for child in handler.body:
                 child.accept(self)
             end_handler = self._current_block
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -201,18 +201,16 @@ class CFGVisitor:
         self._current_block = unreachable_block
 
     def visit_tryexcept(self, node: astroid.TryExcept) -> None:
-        previous_block = self._current_block
-        self._current_block = self._current_cfg.create_block()
-        self._current_cfg.link_or_merge(previous_block, self._current_block)
+        self._current_block = self._current_cfg.create_block(self._current_block)
         node.cfg_block = self._current_block
 
         for child in node.body:
             child.accept(self)
+        end_body = self._current_block
+
+        end_block = self._current_cfg.create_block()
 
         handlers = []
-
-        end_body = self._current_block
-        end_block = self._current_cfg.create_block()
         for handler in node.handlers:
             h = self._current_cfg.create_block()
             h.statements.append(handler)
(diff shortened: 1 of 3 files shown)


python_ta/cfg/visitor.py
```diff
@@ -199,3 +199,41 @@ def _visit_jump(self, node: Union[astroid.Break, astroid.Continue, astroid.Retur
...
+            h = self._current_cfg.create_block()
+            self._current_block = h
+            handler.cfg_block = h
+            for child in handler.body:
+                child.accept(self)
+            end_handler = self._current_block
+            self._current_cfg.link_or_merge(end_handler, end_block)
+            handlers.append(h)
+
+        after_body = handlers
```
david-yz-liu: In general, prefer not creating aliases. You can name the original variable (line 213) `after_body` and stick with that throughout.

python_ta/cfg/visitor.py
```diff
@@ -199,3 +199,41 @@ def _visit_jump(self, node: Union[astroid.Break, astroid.Continue, astroid.Retur
...
+        else_block = self._current_cfg.create_block()
+        after_body.append(else_block)
+
+        self._current_cfg.multiple_link_or_merge(end_body, after_body)
+
+        if node.orelse == []:
```
david-yz-liu: It seems like a better approach is that if `node.orelse == []`, don't create an `else_block` at all (line 225). I guess in this case you would need to explicitly link `end_body` and `end_block`.

kavins14: The issue with doing a `multiple_link_or_merge` and then a `link_or_merge` with the `source` being `end_body` is that since it has no statements it will be discarded as its predecessors will be directly linked to its successors after the `multiple_link_or_merge` call. This is why everything linked after the body is in one `after_body` list. 

Now, to avoid having the `else_block` being created regardless of what `node.orelse` evaluates to, I had to move the `multiple_link_or_merge` call to right after the `node.orelse` `if` block. This creates a problem highlighted in the following example:

```
try: // 1
    pass
except:
    pass
else:
    try:  // 2
        pass
    else:
        pass
```
When the visitor method is called on the `try/except` block marked by the comment `// 2`, it will create a new block and `link_or_merge` it with the previous one. In this particular example, the previous block will be the empty `else_block` that will be discarded because of the `link_or_merge` operation. Then when the visitor method continues evaluating the `try/except` block marked by the comment `// 1`, `end_body` will be linked the discarded `else_block`. This is solved by this fix:

https://github.com/pyta-uoft/pyta/blob/6bb79f1bba37702ca3a676aaafec9bb8b4cb7c18/python_ta/cfg/visitor.py#L204

Which in turn motivates this:

https://github.com/pyta-uoft/pyta/blob/6bb79f1bba37702ca3a676aaafec9bb8b4cb7c18/python_ta/cfg/graph.py#L145-L146

Which then motivates me the replace every `CFGBlock.statement == []` simply with `not CFGBlock`. I think that a block with no statements is similar to an empty list and so this refactoring makes sense. If you say so then I'll make a PR for this right away.

python_ta/cfg/visitor.py
```diff
@@ -201,7 +201,7 @@ def _visit_jump(self, node: Union[astroid.Break, astroid.Continue, astroid.Retur
         self._current_block = unreachable_block
 
     def visit_tryexcept(self, node: astroid.TryExcept) -> None:
-        self._current_block = self._current_cfg.create_block(self._current_block)
+        self._current_block = self._current_block and self._current_cfg.create_block(self._current_block)
```
david-yz-liu: I'm not a fan of this style. Revert the `__bool__` and just do `if self._current_block.statements != []`.

PR #609, opened 2019-10-10
Redundant Assignment Checker

kavins14:
This checker reports an error if a `redundant` assignment statement is found in a python program.

For an assignment statement `s` of variable `v` to be `redundant`, it must satisfy the following properties:

> 1. Every path from `s` to `v`'s  first usage goes through at least one re-definition of `v`.

> 2. Removing the statement from the program does not affect the behavior of the program in any way.

An example:
```python
y = 10    # redundant assignment
y = 20
```

A more exciting example:
```python
t = 10    # redundant assignment
for t in range(1, 10):
    print(t)
    x = 5    # redundant assignment
else:
    x = 10
```

**Note**: only `local` assignment statements are covered in this check. This is because finding statements that satisfy property (2) is hard, however, for `local` assignments, property (1) implies property (2).

### DIFF
--- a/python_ta/checkers/redundant_assignment_checker.py
+++ b/python_ta/checkers/redundant_assignment_checker.py
@@ -1,9 +1,9 @@
-"""checker for redundant assignment in the program.
+"""checker for redundant assignment statements in the program.
 """
 from typing import Union
 import astroid
 from pylint.interfaces import IAstroidChecker
-from pylint.checkers import BaseChecker
+from pylint.checkers import BaseChecker, utils
 from pylint.checkers.utils import check_messages
 from python_ta.cfg.graph import CFGBlock, ControlFlowGraph
 from typing import Set
@@ -83,7 +83,9 @@ class RedundantAssignmentChecker(BaseChecker):
         kill = set()
         for statement in reversed(block.statements):
             if isinstance(statement, astroid.FunctionDef):
-                continue
+                for node in statement.nodes_of_class((astroid.DelName, astroid.Name, astroid.Global, astroid.Nonlocal)):
+                    if not (node.name in astroid.Module.scope_attrs or utils.is_builtin(node.name)):
+                        kill.add(node.name)
             for node in statement.nodes_of_class((astroid.AssignName, astroid.DelName, astroid.Name),
                                               astroid.FunctionDef):
                 if isinstance(node, astroid.AssignName):
@@ -96,6 +98,7 @@ class RedundantAssignmentChecker(BaseChecker):
                     gen.add(node.name)
                 else:
                     kill.add(node.name)
+
         return gen.difference(kill)
 
     def _get_assigns(self, node: Union[astroid.FunctionDef, astroid.Module]) -> Set[str]:
--- a/python_ta/checkers/redundant_assignment_checker.py
+++ b/python_ta/checkers/redundant_assignment_checker.py
@@ -15,7 +15,7 @@ class RedundantAssignmentChecker(BaseChecker):
     # name is the same as file name but without _checker part
     name = 'redundant_assignment'
     # use dashes for connecting words in message symbol
-    msgs = {'E9969': ('Description',
+    msgs = {'E9959': ('Description',
                       'redundant-assignment',
                       'Description'),
            }
@@ -29,8 +29,15 @@ class RedundantAssignmentChecker(BaseChecker):
 
     @check_messages('redundant-assignment')
     def visit_assign(self, node):
-        """Adds message if there exists a path from the start block to node where
-        a variable used by node might not be defined."""
+        """Adds message if the assignment statement is <redundant>, i.e. it satisfies
+        the following two properties:
+
+            1. Every path from a definition of a variable `v` to it's first usage
+            goes through at least one re-definition of `v`.
+
+            2. Removing the statement from the program does not in any way change
+            the behavior of the program.
+        """
         if node in self._redundant_assignment:
             self.add_message('redundant-assignment', node=node)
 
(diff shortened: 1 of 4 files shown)


python_ta/checkers/redundant_assignment_checker.py
```diff
@@ -0,0 +1,130 @@
...
+        self._analyze(node)
+
+    def _analyze(self, node: Union[astroid.Module, astroid.FunctionDef]) -> None:
+        """Runs the backward data flow algorithm on a `Module` or `Function` CFG, which in turn
+        appends `Assign` nodes to `self.redundant_assignment` if it is redundant.
+
+        Data flow algorithms retrieved from:
+        https://www.seas.harvard.edu/courses/cs252/2011sp/slides/Lec02-Dataflow.pdf#page=31
+        """
+        out_facts = {}
```
david-yz-liu: You should clearly document what the "facts" are in this algorithm.

python_ta/checkers/redundant_assignment_checker.py
```diff
@@ -0,0 +1,130 @@
...
+            temp = self._transfer(b, in_facts, node)
+            if temp != out_facts[b]:
+                out_facts[b] = temp
+                worklist.extend([pred.source for pred in b.predecessors])
+
+    def _transfer(self, block: CFGBlock, out_facts: Set[str], scope: Union[astroid.Module, astroid.FunctionDef]) -> Set[str]:
+        gen = out_facts.copy()
+        kill = set()
+        for statement in reversed(block.statements):
+            if isinstance(statement, astroid.Expr) and isinstance(statement.value, astroid.Call) and \
```
david-yz-liu: I don't understand the purpose of this block, please explain.

python_ta/checkers/redundant_assignment_checker.py
```diff
@@ -0,0 +1,130 @@
...
+
+        Note that `local variable` in the context of a module level analysis,
+        refers to global variables.
+        """
+        assigns = set()
+        kills = set()
+        for name, nodes in node.scope().locals.items():
+            if any(isinstance(elem, astroid.AssignName) for elem in nodes):
+                assigns.add(name)
+        for statement in node.nodes_of_class((astroid.Nonlocal, astroid.Global), astroid.FunctionDef):
```
david-yz-liu: I don't see why this part is necessary. If you're doing a loop through just `node.scope().locals`, doesn't that only include local variables anyway?

kavins14: Screen Shot 2019-11-05 at 5 52 58 PM

The `locals` attribute is not entirely accurate, that's why I have to do an extra check.

python_ta/checkers/redundant_assignment_checker.py
```diff
@@ -83,12 +82,13 @@ def _transfer(self, block: CFGBlock, out_facts: Set[str], scope: Union[astroid.M
         gen = out_facts.copy()
         kill = set()
         for statement in reversed(block.statements):
+            if isinstance(statement, astroid.FunctionDef):
```
kavins14: This check is added because the `nodes_of_class` method in the `for` loop below will look into function definitions for the required nodes, which is not what it wanted. The new test case passes because of this specific check.

python_ta/checkers/redundant_assignment_checker.py
kavins14: This block specifically addresses function calls of locally defined functions. We're interested in them because they could affect the redundancy of an assignment statement (think `nonlocal` variables). The logic is that if a call to a locally defined function call is made and the local variables of that function are not a subset of `gen`, then we `kill` all the `gen` variables. This is because there is a possibility of usage of one of the `gen` variables within that function definition.

kavins14: Now that I'm thinking about it, the `locals` attribute is not entirely accurate. Should I fix the check? or should remove this entirely so that `gen` variables would be `kill`ed if a call to a locally defined function is made.

david-yz-liu: We've previously discussed locally-defined functions. The bottom line is that I am not interested in tracing behaviour between function calls, at least for now. Please remove this section.

python_ta/checkers/redundant_assignment_checker.py
david-yz-liu: Ah yes, we discussed this with the last checker. Given what I said above about not checking into function calls, though, you can remove this check here an include `global` and `nonlocal` variables in the search for redundant assignments.

The difference between this and the previous checker is that the only _false positives_ that would appear is if you have two assignments to the same global/nonlocal variable, but in between there's a function call that uses this variable. I don't care about checking for this situation right now.

python_ta/checkers/redundant_assignment_checker.py
```diff
@@ -0,0 +1,132 @@
...
+        self._analyze(node)
+
+    def _analyze(self, node: Union[astroid.Module, astroid.FunctionDef]) -> None:
+        """Runs the backward data flow algorithm on a `Module` or `Function` CFG, which in turn
+        appends `Assign` nodes to `self.redundant_assignment` if it is redundant.
+
+        Data flow algorithms retrieved from:
+        https://www.seas.harvard.edu/courses/cs252/2011sp/slides/Lec02-Dataflow.pdf#page=31
+        """
+        # stores all the variables that will be re-defined before any usage at a
```
david-yz-liu: Capitalize "stores". Also your documentation here should specify the type better; you can say the "variable names" (which implies strings) rather than just "variables" (which might be a kind of astroid node).

python_ta/checkers/redundant_assignment_checker.py
david-yz-liu: Okay this is good. Please add a comment in the code.

python_ta/checkers/redundant_assignment_checker.py
```diff
@@ -0,0 +1,126 @@
...
+    def _transfer(self, block: CFGBlock, out_facts: Set[str]) -> Set[str]:
+        gen = out_facts.copy()
+        kill = set()
+        for statement in reversed(block.statements):
+            if isinstance(statement, astroid.FunctionDef):
+                # `nodes_of_class` below does block looking for required nodes
+                # in function definitions, hence this case.
+                continue
+            for node in statement.nodes_of_class((astroid.AssignName, astroid.DelName, astroid.Name,
+                                                  astroid.Nonlocal, astroid.Global),
```
david-yz-liu: Okay it's fine to include these two node classes nodes here, but they need a separate case to remove names from `kill`. `Global` and `NonLocal` nodes don't have a `name` attribute, they have a `names` attribute which is a `List[str]`.

PR #610, opened 2019-11-08
Migration to PyTest

kavins14:
Most of the test suites currently use either the `nosetests` or `unittests` library, both of which share common features with `pytest`, except that they aren't as maintained as `pytest`.

This PR proposes to change all test suites that use `nosetests` or `unittests` to `pytest` where applicable.

Todo: Update `Travis CI` configuration


setup.py
```diff
@@ -28,7 +28,6 @@ def readme():
         'hypothesis',
         'pycodestyle',
         'pylint>=2.1,<2.2',
-        'nose',
```
david-yz-liu: This is good, but you need to add `pytest` to replace it. I *think* it's possible to add this under `extras_require['dev']`, but try it out.

tests/test_setendings.py
```diff
@@ -486,4 +486,5 @@ def test_yieldfrom(self):
 
 
 if __name__ == '__main__':
-    unittest.main()  # run tests
+    unittest.main()     # run tests
```
david-yz-liu: Revert these changes (they aren't related to the PR).

tests/test_type_inference/utils.py
```diff
@@ -0,0 +1,5 @@
+def eq_(a, b, msg=None):
```
david-yz-liu: You have the right instinct for preserving this functionality, but this isn't necessary for our case. It turns out `pytest` itself provides good error messages when assertions fail, and we should rely on that.

So please remove this, and in the test files above, replace `eq_(a, b)` with `assert a == b` and `eq_(a, b, msg)` with `assert a == b, msg`.

david-yz-liu:
@kavins14 looks like the tests fail because of an import error. Not sure why this wasn't happening before, but you should be able to fix this using [relative imports](https://realpython.com/absolute-vs-relative-python-imports/#relative-imports).

PR #611, opened 2019-11-08
Bug: Failing Tests

kavins14:
None of the tests in these test suites was passing because of the incorrect path.


tests/test_setendings.py
```diff
@@ -4,8 +4,11 @@
 """
 import unittest
 from python_ta.transforms.setendings import *
+import os.path as path
 
-PATH = 'examples/ending_locations/'
+# Gives absolute path to root directory
+REPO_PATH = path.normpath(path.join(path.abspath(__file__), '../..'))
+PATH = REPO_PATH + '/examples/ending_locations/'
```
david-yz-liu: @kavins14 please merge this into one big `path.join` expression with the previous line. Also don't use the separator `/` explicitly, instead pass in each segment separately to `join`, e.g. `path.join(path1, '..', '..', 'examples', 'ending_locations')`

kavins14: @david-yz-liu I have made the change you requested not only for this specific case but also everywhere we're concatenating multiple paths. Though this is not what you explicitly asked for, I feel It's in line with what we're trying to achieve in this PR, which is to make the code more robust.

david-yz-liu: @kavins14 this is a great point. But since you want to change all these anyway, I think the best approach is actually to move the path joining into the helper function `get_file_as_module`, since that function is specific to this file and use case anyway.

PR #622, opened 2020-01-12
Removes duplicated messages for config file loading

ihasan98:
Replaces that message with a message that indicated what file was linted with what config.
This new message is printed after the linting is done.

Before:
Screen Shot 2020-01-12 at 1 43 27 PM

After:
Screen Shot 2020-01-12 at 1 44 01 PM


david-yz-liu:
Thanks @ihasan98!

PR #627, opened 2020-02-02
Allows users to use their own functions in their contracts

ihasan98:
Fixes the ability for users to add their own functions in function contracts.

Adds test cases too


david-yz-liu:
Thanks @ihasan98!

PR #628, opened 2020-02-08
Adds class contracts and representation invariants

ihasan98:
This sets up the basic structure for class contracts and adds tests. A sample use case is contained in the testing file that's added.

### DIFF
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -4,10 +4,9 @@ import inspect
 import wrapt
 
 
-class InvalidCondition(Exception):
-    """Exception raised when a condition is invalid
+class InvalidAssertion(Exception):
+    """Exception raised when a Assertion is invalid
     """
-
 
 @wrapt.decorator
 def check_contracts(wrapped, instance, args, kwargs):
@@ -20,6 +19,18 @@ def add_class_contracts(wrapped, args, kwargs):
     rep_invariants = parse_conditions(
         wrapped.__doc__ or '', parse_token="Representation Invariant")
     setattr(wrapped, "__representation_invariants__", rep_invariants)
+
+    def new_setattr(self, name, value):
+        super_class = wrapped.__mro__[1]
+        super_class.__setattr__(self, name, value)
+        curframe = inspect.currentframe()
+        callframe = inspect.getouterframes(curframe, 2)
+        if callframe[1][3] not in wrapped.__dict__:
+            # Only validating if the attribute is not being set in a instance/class method
+            validate_invariants(wrapped, self)
+
+    wrapped.__setattr__ = new_setattr
+
     return wrapped(*args, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


python_ta/contracts/__init__.py
```diff
@@ -1,40 +1,70 @@
 from typing import Any, List, Optional
 import typing
+import inspect
 import wrapt
 
 
-class InvalidPrecondition(Exception):
-    """Exception raised when precondition is invalid
+class InvalidCondition(Exception):
```
david-yz-liu: It's a good idea to make this name more general now; let's call this `InvalidAssertion` instead, as an indicator that the strings are fundamentally wrapping `assert`s.

python_ta/contracts/__init__.py
```diff
@@ -1,40 +1,70 @@
...
+    """Exception raised when a condition is invalid
     """
 
+
 @wrapt.decorator
-def check_contracts(wrapped, _instance, args, kwargs):
+def check_contracts(wrapped, instance, args, kwargs):
+    if instance is None and inspect.isclass(wrapped):
+        return add_class_contracts(wrapped, args, kwargs)
+    return check_function_contracts(wrapped, instance, args, kwargs)
```
david-yz-liu: This is okay, but please use an `if`-`else` structure when the branches are roughly symmetric.

tests/test_class_contracts.py
```diff
@@ -0,0 +1,68 @@
...
+    assert 'len(self.name) > 0' in msg
+
+def test_change_age_invalid_under() -> None:
+    """
+    Change the age to a negative number. Expect an exception.
+    """
+    with pytest.raises(AssertionError) as excinfo:
+        p1.change_age(-10)
+    msg = str(excinfo.value)
+    assert 'self.age > 0' in msg
```
david-yz-liu: Please end the file with a newline.

python_ta/contracts/__init__.py
david-yz-liu: Similarly below, when you've used "condition" let's change the generic name to "assertion", e.g. `parse_assertions` as the function name.

python_ta/contracts/__init__.py
```diff
@@ -1,40 +1,70 @@
...
+                            type(instance).__representation_invariants__)
+
     if 'return' in annotations:
         return_type = annotations['return']
         assert check_type_annotation(return_type, r),\
             f'Return value {r} does not match annotated return type {return_type.__name__}'
     return r
 
+
+def validate_conditions(wrapped, function_locals, conditions):
```
david-yz-liu: In addition to swapping out "conditions" for "assertions", I worry that the use of "validate" here makes it sound like this function checks whether the assertions are formed properly, rather than whether they're actually satisfied by the given `function_locals`.

Let's stick with "check", i.e., `def check_assertions` as the header here. This is a generic enough name that it can still be used for both preconditions and representation invariants.

python_ta/contracts/__init__.py
```diff
@@ -1,40 +1,70 @@
...
+    for condition in conditions:
+        try:
+            check = eval(condition, wrapped.__globals__, function_locals)
+        except:
+            # TODO: Decide what to do here, e.g. "Invalid condition"
+            raise InvalidCondition(
+                f'Error evaluating condition: {condition}')
+        else:
+            assert check,\
+                f'Condition "{condition}" violated for arguments {function_locals}'
```
david-yz-liu: Okay this is interesting. I'd like to have separate error messages for the violation of a precondition vs. representation invariants. (Indeed, `function_locals` is too broad when it's only the `self` argument being checked for a representation invariant.)

I'll let you decide how you want to handle this: either with another parameter, or using a `try`-`except` to handle the `AssertionError` in context, etc.

python_ta/contracts/__init__.py
```diff
@@ -55,35 +85,36 @@ def check_type_annotation(annotation: Optional[type], v: Any) -> bool:
             return True
 
 
-def parse_preconditions(docstring: str) -> List[str]:
-    """Return a list of preconditions parsed from the given docstring.
+def parse_conditions(docstring: str, parse_token: str = "Precondition") -> List[str]:
```
david-yz-liu: Reminder: "conditions" -> "assertions", and remember to change the local variables tool.

python_ta/contracts/__init__.py
ihasan98: Done

python_ta/contracts/__init__.py
ihasan98: Done

python_ta/contracts/__init__.py
ihasan98: That makes sense. I've changed it.

python_ta/contracts/__init__.py
ihasan98: I now have a separate function to check invariants hence there is a different error message.

david-yz-liu:
Awesome, great work @ihasan98!

PR #632, opened 2020-03-16
Auto-decorates class methods, and ensures method name collisions don't…

ihasan98:


python_ta/contracts/__init__.py
```diff
@@ -27,10 +27,19 @@ def new_setattr(self, name, value):
         super_class.__setattr__(self, name, value)
         curframe = inspect.currentframe()
         callframe = inspect.getouterframes(curframe, 2)
-        if callframe[1][3] not in wrapped.__dict__:
+        frame_members = dict((name, member) for name,
+                             member in inspect.getmembers(callframe[1].frame))
+        frame_locals = frame_members['f_locals']
```
david-yz-liu: If you're just accessing `f_locals` here, can't you just do `callframe[1].frame.f_locals`? Seems wasteful to build the dictionary.

python_ta/contracts/__init__.py
```diff
@@ -27,10 +27,19 @@ def new_setattr(self, name, value):
...
+        if 'self' in frame_locals and self is frame_locals['self']:
```
david-yz-liu: (assuming `.f_locals` yields a dictionary) You can use the `dict.get` method, which returns `None` when the key isn't in the dictionary.

But also since we're at the bottom of the function here, you can negate this condition and remove the `return` branch, doing something like

```python
if ...:
    check_invariants(self)
```

python_ta/contracts/__init__.py
ihasan98: As far as I know, `callframe[1].frame` is a list of tuples with the first value being the name, and second being the value. I wanted to convert it into a dictionary for readability purposes rather than iterating over it and trying to find the one with name 'f_locals'

python_ta/contracts/__init__.py
ihasan98: Done

PR #634, opened 2020-03-31
Adds Subclasses and Class Scope

ihasan98:
The following have been added:
- The scope in which the class is defined is now accessible while evaluating representation invariants.
- Basic support for sub classes. This will continue being tested and improved upon.

### DIFF
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -30,9 +30,7 @@ def add_class_contracts(wrapped, args, kwargs):
         frame_members = dict((name, member) for name,
                              member in inspect.getmembers(callframe[1].frame))
         frame_locals = frame_members['f_locals']
-        if 'self' in frame_locals and self is frame_locals['self']:
-            return
-        else:
+        if self is not frame_locals.get('self'):
             # Only validating if the attribute is not being set in a instance/class method
             check_invariants(self)
 
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -32,9 +32,13 @@ def instance_method_wrapper(wrapped, rep_invariants=None):
 
 
 def add_class_contracts(wrapped, args, kwargs):
+    if "__representation_invariants__" in wrapped.__dict__:
+        # This means it has already been decorated
+        return wrapped(*args, **kwargs)
+
     rep_invariants = []
 
-    # Iterating over all inhertied classes except Object
+    # Iterating over all inherited classes except Object
     for cls in wrapped.__mro__[:-1]:
         rep_invariants.extend(parse_assertions(
             cls.__doc__ or '', parse_token="Representation Invariant"))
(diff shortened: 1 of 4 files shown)


python_ta/contracts/__init__.py
```diff
@@ -17,13 +17,32 @@ def check_contracts(wrapped, instance, args, kwargs):
...
+
+    return wrapper(wrapped)
+
+
 def add_class_contracts(wrapped, args, kwargs):
-    rep_invariants = parse_assertions(
-        wrapped.__doc__ or '', parse_token="Representation Invariant")
+    rep_invariants = []
+
+    # Iterating over all inhertied classes except Object
```
david-yz-liu, line 37: There's a typo in "inherited". Also don't capitalize `object` in Python.

python_ta/contracts/__init__.py
```diff
@@ -17,13 +17,32 @@ def check_contracts(wrapped, instance, args, kwargs):
...
+    for cls in wrapped.__mro__[:-1]:
+        rep_invariants.extend(parse_assertions(
```
david-yz-liu, line 39: This is a natural place to check whether `cls` already has a `__representation_invariants__` attribute.

python_ta/contracts/__init__.py
```diff
@@ -17,13 +17,32 @@ def check_contracts(wrapped, instance, args, kwargs):
...
+            cls.__doc__ or '', parse_token="Representation Invariant"))
+
     setattr(wrapped, "__representation_invariants__", rep_invariants)
```
david-yz-liu, line 42: I didn't catch this before, but it seems this should be `setattr(type(wrapped), ...)`. Similar to what we discussed in yesterday's meeting.

python_ta/contracts/__init__.py
```diff
@@ -32,7 +51,13 @@ def new_setattr(self, name, value):
         frame_locals = frame_members['f_locals']
         if self is not frame_locals.get('self'):
             # Only validating if the attribute is not being set in a instance/class method
-            check_invariants(self)
+            init = getattr(wrapped, "__init__")
+            check_invariants(self, rep_invariants, init.__globals__)
+
+    for attr in wrapped.__dict__:
```
david-yz-liu, line 57: Right, so in the same vein, it should be possible to wrap the base methods in `type(wrapped).__dict__` rather than `wrapped.__dict__`.

tests/test_class_contracts.py
```diff
@@ -67,6 +72,16 @@ def test_change_name_invalid() -> None:
     msg = str(excinfo.value)
     assert 'len(self.name) > 0' in msg
 
+def test_change_name_invalid_nonalpha() -> None:
+    """
+    Change name to contain non-alphabet characters. Expect an exception.
+    """
+    with pytest.raises(AssertionError) as excinfo:
+        p1.change_name("David")  # Changing it back to a valid value.
```
david-yz-liu, line 80: You might have added this line because the test was failing? It would be great if you explored [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html#fixtures-as-function-arguments) as a way to remove the global variable in your tests (same as the new tests down below, too).

ihasan98: I'll do this for my next PR in which I focus on rewriting tests and adding more complex ones

python_ta/contracts/__init__.py
ihasan98: Unless I'm mistaken, wrapped is already the class not an instance of the class. I've pushed a new change that should hopefully prevent classes from being decorated twice.

python_ta/contracts/__init__.py
david-yz-liu: @ihasan98 please add a `__dict__` check here as well

ihasan98: Changed it to sets

PR #639, opened 2020-05-22
Class Type Hints and Decorating Modules

ihasan98:


python_ta/contracts/__init__.py
```diff
@@ -5,18 +5,28 @@
 import wrapt
 
 
-def check_all_contracts(module_name: str = '__main__') -> None:
+def check_all_contracts(*args, **kwargs) -> None:
```
david-yz-liu, line 8: You can use a keyword-only argument (https://www.python.org/dev/peps/pep-3102/) instead of `**kwargs`. This is a tighter specification for this function.

python_ta/contracts/__init__.py
```diff
@@ -114,13 +130,25 @@ def wrapper(wrapped, instance, args, kwargs):
...
             _check_invariants(instance, rep_invariants, init.__globals__)
+            _check_class_type_annotations(instance)
         except AssertionError as e:
             raise AssertionError(str(e)) from None
         else:
             return r
 
     return wrapper(wrapped)
 
+def _check_class_type_annotations(instance):
```
david-yz-liu, line 141: You'll notice I put in type annotations everywhere. You can use `Any` for `instance` and `None` for the return type.

python_ta/contracts/__init__.py
```diff
@@ -114,13 +130,25 @@ def wrapper(wrapped, instance, args, kwargs):
...
+    """Checks that the type annotations for the class still hold
```
david-yz-liu, line 142: Use "Check" instead of "Checks" and end the sentence with a period.

PR #642, opened 2020-08-12
pylint errors

adamgleizer:


examples/pylint/R0916_too_many_boolean_expressions.py
```diff
@@ -0,0 +1,6 @@
+def numberChoice(x: int, y: int, z: int) -> None:
```
david-yz-liu, line 1: Use `number_choice` here (and in general convert "CamelCase" to "snake_case" for function/variable names.

examples/pylint/W0108_unnecessary_lambda.py
```diff
@@ -0,0 +1,10 @@
+from typing import List
+
+def add(lst: List[int]) -> int:
```
david-yz-liu, line 3: This is the right idea but a bit complicated. Make this a simpler function, like `add_one`.

examples/pylint/W0108_unnecessary_lambda.py
```diff
@@ -0,0 +1,10 @@
...
+    """Calculate the sum of the elements in the given list."""
+    temp = 0
+    for item in lst:
+        temp += item
+    return temp
+
+print((lambda x: add(x))([1,2,3]))  # Error on this line
```
david-yz-liu, line 10: Remove the `print` call, so that you only keep the argument expression (which has the error).

examples/pylint/W0127_self_assigning_variable.py
```diff
@@ -0,0 +1,5 @@
+i = 0
+while i < 10:
```
david-yz-liu, line 2: You don't need a while loop here. Even something like `x = 1`, followed by `x = x` should cause this error.

examples/pylint/W0410_misplaced_future.py
```diff
@@ -0,0 +1,3 @@
+import sys
+
+from __future__ import print_function #  Error on this line
```
david-yz-liu, line 3: There should be two spaces before the `#` and one space after it.

examples/pylint/W0511_fixme.py
```diff
@@ -0,0 +1,3 @@
+def fishyFunction() -> None:
```
david-yz-liu, line 1: Make sure this is snake case (and there are other places too). You could keep the `FIXME` but follow it with a comment like "implement this function", and remove the `print` call.

examples/pylint/W0706_try_except_raise.py
```diff
@@ -0,0 +1,4 @@
+try:
+    1 / 0
+except:
```
david-yz-liu, line 3: Catch the `ZeroDivisionError` here

examples/pylint/W0715_raising_format_tuple.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(self, message: str, numerator: int, denominator : int):
+        self.message = message
+        self.numerator = numerator
+        self.wasZeroDivide = (denominator == 0)
+        super().__init__()
+
+
+def divide(numerator: int, denominator: int) -> float:
+    """Divide the numerator by the denominator."""
+    try:
```
david-yz-liu, line 11: This is a good idea, but you don't need a `try` here. Check if the `denominator` is 0, and if so do a `raise DivisionException(...)` the way that you have it below.

(`except` requires an exception class, not an exception instance, so the current example is misleading)

examples/pylint/W0716_wrong_exception_operation.py
```diff
@@ -0,0 +1,4 @@
+try:
+    1 / 0
+except ZeroDivisionError >= Exception:  # Error on this line
```
david-yz-liu, line 3: for this one I would probably try `ZeroDivisionError or Exception`

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                      'something that isn\'t a str, list, or tuple.',
+             'W0101': 'Code after a return or raise statement will '
+                      'never be run, so either the function is '
+                      'returning early or you should remove it.',
+             'W0102': 'Dangerous default value %s as argument. '
+                      'Using mutable values as a default argument '
+                      'is a common cause of confusing and unwanted behaviour.',
+             'W0104': 'This statement doesn\'t have any effect, and could be '
+                      'removed without changing the behaviour of the program.',
+             'W0106': 'Expression "%s" isn\'t assigned to a variable.',
```
david-yz-liu, line 19: Say "Right now, this expression isn't being used by the rest of your code. You should either remove this expression or save it in a variable to use it later."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+             'W0108': 'The lambda function isn\'t necessary since the lambda '
```
david-yz-liu, line 20: "...since you can pass the arguments directly to the function you're calling in the lambda's body."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+             'W0122': 'Using `exec()` to evaluate a code-block represented as a '
+                      'str can be potentially dangerous, and is typically a sign of poor design.',
+             'W0123': 'Using `eval()` to evaluate an expression represented as a '
+                      'str can be potentially dangerous, and is typically a sign of poor design.',
+             'W0199': 'Calling assert on a tuple will only evaluate to false if '
+                      'it is empty. It\'s likely you intended \'assert x, y\' rather '
+                      'than \'assert (x,y)\'.'
+             },
+        'ComparisonChecker':
+            {'R0123': 'tHE \'is\' keyword compares two objects location in memory rather '
```
david-yz-liu, line 34: Replace first sentence with: "The 'is' operator compares the ids of its two arguments, not their values."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                      'than value equality. When performing comparison with a literal, '
+                      'this is usually not what you want. Did you mean \'==\'?',
+             'C0123': 'You should use \'isinstance\' rather than \'type\' to '
+                      'perform a typecheck.'
```
david-yz-liu, line 38: "perform a type check" -> "check the type of a value"

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                      'Please refer to the PyTA Help Documentation for more information '
+                      'on Python Naming Conventions.'
+             },
         'PassChecker':
-             {'W0107': 'Unnecessary pass statement (you should remove this)'}
+            {'W0107': 'Unnecessary pass statement (you should remove this). '
+                      'A pass statement should only be used to represent '
+                      'an \"empty\" code block, where nothing happens.'},
+        'BasicErrorChecker':
+            {'E0101': '__init__ is called by the special method __new__ '
```
david-yz-liu, line 52: This is too technical. Replace with "__init__ is called only to initialize the attributes of a new object. It should not return anything."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                      'when a new object is instantiated, and will raise '
+                      'a TypeError if __init__ returns anything other than None.',
+             'E0102': '%s already defined line %s. '
```
david-yz-liu, line 55: "was already defined on line"

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                      'You should make sure all functions, methods and classes '
+                      'defined have different names.',
+             'E0103': 'The %r keyword should only be used inside a loop.',
+             'E0104': 'The \'return\' keyword is used to end the execution of a function '
+                      'and return a result, so you should only use it inside of a function. '
+                      'Note that if no return statement is written, Python implicitly '
```
david-yz-liu, line 61: This second sentence is unrelated to the error, so it should be removed.

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
     },
     'pylint.checkers.refactoring': {
         'RefactoringChecker':
-             {'R1710': 'Since this function can return a non-None value, you should '
-                       'explicitly write "return None" whenever None is returned by '
-                       'this function. (Possibly including possibly at the end of the function body.)'
+            {'R1710': 'Since this function can return a non-None value, you should '
+                      'explicitly write "return None" whenever None is returned by '
+                      'this function. (Possibly including at the end of the function body.)'
              }
```
david-yz-liu, line 71: This `}` should align with the opening `{`

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
     },
     'pylint.checkers.design_analysis': {
-        'MisdesignChecker': {'R0913': 'Too many parameters (%s/%s)'}
+        'MisdesignChecker': {'R0912': 'The function has too many branches (%s out of a default limit '
+                                      'of %s). You should try to reduce the complexity of the function '
+                                      'by splitting it up.',
```
david-yz-liu, line 76: up -> up into helper fucntions.

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
-                       'explicitly write "return None" whenever None is returned by '
-                       'this function. (Possibly including possibly at the end of the function body.)'
+            {'R1710': 'Since this function can return a non-None value, you should '
+                      'explicitly write "return None" whenever None is returned by '
+                      'this function. (Possibly including at the end of the function body.)'
              }
     },
     'pylint.checkers.design_analysis': {
-        'MisdesignChecker': {'R0913': 'Too many parameters (%s/%s)'}
+        'MisdesignChecker': {'R0912': 'The function has too many branches (%s out of a default limit '
```
david-yz-liu, line 74: Remove the word "default" (since these are configurable), and rephrase to "(%s, exceeding limit %s)". Applies for similar messages below.

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                                      'of %s). You should try to reduce the complexity of the function '
+                                      'by splitting it up.',
+                             'R0913': 'The function has too many parameters (%s out of a default limit '
+                                      'of %s). You should try to reduce the complexity of the function '
+                                      'by splitting up it, or combining related objects as a single one.',
+                             'R0914': 'The function has too many local variables (%s out of a default limit '
+                                      'of /%s).',
```
david-yz-liu, line 81: Include the "You should try to reduce the complexity..." sentence here too.

python_ta/patches/error_messages.py
david-yz-liu: Also, use "This" instead of "the" (throughout this section's messages)

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                             'R0915': 'The function or method has too many statements (%s out of a default limit '
```
david-yz-liu, line 82: You don't need to say "or method" here or in the next sentence.

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                                      'of /%s). It should be split up into smaller functions or methods.',
+                             'R0916': 'The if statement contains too many boolean expressions '
+                                      '(%s out of a default limit of %s). You should see if you can reduce '
+                                      'the amount of expressions, or split it into branches.',
+                             }
+    },
+    'pylint.checkers.classes': {
+        'ClassChecker': {'E0203': 'You can\'t use class member %r prior to it\'s assignment '
```
david-yz-liu, line 90: class member -> instance attribute

it's -> its

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                                  'on line %s.',
+                         'E0211': 'Every class method requires at least one parameter, '
```
david-yz-liu, line 92: class method -> instance method

parameter -> parameter, since the first parameter refers to the object on which this method is called. Python's convention is to name this parameter 'self'.

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                                  'which we conventionally name \'self\'.',
+                         'E0213': 'The first parameter of a method should always be called \'self\'. '
+                                  'This is a strongly adhered to convention.',
```
david-yz-liu, line 95: Replace with "Even though this is technically required, it is a strong convention used by all Python programmers."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                         'E0241': 'The class %r should not inherit from the same class multiple times.',
+                         'R0201': 'Method does not make use of \'self\', so you should rewrite the '
+                                  'method as a function and move it outside the class.',
+                         'R0205': 'Class %r inherits from \'object\', which can be omitted in Python3.',
```
david-yz-liu, line 99: No need to mention Python 3 (PyTA only supports Python 3).

"All classes inherit from `object` by default, so you do not need to specify this in the class header."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+    },
+    'pylint.checkers.classes': {
+        'ClassChecker': {'E0203': 'You can\'t use class member %r prior to it\'s assignment '
+                                  'on line %s.',
+                         'E0211': 'Every class method requires at least one parameter, '
+                                  'which we conventionally name \'self\'.',
+                         'E0213': 'The first parameter of a method should always be called \'self\'. '
+                                  'This is a strongly adhered to convention.',
+                         'E0241': 'The class %r should not inherit from the same class multiple times.',
+                         'R0201': 'Method does not make use of \'self\', so you should rewrite the '
```
david-yz-liu, line 97: "so you should remove the self parameter and move this method outside of the class (turning it into a top-level function)."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                                  'method as a function and move it outside the class.',
+                         'R0205': 'Class %r inherits from \'object\', which can be omitted in Python3.',
+                         'W0211': 'Static method %r shouldn\'t contain \'self\' as the first parameter, '
```
david-yz-liu, line 100: method -> methods

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                                  'since this suggests we are expecting a class instance as the first argument.',
+                         'W0212': 'Since %s starts with an underscore, it should be considered '
+                                  '"private", and not accessed outside of the class in which it is '
+                                  'defined.',
+                         'W0221': 'Method must take the same number of arguments as %s %r method.',
+                         'W0222': 'The method parameters must have the same name, order, and default '
+                                  'arguments as %s %r method.',
+                         'W0233': 'The abstract method %r in class %r must be overridden within a '
+                                  'concrete class.',
```
david-yz-liu, line 109: class -> subclass

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                         'W0231': 'Child class should call the \'__init__\' method from it\'s base class %r.',
```
david-yz-liu, line 110: child class -> subclass __init__ method

it's -> its

base class -> superclass

david-yz-liu: similar changes to W0233 on the next line

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                         'W0233': 'Child class should call the \'__init__\' method of the parent class '
+                                  'rather than some unrelated class.',
+                         },
+        'SpecialMethodsChecker': {
+            'E0302': 'The special method %r expects to take %s parameters, but %d %s given.',
+        },
+    },
+    'pylint.checkers.exceptions': {
+        'ExceptionsChecker': {
+            'E0701': '\'except\' blocks are analyzed top to bottom, so generic exception types should '
```
david-yz-liu, line 120: analyzed -> checked

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'come after specific exception types. Otherwise, the block for the specific '
+                     'exception type will never be reached. (%s)',
+            'E0702': '\'raise\' is used to raise Exceptions. Anything else will lead to an error. '
```
david-yz-liu, line 123: don't capitalize "Exception"

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     '(Raising %s)',
+            'E0704': '\'raise\' may only be called without an expression inside an \'except\' block.',
```
david-yz-liu, line 125: Don't refer to "calling" a raise.

Here, say "This raise statement must have an exception class or instance."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+            'E0710': '\'raise\' expects an object derived from the BaseException class. Attempting to '
```
david-yz-liu, line 126: "This raise statement is using a class that is not a subclass of BaseException."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'raise any other object will lead to an error.',
+            'E0711': '\'NotImplemented\' is strictly used as a return value for binary special methods '
+                     '(e.g __eq__), indicating that they\'re not implemented. You mean to raise '
```
david-yz-liu, line 129: Switch the order of the two sentences (so "You mean to raise..." comes first).

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     '\'NotImplementedError\', which indicates that the abstract method must be '
+                     'implemented by the derived class.',
+            'E0712': '\'except\' expects an object derived from the BaseException class. Attempting to '
```
david-yz-liu, line 132: object -> class

derived from -> subclass of

Remove the second sentence (it is unrelated)

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'raise any other object will lead to an error.',
+            'W0702': 'It\'s bad practice to use the \'except\' keyword without passing an expression, '
+                     'since any and all expressions will be caught, which may lead to bugs.',
```
david-yz-liu, line 135: expressions -> exceptions

which may lead to bugs -> "which may lead to undetected errors in your code"

david-yz-liu: similarly for W0703 below

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+            'W0703': 'It\'s bad practice to use \'except: Exception\' since any and all expressions '
+                     'will be caught, which may lead to bugs.',
+            'W0705': '\'except\' blocks are analyzed top to bottom, so if we try to catch the same '
```
david-yz-liu, line 138: analyzed -> checked

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'exception multiple times, only the first \'except\' block for the exception '
+                     'will be reached.',
+            'W0706': 'If the \'except\' block uses \'raise\' as it\'s first or only operator, it '
```
david-yz-liu, line 141: it's -> its

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'will just raise back the exception immediately.',
+            'W0711': '\'except\' can catch multiple exceptions, but they have to be passed in as a '
```
david-yz-liu, line 143: catch -> handle

passed in -> written

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'tuple. Your exception to catch is a result of a binary operation (%s), '
```
david-yz-liu, line 144: The second sentence gives too much information. Replace with an example: "For example, use except (A, B) instead of except A or B"

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'which will result in only one of the exceptions being caught by the  \'except\' block.',
+            'W0716': '%s is not a valid operation on exceptions.'
+        }
+    },
+    'pylint.checkers.format': {
+        'FormatChecker': {
+            'W0301': '\';\' in Python is used to put multiple statements on a single line, so this '
```
david-yz-liu, line 151: The first sentence provides unnecessary information. Simply say "You should remove this semicolon. In Python, statements are separated by starting each one on a new line."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'semicolon is unnecessary. In general, there is no good reason to ever use a \';\' '
+                     'in Python.',
+            'W0312': 'This code is indented with both %ss and %ss. You should stick to just one, '
```
david-yz-liu, line 154: Second sentence: "You should use spaces to indent your code."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'but note that spaces are preferred.',
+            'C0301': 'This line is %s characters long, exceeding our default limit of %s characters.',
```
david-yz-liu, line 156: use same wording as above "limit" error messages

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+            'C0303': 'Trailing whitespace is poor formatting, and should be removed.',
+            'C0304': 'It\'s proper practice to have the final line in your file be a newline, '
```
david-yz-liu, line 158: By convention, each Python file should end with a blank line. To fix this, go down to the bottom of your file and press Enter/Return to add an empty line.

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'which it is not.',
+            'C0321': 'Writing multiple statements on a single line is discouraged by PEP8.',
```
david-yz-liu, line 160: Don't refer to pep8. "Writing multiple statements on a single line makes your code harder to read and understand, and should be avoided."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+            'C0330': 'Wrong %s indentation%s%s.\n%s%s. This error occured because of an '
+                     'inconsistent use of number of spaces for indentation.'
+        }
+    },
+    'pylint.checkers.imports': {
+        'ImportsChecker': {
+            'E0401': 'Unable to import %s. Check the spelling of the module name, or '
+                     'whether the module is in the correct directory.',
```
david-yz-liu, line 168: "whether the module is installed (if it is a module you did not write), or whether the module is in the same directory as this file (if it is a module you did write)."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+            'W0401': 'Wildcard imports add all objects from the import module into the '
```
david-yz-liu, line 169: Wildcard -> Wildcard (*)

import -> imported

into the global namespace -> as global variables

may cause problems -> may result in name collisions

need -> need, or import the whole module ("import module") and use dot notation to access the module's functions/classes.

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'global namespace, which may cause problems. You should only '
+                     'import from %s what you need.',
+            'W0404': 'Module %r was reimported on line %s. A module should not be '
+                     'imported more than once.',
+            'W0410': 'Python 2.5 and greater requires \'__future__\' imports to be the '
```
david-yz-liu, line 174: Remove the reference to python version. "Imports of __future__ must be..."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+                     'first non-docstring statement in the module.',
+            'C0410': 'Different modules should not be imported on a single line. You '
+                     'should import each module on line %s on separate lines.',
+            'C0414': 'Aliasing an import as the name of the package is useless.',
```
david-yz-liu, line 178: "useless" is not helpful. Replace this message with "The 'import <module> as <name>" is used to import a module and give it a different name in this file. If you don't want to rename the module, simply use "import <module>" instead."

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+            'C0415': 'Import statements should be in the toplevel of the module. You '
```
david-yz-liu, line 179: "level" is actually not helpful here.

Replace first sentence with "Import statement should always be at the top of a module, or within an if __name__ == '__main__' block." Keep second sentence, that's good.

python_ta/patches/error_messages.py
```diff
@@ -4,18 +4,193 @@
...
+    },
+    'pylint.checkers.misc': {
+        'EncodingChecker': {
+            'W0511': 'The warning %s was found.'
+
+        }
+    },
+    'pylint.checkers.newstyle': {
+        'NewStyleConflictChecker': {
+            'E1003': 'The first argument to \'super\' should be the current class, '
```
david-yz-liu, line 191: Turn the comma into a period (comma splice).

examples/pylint/W0716_wrong_exception_operation.py
david-yz-liu: Hmm, looks like `or` causes a different exception to be changed. I will change this to `+`.

python_ta/patches/error_messages.py
david-yz-liu: This change was made on W0104 instead of W0106. I will fix this.

python_ta/patches/error_messages.py
david-yz-liu: (This is actually true throughout this file.)

Issue #657, opened 2020-11-11
PyTA thinks comprehension scope is shared with function scope


eamonma:
For example, the following code is perfectly valid. However, PyTA raises errors `E9969 (possibly-undefined)` and `E9988 (shadowing-inside-comprehension)`, even though the loop variable is in a different scope from the comprehension variable.

```
from typing import List


def example(example_list: List[int]) -> None:
    new_list = [same_name / 1 for same_name in example_list]

    for same_name in new_list:
        useless = same_name


if __name__ == '__main__':
    import python_ta

    python_ta.check_all(config={
        'max-line-length': 100,
        'extra-imports': ['dataclasses', 'python_ta.contracts'],
        'disable': ['C0116', "C0114"]
    })

    import python_ta.contracts

    python_ta.contracts.DEBUG_CONTRACTS = False
    python_ta.contracts.check_all_contracts()
```

kavins14:
Hi, my understanding is that error `E9988 (shadowing-inside-comprehension)` is there to prevent students from using the same variable name inside the comprehension as well as outside, as it can be considered as bad practice. This is in spite of the fact that the loop variable and comprehension variable are in different scopes.

I did some research as to why error E9969 (possibly-undefined) is raised, and this is what I found:
*  astroid does not consider the variable `same_name` inside the comprehension as a local variable, which is correct. And so the checker algorithm considers `same_name` as possible undefined because it thinks `same_name` refers to a local variable and not a variable in the comprehension scope. If `same_name` was not a local variable in the first place, this error would not be raised.
* One way to avoid this issue is to check if every variable `n` that is possible undefined is being used inside a comprehension that defines `n`. If so, we don't consider it possibly undefined.

I might create a PR within a couple of weeks to resolve this issue.

PR #674, opened 2021-05-07
Bug Fix: Possibly Undefined Checker

kavins14:
- Fixes issue #657 

Consider the following python program:
```python 
[x for x in lst] 
```
- It turns out that that the algorithm was taking into account the assignment of `x` from `for x in lst` AFTER taking into account the expression `x` making it seem as though `x` was not defined before being used. This is because the control flow graph does not take into consideration the order in which the sub-statements are evaluated. 
- A similar issue arises in the following statement `x = x + 1`. The algorithm would first consider `x` being assigned then `x + 1` being evaluated. It should actually be the other way around.

This PR fixes both these issues.


david-yz-liu (commented):
Hi @kavins14! Thanks for looking into this. It seems to me that you've correctly identified the (two!) problems with the current checker, but I think the solution could be cleaner. Right now the `get_nodes` function is really relying on the fact that it's only being called on simple statements to ensure the overall ordering of `Name`s/`AssignName`s is correct. (For example, if `statement` was a for loop containing multiple assignment statements, you can't extract all of the `Name` nodes and yield them first.)

Given this, I think it's better to be more explicit about how specific node classes require a different ordering than "nodes of class" provides. I would imagine something like:

```python
def get_nodes(self, statement: astroid.node_classes.NodeNG) -> Generator[astroid.node_classes.NodeNG, None, None]:
    if isinstance(statement, astroid.Assign):
        # Yield from calling nodes_of_class on RHS, then yield target(s)
    elif isinstance(statement, (astroid.ListComprehension, astroid.ListComprehension, ...):
        # Yield from calling nodes_of_class on comprehensions, then yield body
    else:
        # Yield from calling nodes_of_class on statement
```

It's less elegant to have to list out the comprehension expression types, but overall I think the logic will be more clear this way.

kavins14:
You are absolutely right @david-yz-liu, implementing the function that way seems to be the right way to do it.

python_ta/checkers/possibly_undefined_checker.py
```diff
@@ -119,6 +118,19 @@ def _get_assigns(self, node: Union[astroid.FunctionDef, astroid.Module]) -> Set[
 
         return assigns.difference(kills)
 
+    def get_nodes(self, statement: astroid.node_classes.NodeNG) -> Generator[astroid.node_classes.NodeNG, None, None]:
+        multiple_nodes = lambda nodes : chain.from_iterable(self.get_nodes(node) for node in nodes)
+        if isinstance(statement, astroid.Assign):
+            # RHS is evaluated before assigned in an assignment statement
+            yield from self.get_nodes(statement.value)
+            yield from multiple_nodes(statement.targets) # statement.targets is a list of nodes
+        elif isinstance(statement, (astroid.ListComp, astroid.SetComp)):
```
david-yz-liu, line 127: Please add `astroid.DictComp` and `astroid.GeneratorExp` here

tests/test_custom_checkers/test_possibly_undefined_checker.py
```diff
@@ -393,3 +393,42 @@ def func2():
...
+        print(list(func_node.nodes_of_class(astroid.Name)))
+
+        self.checker.visit_functiondef(func_node)
+        with self.assertAddsMessages(
+            pylint.testutils.Message(
+                msg_id='possibly-undefined',
+                node=name_node_x,
+            )
+        ):
+            self.checker.visit_name(name_node_x)
```
david-yz-liu, line 434: Please insert a blank line at the end of this file.

david-yz-liu (commented):
Thanks @kavins14! I left a few minor comments, but once you're done that I will merge this in!

PR #683, opened 2021-05-22
CLI entry point

kavins14:
This PR ports python_ta's existing public APIs to the command line. 

I decided to use the [click](https://pypi.org/project/click/) module to implement the command line interface since it's simple, powerful and it handles all the grunt work like cross-platform discrepancies (windows/mac os), options/arguments parsing and error handling.

To test it out:
```bash
$ pip install -e .[dev] # install the latest packages
$ virtualenv venv
$ . venv/bin/activate
$ pip install --editable
```

Afterwards, the command should be available:

```bash
$ python_ta -h
Usage: python_ta [OPTIONS] [FILENAMES]...

  A code checking tool for teaching Python.

  FILENAMES can be a string of a directory, or file to check (`.py` extension
  optional) or a list of strings of directories or files.

Options:
  -c, --config FILE  python_ta configuration file
  -E, --errors-only  Displays errors only
  -h, --help         Show this message and exit.
```

Suggestions:

* For the command line interface, I think it would be a good idea to use a reporter that prints to stdout and not generate a webpage.
* Currently the `doc` method redirects to a website, would it be overkill to have a man page or a subcommand that displays an error code's description?
* It feels a bit weird to have an underscore in a command `python_ta`. It can be awkward to type it out. What do you think of having the command be `pyta` instead? It might be confusing for end users but I'm not certain the extent.


david-yz-liu:
@kavins14 thanks, this is great! I will take a more thorough look later, but one thought is that it would be good to make this script the [`__main__.py`](https://docs.python.org/3/library/__main__.html) of the package. For example here is pylint's: https://github.com/PyCQA/pylint/blob/master/pylint/__main__.py.

david-yz-liu:
@kavins14 thanks! Sorry for the delay in looking at this. I merged this PR, but you have some interesting suggestions.

1. I could definitely see using `PlainReporter` by default for the CLI, though we'd need to be a bit careful in having that option merge with whatever configuration file is passed in.
2. I think there's definitely room for subcommands. I like `help` (easy to remember), so for example I could see `python_ta help E9999` to get information about that error. The messages are all compiled dynamically when pylint is loaded, so the "lookup" would require a bit of work, but not impossible.
3. This is a good point, but given that the module name (`python_ta`) isn't going to change, I want to keep the executable consistent with that.

PR #692, opened 2021-05-25
Error E1134 docs

lorena-b:
- Created example
- Added documentation to index.md


docs/checkers/index.md
```diff
@@ -312,6 +312,13 @@ for number in [1, 2, 3]:
     print(number)
 ```
 
+### Not a mapping (E1134) [](#E1134)
+
+This error occurs when a non-mapping value is used in a place where mapping is expected.
```
david-yz-liu, line 317: You can be more specific here---this check only occurs when using the `**` operation in a function call, so please describe that in more detail.

PR #696, opened 2021-05-26
Error R0401 docs

SimonChenWasTaken:
Add documentation and examples for Pylint error R0401 "Cyclic import". Ensured that example filenames will work well with test build.

In addition:
- Added error to ignored tests.
- Ignore examples that don't start with a code rather than assert false.


david-yz-liu (commented):
@SimonChenWasTaken looks like the test is still failing. You can change the test code itself so that it skips files under `examples/pylint` that don't start with a code (right now the code does an `assert False`, you can change that to simply return).

You'll need to look into why the test for `R0401_cyclic_import.py` is failing though. Make sure to check that you can actually see the error being reported.

Also, you don't need to close pull requests or change branches, just add new commits and push them to an existing branch.

PR #697, opened 2021-05-26
Error W0128 docs

SimonChenWasTaken:
Add error Pylint W0128 "Redeclared assigned name" documentation and examples.


david-yz-liu (commented):
@SimonChenWasTaken I don't think you're interepreting this error correctly. The Pylint documentation refers to this check happening when the same variable is assigned multliple times in the same statement, whereas your example uses two different statement. I'd try out something like `x, x = ...`.

PR #698, opened 2021-05-26
Error W1301 docs

lorena-b:
- Create example
- Add documentation to index.md


david-yz-liu:
@lorena-b whoops, you documented this error properly but I realized that we actually want to disable this check for Pylint. I'll update the Wiki accordingly.Sorry about the mixup!

PR #705, opened 2021-05-28
R1727 condition eval to const docs

ckatsube:
Added R1727 example file to `examples/pylint`

Added R1727 to `index.md` (Under `Code complexity` and `Unneeded not (C0113)` because it is a complexity-related boolean error). **Should be merged underneath R1726 when resolving the conflict**
Also specified that R1727 is only flagged in a test condition, not for just any boolean operation (made the additional line consistent with R1726). 

Added an example correcting R1727 in `index.md` (added an additional notation  `# Error was on this line` to explicitly show the difference)


ckatsube:
Resolved merge conflict with #704 (R1726)

PR #712, opened 2021-06-08
Resolving Issue #699 - Fix runtime error 

### LINKED ISSUE #699: unnecessary-indexing runtime error
The `UnnecessaryIndexingChecker` raises a runtime error on the following code:

```python
def f(lst: list) -> None:
    i = 0
    for i in range(len(lst)):
        if lst[0]:
            i = 1
        else:
            i = 2
```

lorena-b:
## Motivation and Context

This pull request is required to resolve #699. This change resolves a runtime error which occurs when running the `UnnecessaryIndexingChecker` on a block of code where an assignment node has an empty scope. 

This pull request also fixes functions that no longer worked correctly due to changes in the astroid node representation, which caused the example file cases to fail. 

## Your Changes

**Description**:

This change handles the case where the second argument of the tuple returned by `node.lookup()` is an empty tuple so that an index error does not occur. 

 `_is_load_subscript()` helper function was rewritten to produce the correct output. 

## Testing

In order to test this change a test suite was created for this checker using pylints testing capabilities. The checker was tested with the example code and the code that raised the original error. 

The tests for the code examples are failing because the checker seems to not have been implemented correctly. I am trying to figure out what is wrong with the current implementation that makes the checker not work even in basic cases. 

Update: 
- There was an issue with the `_is_load_subscript()` helper function due to changes in the astroid node representation. When indexing the iterable the Name node now directly has a Subscript as the parent, instead of having an Index node in between. 
- Checker now works to pass all of the example cases, issue #699 case, and some extra cases such as unused iteration variables and looping through the iterable directly. 

## Checklist


python_ta/checkers/unnecessary_indexing_checker.py
```diff
@@ -89,30 +89,32 @@ def _is_load_subscript(index_node: astroid.Name, for_node: astroid.For) -> bool:
...
-    # Name node is not inside Subscript node iterable[index].
-    if not (isinstance(index_node.parent, astroid.Index) and isinstance(index_node.parent.parent, astroid.Subscript)
-            and isinstance(index_node.parent.parent.value, astroid.Name)
-            and index_node.parent.parent.value.name == iterable):
-        return False
-    # Use ctx attribute to find out what context the subscript is being used in.
-    subscript_node = index_node.parent.parent
-    return subscript_node.ctx == astroid.Load
+    # If the Name node is inside a Subscript node and the subscript is being used in a load context
+    if isinstance(index_node.parent, astroid.Subscript) \
```
david-yz-liu, line 93: This is actually a case where there are four different conditions, and we should be returning the `and` of all 4. This means that you can simplify the code into something like:

```python
return (
    isinstance(index_node.parent, astroid.Subscript) and
    ...
)
```

Rater than use inline comments, please add to the function docstring to explain what these four conditions are.

python_ta/checkers/unnecessary_indexing_checker.py
```diff
@@ -89,30 +89,32 @@ def _is_load_subscript(index_node: astroid.Name, for_node: astroid.For) -> bool:
...
+        subscript_node = index_node.parent
+        return subscript_node.ctx == astroid.Load
+
+    # If Name node is not inside a Subscript node iterable[Name]
+    return False
 
 
-def _is_redundant(index_node: Union[astroid.AssignName, astroid.Name], for_node: astroid.For) -> bool:
+def _is_redundant(index_node: Union[astroid.AssignName, astroid.Name],
+                  for_node: astroid.For) -> bool:
```
david-yz-liu, line 105: Revert this change

python_ta/checkers/unnecessary_indexing_checker.py
```diff
@@ -89,30 +89,32 @@ def _is_load_subscript(index_node: astroid.Name, for_node: astroid.For) -> bool:
...
-    else:
-        return _scope_lookup(index_node) != for_node.target or _is_load_subscript(index_node, for_node)
+    if isinstance(index_node, astroid.Name):
+        return _scope_lookup(index_node) != for_node.target \
+               or _is_load_subscript(index_node, for_node)
 
 
-def _index_name_nodes(index: str, for_node: astroid.For) -> List[Union[astroid.AssignName, astroid.Name]]:
+def _index_name_nodes(index: str, for_node: astroid.For) -> List[
+    Union[astroid.AssignName, astroid.Name]]:
```
david-yz-liu, line 117: Revert this change

python_ta/checkers/unnecessary_indexing_checker.py
```diff
@@ -89,30 +89,32 @@ def _is_load_subscript(index_node: astroid.Name, for_node: astroid.For) -> bool:
...
 
-def _is_redundant(index_node: Union[astroid.AssignName, astroid.Name], for_node: astroid.For) -> bool:
+def _is_redundant(index_node: Union[astroid.AssignName, astroid.Name],
+                  for_node: astroid.For) -> bool:
     """Return whether or not <index_node> is redundant in <for_node>.
 
     The lookup method is used in case the original loop variable is shadowed
     in the for loop's body.
     """
-    if isinstance(index_node, astroid.AssignName):
```
david-yz-liu, line 108: It looks like you've removed this branch, but this loses existing behaviour, which is not good. Instead you should be attempting to fix the runtime error directly.

You should create test cases to explore what this code is doing. `AssignName` nodes can appear as the left-hand side of an `=`, as the target of a for loop, and as the target in a comprehension. Play around with each of these cases by filling in a code template that looks like

```python
for i in range(len(lst)):
    ...
```

david-yz-liu (commented):
@lorena-b great detective work. For the issue of the checker no longer working, you determined that the `Index` node is no longer available, and replaced that with accessing the `Subscript` directly.

However, this is a change in Python 3.9, so the tests are now failing in Python 3.8. In cases like this, we can check `sys.version` (imported from the `sys` module) to see what the current version of Python is, and then based on the result execute different branches of code. You should do this for the `is_load_subscript` function.

python_ta/checkers/unnecessary_indexing_checker.py
```diff
@@ -87,16 +88,38 @@ def _iterable_if_range(node: NodeNG) -> Optional[str]:
...
     iterable = _iterable_if_range(for_node.iter)
-    # Name node is not inside Subscript node iterable[index].
-    if not (isinstance(index_node.parent, astroid.Index) and isinstance(index_node.parent.parent, astroid.Subscript)
-            and isinstance(index_node.parent.parent.value, astroid.Name)
-            and index_node.parent.parent.value.name == iterable):
-        return False
-    # Use ctx attribute to find out what context the subscript is being used in.
-    subscript_node = index_node.parent.parent
-    return subscript_node.ctx == astroid.Load
+    version = sys.version[:3]
```
david-yz-liu, line 106: You're on the right track, but this is a less common way of checking versions, especially since we actually want to check for versions >= 3.9.

Use [`sys.version_info`](https://docs.python.org/3/library/sys.html#sys.version_info) and tuple comparison in the check below.

python_ta/checkers/unnecessary_indexing_checker.py
```diff
@@ -106,8 +129,12 @@ def _is_redundant(index_node: Union[astroid.AssignName, astroid.Name], for_node:
     in the for loop's body.
     """
     if isinstance(index_node, astroid.AssignName):
-        return index_node.lookup(index_node.name)[1][0] != for_node.target or \
-               not isinstance(index_node.parent, astroid.AugAssign)
+        if index_node.lookup(index_node.name)[1] != ():
```
david-yz-liu, line 132: This is good, but you need a matching `else` branch to ensure that a boolean is always returned.

python_ta/checkers/unnecessary_indexing_checker.py
```diff
@@ -106,8 +129,12 @@ def _is_redundant(index_node: Union[astroid.AssignName, astroid.Name], for_node:
     in the for loop's body.
     """
     if isinstance(index_node, astroid.AssignName):
```
david-yz-liu, line 131: To clarify my previous comment, you should add these new test cases to the test file you've created (I didn't see any new test cases in the recent commits you made).

python_ta/checkers/unnecessary_indexing_checker.py
```diff
@@ -106,8 +128,14 @@ def _is_redundant(index_node: Union[astroid.AssignName, astroid.Name], for_node:
...
-        return index_node.lookup(index_node.name)[1][0] != for_node.target or \
-               not isinstance(index_node.parent, astroid.AugAssign)
```
lorena-b, line 110: I'm still confused about this condition for AugAssign, because I don't really understand what the checker is supposed to flag that doesn't involve directly subscripting the iterable (based on the documentation). If I don't change this then some of the cases based off the example files (nested comprehensions (2, 3, 4), `test_loop_variable_reassigned_no_msg`, `test_comp_shadow_no_msg`) will fail. In the documentation for this checker it does say that it may not work properly for cases like these and raise a false error, so I'm wondering if these cases are supposed to be ignored?

david-yz-liu: @lorena-b I'm not sure exactly what you're asking. The tests you have added look good, and no cases should be ignored. In fact, if you have improved the checker, then you should update the documentation to remove this comment. It is okay to modify the code in this branch (after all, there was a bug), and my previous comment was that you should not remove the branch entirely.

However, there is a case that is suggested by the `AugAssign` that you haven't tested, and that's when we have a statement like `lst[i] += 1`. In this case, even though the `i` is inside a subscript node, it is not redundant. You should add a test for this case as well, and ensure that this implementation correctly handles it.

PR #713, opened 2021-06-11
Speeding up the `test_examples.py` test suite

ckatsube:
## Motivation and Context

The previous test suite called pylint in a subprocess for every file in `examples/pylint`, taking almost an entire second for the 161 example files. This PR aims to improve the running time of this test file by running pylint on all files simultaneously. Dividing up the output by the respective files and testing the precomputed outputs runs much more quickly than rerunning a subprocess per test. The current result is 2.5 seconds for the entire 161 tests, which most of the time being spent in the precomputation while the actual tests are carried out almost instantaneously.

I personally think this is a good change as having a test suite that takes a long time to run discourages actually running it. Having a fast test suite encourages collaborators to run it more often as well as add onto it without worrying about making it clunkier.

## Your Changes

**Description**: The previous subprocess call inside of the test was moved to an external function that takes in an entire list of file paths. The parametrized test was moved inside of a class to support a setup function that added the precomputed results to an easily accessible scope (an alternative was to use `global`, which I was not comfortable with). The additional functions are largely helpers for readability, not for increasing the functionality of the module.

## Testing

I've confirmed that the test cases pass (locally) when running `python -m pytest python_ta tests` in the CLI as well as running `pytest.main(['tests/test_examples.py'])` in the Python Console in the root directory. I momentarily deleted the contents of some files (not part of the PR) to check that the tests are not showing false positives. 

Possible downside: running pylint on all files at once denies the ability to add any additional checks on individual example files that may need exceptions (e.g. circular import). 
There is also one additional warning that I don't believe existed in the earlier test suite, but I'm unable to identify the cause of it. 
```python
<unknown>:9
  <unknown>:9: DeprecationWarning: invalid escape sequence \d
```
At first I thought this was the warning for the regex expression (which contains a \d on a line 9); however, that appears as its own warning so I'm once again unsure of what this warning expresses. 

## Checklist


ckatsube:
The comment about the additional warning is visible in the travis-ci Python 3.9 test. It appears that it really is flagging the `tests/test_example.py` regex expression inside of the string as both lines are now line 11 (because the addition of the type annotation imports).

tests/test_examples.py
```diff
@@ -34,39 +36,69 @@ def get_file_paths():
...
+    """Return a dict mapping example file path to its pylint output
     """
-    # The following are captured when this function is created.
-    def new_test_func():
+
+    pylint_output = {}
+
+    file_paths = get_file_paths()
+    full_output = _get_full_pylint_output(file_paths)
+    _, *individual_outputs = full_output.split("*************")
```
david-yz-liu, line 47: You can investigate how to use pylint's `json` reporter so that you get back JSON. This way you should be able to eliminate all string parsing, and simplify this function. You shouldn't need to define any helper functions.

tests/test_examples.py
```diff
@@ -34,39 +36,69 @@ def get_file_paths():
...
+        ['pylint', '--reports=n',
+         '--rcfile=python_ta/.pylintrc',
+         *file_paths],
+        stderr=subprocess.STDOUT,
+        stdout=subprocess.PIPE)
+
+    return output.stdout.decode('UTF-8')
+
+
+class TestExamples:
```
david-yz-liu, line 74: You don't need a class here. Below, you're using a pytest fixture, which is a good approach. You can just make the fixture return the dictionary to avoid storing it in a variable, and then pass the fixture to the actual test function.

You can read more about pytest fixtures here: https://docs.pytest.org/en/6.2.x/fixture.html.

tests/test_examples.py
```diff
@@ -34,39 +36,69 @@ def get_file_paths():
...
-            ['pylint', '--reports=n',
-             '--rcfile=python_ta/.pylintrc',
-             test_file],
-            stderr=subprocess.STDOUT,
-            stdout=subprocess.PIPE)
-        for line in output.stdout.decode('utf-8').split('\n'):
+        for line in output.split("\n"):
             if checker_name in line:
                 found_pylint_message = True
-                break
```
david-yz-liu, line 55: Revert this change

tests/test_examples.py
```diff
@@ -34,39 +36,69 @@ def get_file_paths():
...
         if not found_pylint_message:
             print('Failed: ' + test_file)  # test doesn't say which file
+
```
david-yz-liu, line 103: Revert this change

david-yz-liu (commented):
@ckatsube great initiative here, this is an exciting improvement to the test suite. I've left a few comments for you to work on.

ckatsube:
Changelog:
- Updated W0233's example file (it was missing a `-called` suffix) for exact-matching instead of substring matching. 
- Removed the Test class and changed the Pytest fixture to be returning.
  - The parametrized test now takes in the above fixture return as an argument.
- Changed the Pylint output type to `json` to decouple the divvying process from the \*s in the output message (which had previously been used to split up the messages).

Comments:
Regarding the remark about not needing helpers, I still ended up using three.
1. get the Pylint output and parse the json string to a Python `dict` object. 
2. strip the original Pylint json output as there appeared to be a random trailing ASNI code (`'\x1b[0m'`) depending on how you ran the Pytest (CLI did not have it while running it in the Python Console did). This one is arguable extra, but I imagine that the process of striping the ASNI code is better documented inside a helper rather than merged into the larger json-getting helper.
3. a helper to convert the dict-version of Pylint's json output to some more easily queryable by each test, as the raw output would require iterating over every individual raised message (including the ones not relevant to the test file being tested in that iteration).

tests/test_examples.py
```diff
@@ -34,39 +36,72 @@ def get_file_paths():
...
+         *get_file_paths()],
+        stderr=subprocess.STDOUT,
+        stdout=subprocess.PIPE)
+
+    jsons_output = output.stdout.decode('UTF-8')
+    jsons_output = _strip_trailing_asni(jsons_output)
+    return json.loads(jsons_output)
+
+
+def _strip_trailing_asni(str_to_strip: str) -> str:
```
david-yz-liu, line 53: I tried this out by running pytest both in the command line and in the Python Console without calling this function, and all of the tests passed. You can remove this function.

tests/test_examples.py
```diff
@@ -34,39 +36,72 @@ def get_file_paths():
...
+        symbols_by_file[file_name].add(symbol)
+
+    return symbols_by_file
+
+
+@pytest.fixture(scope='session', autouse=True)
+def symbols_by_file() -> Dict[str, Set[str]]:
+    """Run pylint on all the example files return the map of file name to the
+    set of Pylint messages it raises."""
+    pylint_list_output = _get_full_pylint_json_output()
```
david-yz-liu, line 87: In this particular case I find the function more readable if you write all of the code in this function rather than splitting it up.

Also you can simplify the grouping by using [`itertools.groupby`](https://docs.python.org/3/library/itertools.html#itertools.groupby) (but note that you'll still need to convert the returned value from this function into a `dict`).

tests/test_examples.py
```diff
@@ -34,39 +36,72 @@ def get_file_paths():
...
+    return _compile_message_symbols_per_file(pylint_list_output)
 
 
 @pytest.mark.parametrize("test_file", get_file_paths())
-def test_examples_files(test_file):
+def test_examples_files(test_file: str, symbols_by_file: Dict[str, Set[str]]):
```
david-yz-liu, line 92: Also add a return type (`None`)

tests/test_examples.py
```diff
@@ -34,39 +36,72 @@ def get_file_paths():
...
-    test_function = create_checker(test_file, checker_name)
-    test_function()
+
+    test_file_name = os.path.basename(test_file)
+    file_symbols = symbols_by_file[test_file_name]
+
+    found_pylint_message = checker_name in file_symbols
+    if not found_pylint_message:
+        print('Failed: ' + test_file)  # test doesn't say which file
+    assert found_pylint_message
```
david-yz-liu, line 107: Rather than using `if not found_pylint_message` and printing, you can add a string to the assert statement. Something like:

```python
assert found_pylint_message, f'File {test_file} does not cause the expected pylint error.'
```

david-yz-liu (commented):
@ckatsube nice job, I left a few more comments. Also I wonder if you can just pass in all files under `examples/pylint` to the call, and then the cyclic import check can be tested as well (we have it disabled currently). See #696.

david-yz-liu:
@ckatsube Sorry I forgot to reply to your question about the `\d`. The string needs to be marked as a "raw" string so that the backslash is taken literally and not as the start of an escape sequence (like `'\n'`). You can do this by prefixing the string literal with an `r`.

tests/test_examples.py
ckatsube: Hm could I get some insight into what could be going wrong on my side? I still get the following error output from Pytest (if you scroll all the way right you can see the trailing ANSI code).
```python
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
C:\Users\craig\AppData\Local\Programs\Python\Python39\lib\json\__init__.py:346: in loads
    return _default_decoder.decode(s)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
self = <json.decoder.JSONDecoder object at 0x000001760FDA9A60>
s = '[\r\n    {\r\n        "type": "convention",\r\n        "module": "C0103_invalid_name",\r\n        "obj": "",\r\n     ...n        "message": "\\"z\\" is not a valid mode for open.",\r\n        "message-id": "W1501"\r\n    }\r\n]\r\n\x1b[0m'
_w = <built-in method match of re.Pattern object at 0x000001760FD56AB0>
    def decode(self, s, _w=WHITESPACE.match):
        """Return the Python representation of ``s`` (a ``str`` instance
        containing a JSON document).
    
        """
        obj, end = self.raw_decode(s, idx=_w(s, 0).end())
        end = _w(s, end).end()
        if end != len(s):
>           raise JSONDecodeError("Extra data", s, end)
E           json.decoder.JSONDecodeError: Extra data: line 7725 column 1 (char 260860)
C:\Users\craig\AppData\Local\Programs\Python\Python39\lib\json\decoder.py:340: JSONDecodeError
```
I'm running this with
```python
>>> import pytest
>>> pytest.main(['tests/test_examples.py'])

**Comment by ckatsube** [comment]
Two unresolved comments:
1. regarding the trailing ANSI code. It appears to have passed for the Travis-CI; however, I still get the error when running the test locally.
2. still looking into whether `R0401_cyclic_import.py` can be readded.

File: tests/test_examples.py [hunk 9]
```diff
@@ -34,39 +37,45 @@ def get_file_paths():
...
         assert False
     checker_name = base_name[6:-3].replace('_', '-')  # Take off prefix and file extension.
-    test_function = create_checker(test_file, checker_name)
-    test_function()
+
+    test_file_name = os.path.basename(test_file)
+    file_symbols = symbols_by_file[test_file_name]
+
+    found_pylint_message = checker_name in file_symbols
+    if not found_pylint_message:
```
david-yz-liu, line 79: You can delete this part

david-yz-liu (commented):
Regarding the ANSI character, I don't know what's happening on your computer. Could be a pylint configuration issue (if you have other `.pylintrc` files, for example). The character `\x1b0[m` (equivalent to `\0330[m`) is used to reset ANSI colouring, and appears in Pylint here: https://github.com/PyCQA/pylint/blob/master/pylint/reporters/text.py. It could be that this is being printed from that reporter for some reason when pylint is being run, but I'm not sure why.

Regarding the cyclic import test, great investigation. You can leave it in the ignored list, but then also create a new test which runs pylint on the the two specific files with the cyclic import error. Make sure to document why this is a separate test.

ckatsube:
Hm. Found an issue that is unrelated to this PR. It seems like `C0103_invalid_name.py` will add its intended message regardless of the content of the file. The module name in and of itself is an invalid name and will add the following message.
```
examples\pylint\C0103_invalid_name.py:1:0: C0103: Module name "C0103_invalid_name" doesn't conform to snake_case naming style (invalid-name)
```
I don't think Pylint lets you disable a specific form of a message (I know you can disable a message entirely), so this might call for changing the naming convention of the files in `examples/pylint`--which sounds like it should be part of another PR.

PR #714, opened 2021-06-12
Bug Fix: accessing nonexistent DictComp object attribute

kavins14:
In the previous bug fix for the possibly undefined checker, the DictComp object case was being incorrectly processed. This PR fixes it.

## Your Changes

**Description**: DictComp object does not have  `elt` property, instead the `key` and `value` properties is used.

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist


david-yz-liu:
@kavins14 it would be great if you added a test case for this!

david-yz-liu:
Thanks, @kavins14!

PR #717, opened 2021-06-15
Remove Temp HTML Report

### LINKED ISSUE #711: HTMLReporter: Open PyTA report in web browser without creating temporary file
Currently the `HTMLReporter` renders its report in a new file, `pyta_report.html`, and then opens that file in a browser window. This leaves an extra file on the user's computer after running PyTA, which may not be what the user wants. Instead, let's try to mimic what plotly does with its browser rendering, where it opens the generated figures in a browser window without creating a separate file.

Concretely, there should be two updates to the current behaviour of `HTMLReporter`:

1. By default, the `HTMLReporter` should open a browser and display the PyTA report to it using the same idea as plotly, without creating a temporary file.
    - Also, an informative display message should be printed saying not just that the PyTA report is opening in the user's web browser, but that if they want to save the report to a file they should re-run PyTA but do "X" instead (see next point).
2. However, the Reporter should allow the user to specify an option to save the report to a file and then open up that file instead, using the same idea as what we do for `PlainReporter` already.

**Tip**: In the plotly source code, a good starting point is https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/io/_renderers.py.

simonchenwastaken:
## Motivation and Context
To make the experience of using PyTA more streamlined and less annoying for users to end up with a `pyta_report.html` file in their directories when not wanted.

(resolves #711)

## Your Changes

**Description**: Added it so that `python_ta.check_all()`'s `output` parameter determines whether an output file is created or not. If nothing is specified, by default, it will open a window similar to plotly. This does not create a temp file. If an output *is* specified, then a browser window will not be opened and the `pyta_report.html` file will be generated as before.

## Testing

Created a test file to run both `python_ta.check_all()` and  `python_ta.check_all(output="C:\\Users\\Simon Chen\\Desktop")` on. Worked as intended.

## Checklist


simonchenwastaken:
@david-yz-liu 
Hi David, these changes are a working solution to #711 , but very rough and unorganized and I would like to know what semantic changes to names / code changes / or other changes I should make to this current drafted version

david-yz-liu (commented):
@simonchenwastaken overall you're definitely on the right track, good job. Please see my Slack message about using the `output` argument rather than a separate config option.

You should tidy up your new helper functions with type annotations and docstrings, and also make sure to both reformat the docstring of the function you adapted from plotly, and inclue a reference the specific file on plotly's GitHub repo.

python_ta/__init__.py
```diff
@@ -188,11 +188,11 @@ def reset_linter(config=None, file_linted=None):
              'type': 'string',
              'metavar': '<pyta_reporter>',
              'help': 'Template file for html format of htmlreporter output.'}),
-        ('pyta-output-file',
-            {'default': 'output.html',
-             'type': 'string',
-             'metavar': '<pyta_reporter>',
-             'help': 'Output file for htmlreporter.'}),
+        ('pyta-save-html-report',
```
david-yz-liu, line 191: Okay, so I'm actually thinking that we should just delete this option entirely. Whether or not to save to an output file will be based on the argument to `check_all` only, not in the `.pylintrc` file.

Also review the default `.pylintrc` file and remove the old `pyta-output-file` option, if we included it.

python_ta/reporters/html_reporter.py
```diff
@@ -14,6 +15,7 @@
 
 TEMPLATES_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'templates')
 
+
```
david-yz-liu, line 18: There are a few style changes which are correct but beyond the scope of this PR. Please revert them here and make them in a separate pull request.

python_ta/reporters/html_reporter.py
```diff
@@ -62,8 +64,8 @@ def print_messages(self, level='all'):
 
         MessageSet = namedtuple('MessageSet', 'filename code style')
         append_set = MessageSet(filename=self.filename_to_display(self.current_file_linted),
-                               code=self._messages_shown(self._sorted_error_messages),
-                               style=self._messages_shown(self._sorted_style_messages))
+                                code=self._messages_shown(self._sorted_error_messages),
```
david-yz-liu, line 67: Revert this and the line below (see above comment)

python_ta/reporters/html_reporter.py
```diff
@@ -84,21 +86,51 @@ def output_blob(self):
...
-        output_path = os.path.join(os.getcwd(), self.linter.config.pyta_output_file)
+
+        # Render the jinja template
+        rendered_template = template.render(date_time=dt, reporter=self)
+
+        # If a filepath was specified, write to the file
+        if self._output_filepath:
+            self._write_html_to_file(rendered_template)
+        else:
+            self._open_html_in_browser(rendered_template.encode('utf8'))
```
david-yz-liu, line 97: I'm thinking it might be best to do the `.encode('utf8')` call for both kinds of outputs, which means you could add it to Line 91 above.

You can also test this yourself by running PythonTA on a file where there's an error on a line with Unicode characters (you can include such characters in variable names, for example).

python_ta/reporters/html_reporter.py
```diff
@@ -84,21 +86,51 @@ def output_blob(self):
...
+
+    def _write_html_to_file(self, rendered_template):
```
david-yz-liu, line 99: Please include type annotations (we haven't always been consistent on this, but should be for new code).

python_ta/reporters/html_reporter.py
```diff
@@ -84,21 +86,51 @@ def output_blob(self):
...
         with open(output_path, 'w', encoding='utf-8') as f:
-            f.write(template.render(date_time=dt,
-                                    # pyta_logo=pyta_logo_base64_encoded,
-                                    reporter=self))
-        print('Opening your report in a browser...')
-        output_url = 'file:///{}'.format(output_path)
-        webbrowser.open(output_url, new=2)
+            f.write(rendered_template)
+
+    def _open_html_in_browser(self, html, new=0, autoraise=True):
```
david-yz-liu, line 105: For our purposes you can omit the `new` and `autoraise` arguments. We hard-coded these in the old code.

python_ta/reporters/html_reporter.py
```diff
@@ -84,21 +86,51 @@ def output_blob(self):
...
+                self.send_response(200)
+                self.send_header("Content-type", "text/html")
+                self.end_headers()
+
+                buffer_size = 1024 * 1024
+                for i in range(0, len(html), buffer_size):
+                    self.wfile.write(html[i: i + buffer_size])
+
+        server = HTTPServer(('127.0.0.1', 0), OneShotRequestHandler)
+        webbrowser.open(f"http://127.0.0.1:{server.server_port}", new=new, autoraise=autoraise)
```
david-yz-liu, line 125: Note that previously, the `new` parameter was set to `2`.

python_ta/reporters/html_reporter.py
```diff
@@ -84,21 +86,51 @@ def output_blob(self):
...
+        server.handle_request()
 
     @classmethod
     def _vendor_wrap(self, colour_class, text):
         """Override in reporters that wrap snippet lines in vendor styles, e.g. pygments."""
         if '-line' not in colour_class:
             text = highlight(text, PythonLexer(),
-                            HtmlFormatter(nowrap=True, lineseparator='', classprefix='pygments-'))
+                             HtmlFormatter(nowrap=True, lineseparator='', classprefix='pygments-'))
```
david-yz-liu, line 133: Revert this (see above comment)

python_ta/__init__.py
simonchenwastaken: that's strange, I thought I removed the `pyta-save-html-report` on the remote commit, that's my bad. Also, do you want me to remove the `pyta-output-file` option from `__init__.py` too?

python_ta/reporters/html_reporter.py
simonchenwastaken: okay, the only thing I was scared of originally not to pass in a `.encode`d version of `rendered_template` to `self._write_html_to_file` was because the original code to write to the file already had something that handled encoding and I did not want to modify too much of the old functionality and code.

```python
    def _write_html_to_file(self, rendered_template):
        output_path = os.path.join(os.getcwd(), self.linter.config.pyta_output_file)
        with open(output_path, 'w', encoding='utf-8') as f:
            f.write(rendered_template)
```

would I need to change anything to the function if I passed in an `.encode('utf8')`d version?

python_ta/reporters/html_reporter.py
simonchenwastaken: okay, thank you for the clarification. I didn't include them originally to try and be consistent with the codebase, but I will gladly add them for future changes 👍

simonchenwastaken:
Hi david, I think I have done all your requested changes besides adding tests so far. There also was an issue with the `_write_output_to_html()` method that didn't add the `.html` to the filename extension. This might have been a byproduct of removing the config option for filename, but I managed to fix it by making a modification to `OUTPUT_FILENAME` in `plain_reporter.py`, if you were wondering why I changed that. I will add tests for the changes soon.

python_ta/__init__.py
david-yz-liu: Yes, please remove the `pyta-output-file` option from this file.

python_ta/reporters/html_reporter.py
david-yz-liu: @simonchenwastaken you should test it yourself, in the way that I described.

python_ta/__init__.py
david-yz-liu: @simonchenwastaken Yes, remove the option.

python_ta/reporters/plain_reporter.py
```diff
@@ -5,7 +5,7 @@
 from collections import defaultdict, namedtuple
 from .node_printers import LineType, render_message
 
-OUTPUT_FILENAME = 'pyta_output'
+OUTPUT_FILENAME = 'pyta_output.html'
```
david-yz-liu, line 8: @simonchenwastaken so it's not great to change this extension because for the `Plain` and `Color` reporter, the output is plaintext, not HTML.

What I recommend doing is moving this constant into the `PlainReporter` class itself, so that in the methods you can access `self.OUTPUT_FILENAME` instead. Then, this constant can be overridden in `HTMLReporter`.

david-yz-liu (commented):
@simonchenwastaken I think I forgot to mention this, but you also should remove the `pyta-output-file` option from `.pylintrc`.

PR #720, opened 2021-06-19
Resolve Issue 673: no longer checks subclass type annotations when exiting super's `__init__`

### LINKED ISSUE #673: python_ta.contracts.check_all_contracts checks for class attributes of child class after exiting super class' init
# Issue description
## Functionality with issue
`python_ta.contracts.check_all_contracts()`
python_ta ver. == 1.6.3
## Description
`check_all_contracts` will throw an `AttributeError` that the subclass object does not have one of its instance attributes immediately after exiting its super class' init. Subclass attributes should be allowed to be instantiated after calling the super class' init, not before.

# Reproducible example
```python

"""Reproducible Test Case
"""


class A:
    """Parent class with an attribute annotation"""

    my_int: int

    def __init__(self, my_int: int):
        self.my_int = my_int


class B(A):
    """Child class with another attribute annotation"""

    my_int2: int

    def __init__(self, my_int2: int):
        super().__init__(3)
        self.my_int2 = my_int2


if __name__ == "__main__":
    import python_ta.contracts
    python_ta.contracts.check_all_contracts()

    my_b = B(5)

```
Error message:
```python
AttributeError: 'B' object has no attribute 'my_int2'
```

ckatsube:
## Motivation and Context

This PR aims to resolve #673, as super classes are not expected to have initialized the subclass' instance attributes nor are subclasses expected to instantiate them before calling `super().__init__`.

## Your Changes

**Description**: the two commits are actually two different versions proposed to resolve this issue.
1st commit (3e3f7fc): after calling the instance's wrapped method, the type annotation check is run on the class in `instance`'s hierarchy that the wrapped method was defined in. I thought this was appropriate as a non-overriden super class method is not responsible for modifying any of its subclass' instance attributes. Even if said subclass overrode one of its super's instance attributes, by LSP the behavior when using the super's method should be the same as if it were just an instance of the super. This does somewhat lean on the wrapped class utilizing proper OOP design, which leads to the next version.
2nd commit (bfcad20): the post-method call type check is only run if the wrapped method is the public/outermost method (it's not an overriden method). I also thought this could be appropriate because an overriden method would be called from the method that overrides it, so ultimately checking all type annotations after the outermost method is finished would also be after all the super calls but before the function exists back to the client code.

## Testing

Confirmed that the pytest suite ran.
Confirmed that the MR provided in the issue passes.

Couldn't quite wrap my head around how to build a test suite for these changes, but one may get added if I figure it out.

## Checklist


ckatsube:
Changelog:
added type annotations to the `tests/test_subclass_contracts.py` classes. Confirmed that adding them would throw the error in #673 and that the provided change will fix it.

david-yz-liu:
@ckatsube great research on this, and I really liked the two different approaches you mentioned. They get at two different limitations of our current approach:

1. An instance method should only be checking type annotations for the class where the method was defined and superclasses of that class, but not subclasses.
2. When `__init__` is called from client code, we should be able to use helper methods (from superclasses or from the same class) to initialize attributes where the type annotations (and in fact all representation invariants) aren't checked yet. We should only check these when the original call to `__init__` returns, and not earlier.

I think I'll get you to address these in two separate PRs. For this current PR, you can tackle the first item by using a variation of the approach from the first commit. But to make it cleaner, modify the interface for `_instance_method_wrapper` so that it takes in the `klass` value from the place where it's called, rather than `rep_invariants`. This way you'll be able to pass `klass` directly to the `_check_class_type_annotations` helper to retrieve its type annotations.

Also, the change you made to the test cases is a good one, but remove the blank line between the class docstring and the attribute type annotations that you added.

david-yz-liu:
@ckatsube one thing I forgot to mention about testing: this change should affect all superclass methods, not just `__init__`. So you should be able to add tests that check that when a superclass method is called inside a subclass method, the superclass method doesn't check the attribute types or representation invariants of the subclass.

ckatsube:
The changes were a lot more than I expected so the changelog is a little more extensive than usual.

Changelog:

`tests/test_subclass_contracts.py`:
- removed space between docstring and type annotations

`python_ta/contracts/__init__.py`:
- updated `_instance_method_wrapper` to take in the class that defined the wrapped method by replacing `rep_invariants` (which can still be accessed from the class itself.
  - extracted the `rep_invariants is None` check to the call-point. There is a slight difference in that before, `check_contracts` decorator (which was returned by `_instance_method_wrapper`) was assigned directly as the method instead of wrapping the method--now `check_contracts(value)` (where `value` is the method to be wrapped) is assigned instead. I believe the previous behavior was unintentional as the decorator was being assigned directly without actually decorating anything.
  - changed `klass_mod` in this changed helper to take the global context of the defined class, not the instance's class. I believe this was the intended behavior (as this was how it was done in the `new_setattr`) but the earlier versions did not have access to the true class context and just used the lesser ideal instance's class.
- replaced `rep_invariants` parameter of `_check_invariants` for `klass`. Seemed redundant to force the client code in `_instance_method_wrapper` to re-access the invariants or to pass it through `_instance_method_wrapper` from its `add_class_invariants` parent caller when the invariants can just be accessed directly from the class. There is the argument that you might want to pass in invariants that are not directly tied to a singular class and maybe a mix of classes, but the current code base didn't seem to demand that kind of behavior.
  - this change did require `new_setattr` in `add_class_invariants` to pass in the class (which it already had access to) instead of the invariants.

ckatsube:
> @ckatsube one thing I forgot to mention about testing: this change should affect all superclass methods, not just `__init__`. So you should be able to add tests that check that when a superclass method is called inside a subclass method, the superclass method doesn't check the attribute types or representation invariants of the subclass.

Alright. I'll get working on this as well!

python_ta/contracts/__init__.py
```diff
@@ -143,29 +144,28 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
     return r
 
 
-def _instance_method_wrapper(wrapped, rep_invariants=None):
-    if rep_invariants is None:
-        return check_contracts
```
david-yz-liu, line 148: It's good that you flagged this, although the fix should be a bit different. It's okay if there are no representation invariants, we should still execute the `wrapper` so that the type annotations are checked. So you should revert the change on line 102/104, _keep_ the removal of these two lines, and see my comment below.

python_ta/contracts/__init__.py
```diff
@@ -143,29 +144,28 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
...
+def _instance_method_wrapper(wrapped, klass: type):
```
david-yz-liu, line 147: Add type annotations for `wrapped` and return type as well.

python_ta/contracts/__init__.py
```diff
@@ -143,29 +144,28 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
...
 
     return wrapper(wrapped)
 
-def _check_class_type_annotations(instance: Any) -> None:
+
+def _check_class_type_annotations(klass: type, instance: Any) -> None:
     """Check that the type annotations for the class still hold.
     """
-    klass = instance.__class__
+    assert isinstance(instance, klass)
```
david-yz-liu, line 168: Make this a precondition instead.

python_ta/contracts/__init__.py
```diff
@@ -178,9 +178,14 @@ def _check_class_type_annotations(instance: Any) -> None:
                 f'{repr(value)} did not match type annotation for attribute "{attr}: {annotation}"')
 
 
-def _check_invariants(instance, rep_invariants: Set[str], global_scope: dict) -> None:
+def _check_invariants(instance, klass: type, global_scope: dict) -> None:
     """Check that the representation invariants for the instance are satisfied.
+
+    Precondition:
```
david-yz-liu, line 184: Remove this precondition. Instead, use `getattr(klass, '__representation_invariants__', set())` to use a default value of an empty set.

tests/test_subclass_contracts.py
```diff
@@ -29,19 +33,53 @@ class Developer(Employee):
...
+    """
+    currently_teaching: List[str]
+    wage_per_class: float
+
+    def __init__(self, name, wage_per_class, currently_teaching):
+        Employee.__init__(self, name, wage_per_class * len(currently_teaching))
+        self.wage_per_class = wage_per_class
+        self.currently_teaching = currently_teaching
+
+    def teach_new_classes(self, currently_teaching):
```
david-yz-liu, line 58: This is a good test, but you should also include another version with the opposite execution order: `self.change_wages` is called first, then `self.currently_teaching` is reassigned. This is what I meant by my earlier comment: `self.change_wages` will violate a representation invariant for `Teacher` but not `Employee`, and so no error should be raised in this case either.

tests/test_subclass_contracts.py
```diff
@@ -128,3 +171,35 @@ def test_change_teamlead_team_invalid(teamlead):
         teamlead.team = ''
     msg = str(excinfo.value)
     assert 'len(self.team) > 0' in msg
+
+
+def test_teacher_rep_violatable(pe_bio_teacher):
```
david-yz-liu, line 176: I think this test and `test_method_violates_class_invariant` are redundant with the ones using the `Developer` class.

david-yz-liu:
@ckatsube btw there's a small merge conflict because of #721, your change still applies, the function call just moved a bit.

tests/test_subclass_contracts.py
ckatsube: I'm making a case that I want to keep one of the two as a test to check that the class' invariant is written properly and actually works so it isn't being skipped and throwing false positives.

david-yz-liu: I see what you're saying, but I do not think this is important enough to justify the additional code. However, instead of creating a new subclass you can reuse the existing subclass (`Developer`) and just create a new method and test with that.

That said, next time I ask you to do something, you should do it even if you want to argue against it. That way I can respond and merge in your pull request at once, rather than waiting another cycle.

ckatsube: Understood. I'll make sure to do that in the future

PR #722, opened 2021-06-21
Implement missing space in doctest checker

### LINKED ISSUE #719: New checker: missing space in doctest example
I'd like to see a new checker that scans function docstrings for uses of `>>>` that aren't following by a space. In this case, the doctest won't actually be picked up and run the doctest. Example:

```python
def f(x: int) -> int:
    """Return one plus x.
    
    >>>f(10)  # Won't actually be parsed as a doctest!
    11
    """
```

lorena-b:
## Motivation and Context

Resolves #719 

This pull request implements a new checker that flags doctests not followed by a space. 

## Your Changes

**Description**: 
- Updated documentation and examples file to contain the new checker
- Implemented the checker 
- Added the checker to `custom_checkers` in python_ta/\_\_init\_\_.py
- Created a test suite for the checker 

## Testing

A test suite was created to test this checker. 

## Checklist


python_ta/checkers/missing_space_in_doctest_checker.py
```diff
@@ -0,0 +1,61 @@
...
+from pylint.checkers.utils import check_messages
+
+DOCTEST = ">>>"
+
+
+class MissingSpaceInDoctestChecker(BaseChecker):
+    __implements__ = IAstroidChecker
+
+    name = 'missing_space_in_doctest'
+    msgs = {'E9973': ('Space missing in doctest(s) in the docstring of function "%s."',
```
david-yz-liu, line 15: We can make this more specific by saying "Space missing after >>>". Also, use singular "doctest" only, I think in general the message style is to refer to a single occurrence of the error.

python_ta/checkers/missing_space_in_doctest_checker.py
```diff
@@ -0,0 +1,61 @@
...
+    def visit_functiondef(self, node: astroid.FunctionDef) -> None:
+        """Visit a function definition"""
+        docstring = node.doc
+
+        if docstring is not None:
+            start_line = node.lineno
+            lines = docstring.split('\n')
+            line_no = start_line
+
+            for line in lines:
```
david-yz-liu, line 32: It's a little cleaner to use `enumerate` here, so that you don't need to use a manual `line_no` counter.

python_ta/checkers/missing_space_in_doctest_checker.py
```diff
@@ -0,0 +1,61 @@
...
+                line_no += 1
+                if self._has_doctest(line):
```
david-yz-liu, line 34: Overall, these two functions should be combined into a single predicate that looks for both a `>>>` and that isn't followed by a space. Also, the test should be more specific so that it only recognizes `>>>` as the start of a doctest when it's the first non-whitespace characters in the line. So for example a line like `abc >>> 1 + 2` isn't recognized as a doctest.

Matching this pattern (whitespace, `>>>`, and then a non-space character) is a really good use of [regular expressions](https://docs.python.org/3/howto/regex.html), so I recommend reading about that.

python_ta/reporters/node_printers.py
```diff
@@ -89,13 +89,25 @@ def render_too_many_arguments(msg, source_lines=None):
...
 
 
+def render_missing_space_in_doctest(msg, source_lines=None):
+    """Render a missing space in doctest message"""
+    node = msg.node
+    line = msg.line
+    start = node.lineno
+
+    # Display code snippet from the start of the function to the line containing the error
+    yield from render_context(start, line, source_lines)
```
david-yz-liu, line 99: For long docstrings it's not good to show this many lines. Instead, you should use the same general approach (in `render_generic`), which displays 2 lines before and after the erroneous line.

PR #723, opened 2021-07-01
Allowing Helper Methods to be called from __init__

ckatsube:
## Motivation and Context

This PR aims at fixing the unintended behavior where calling a helper method from the `__init__` method would prematurely check class post-conditions (class type annotations and representation invariants). This is unintended behavior as a class should be allowed to initialize before having these conditions checked.

## Your Changes

**Description**:
Added extra check layer between method call and post-condition check. The additional check iterates up the call-stack to see whether the `__init__` for the method's instance exists--if said `__init__` does exist then it follows that the instance is still being initialized hence the post-condition check should be skipped.

## Testing

Added test suite to `tests/test_class_contracts.py` (part of PR).

Note: this PR is overly strict in the sense that already-initialized parent class contracts can still be violated under the hood due to the check skipping if the instance as a whole is being initialized--not the specific class the called method was defined in. This could be a possible issue due to a situation where a subclass uses its parent's initialization helper but violates said parent's representation invariants. Even if the specific violation is remedied in the subclass, this situation may cause the parent functionality to malfunction entirely as the invariants were violated once outside of class-internally-defined behavior.

## Checklist

### DIFF
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -184,8 +184,7 @@ def _instance_method_wrapper(wrapped: Callable, klass: type) -> Callable:
         try:
             r = _check_function_contracts(wrapped, instance, args, kwargs)
 
-            previous_frame = inspect.currentframe().f_back
-            if _is_exception_callsite(previous_frame, klass):
+            if not _klass_is_initialized(klass, instance):
                 return r
 
             _check_class_type_annotations(klass, instance)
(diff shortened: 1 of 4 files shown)


python_ta/contracts/__init__.py
```diff
@@ -194,6 +196,19 @@ def wrapper(wrapped, instance, args, kwargs):
...
 
 
+def _instance_init_in_callstack(instance) -> bool:
+    """Return whether instance's init is part of the current callstack
+    """
+    frame = inspect.currentframe().f_back
+    while frame:
+        frame_context_name = inspect.getframeinfo(frame).function
+        frame_context_self = frame.f_locals.get('self', None)
+        if frame_context_name == '__init__' and frame_context_self == instance:
```
david-yz-liu, line 206: use `is` instead of `==` for the latter check

python_ta/contracts/__init__.py
```diff
@@ -194,6 +196,19 @@ def wrapper(wrapped, instance, args, kwargs):
     return wrapper(wrapped)
 
 
+def _instance_init_in_callstack(instance) -> bool:
+    """Return whether instance's init is part of the current callstack
+    """
+    frame = inspect.currentframe().f_back
+    while frame:
+        frame_context_name = inspect.getframeinfo(frame).function
+        frame_context_self = frame.f_locals.get('self', None)
```
david-yz-liu, line 205: You can be a bit more specific here and check that the first parameter name `== 'self'` as well.

That said, there's still the possibility of having a method which is called from a top-level function whose first parameter is `self`, which would cause contract checking to be skipped. This is a bit subtle, but probably worth documenting. Please add a note about this to the documentation page for PyTA's contract checking.

python_ta/contracts/__init__.py
```diff
@@ -194,6 +196,19 @@ def wrapper(wrapped, instance, args, kwargs):
     return wrapper(wrapped)
 
 
+def _instance_init_in_callstack(instance) -> bool:
```
david-yz-liu, line 199: I suppose you can use a type annotation of `Any` here.

tests/test_class_contracts.py
```diff
@@ -303,5 +303,59 @@ def test_override_len_simple() -> None:
...
+    - self.theme.lower() in {"dark", "light"}
+    - self.primary_color.isalpha()
+    - self.secondary_color.isalpha()
+    """
+    size: int
+    theme: str
+    primary_color: str
+    secondary_color: str
+
+    def __init__(self, theme: str, color_palette: Tuple[str, str], options: dict = None):
```
david-yz-liu, line 318: Add return type annotations here and below.

tests/test_class_contracts.py
```diff
@@ -303,5 +303,59 @@ def test_override_len_simple() -> None:
     OverrideLen()
 
 
+class ThemedWidget:
+    """
+    Representation Invariants:
+    - self.theme.lower() in {"dark", "light"}
```
david-yz-liu, line 309: Use single quotes for string literals (throughout).

tests/test_subclass_contracts.py
```diff
@@ -163,7 +163,7 @@ def test_change_teamlead_name(teamlead):
 
 def test_change_teamlead_team_invalid(teamlead):
     """
-    Changes the team of the teamlead to something invalid. Excpects an exception.
+    Changes the team of the teamlead to something invalid. Expects an exception.
```
david-yz-liu, line 166: Revert the changes in this file.

ckatsube:
Made the commented changes.
The primary one I was concerned about was regarding checking the first parameter name. I took the first variable name from the <code>'s `.co_varnames` which appeared to be the appropriate approach. I did try testing some other scenarios to see whether `self` would only appear first in the intended scenario, but ultimately I'm unsure if I thought of enough cases.

david-yz-liu:
@ckatsube good work. I was actually thinking of a different situation, which is a top-level function called `__init__` with first parameter `self`. But your example in the documentation illustrates the same issue, which is good with me!

PR #726, opened 2021-07-13
Add suggestions section to contract checker

ckatsube:
## Motivation and Context

This PR aims at adding a section in the `python_ta/contracts` for additional helpful suggestions to appear during specific circumstances. The added section is specifically for the function argument type annotation preconditions.
The added suggestion is detecting whether the type of an object is passed in instead of its instance.
When conditionally creating subclasses, I've seen new Python developers accidentally forget to actually instantiate their objects, and linters appear to not recognize the potential faulty argument with the conditional branches; furthermore, the actual assertion error might be difficult for a young Python developer read as it would say something like `'argument <class YourClass> did not match type annotation for parameter <class YourClass>'`. The suggestion helps make it clear that the type was passed in whereas an instance was expected.

## Your Changes

**Description**:
The exception handler in the function argument type annotation check has been updated to also allow additional suggestions to be included in the PyTAContractError message.
Although only one suggestion was incorporated for the time being, an entire helper was given to more easily allow for more suggestions to be added in the future (instead of the next person working trying to add a suggestion having to dig into the exception block and extract the suggestions-related logic themselves).

## Testing

Added two test cases to show that the suggestion is added when expected and that it's not erroneously added when not expected.

## Checklist

### DIFF
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -153,15 +153,13 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
                 _debug(f'Checking type of parameter {param} in call to {wrapped.__qualname__}')
                 check_type(param, arg, annotations[param])
             except TypeError:
-                error_message = (
-                    f'{wrapped.__name__} argument {repr(arg)} did not match type annotation for parameter '
-                    f'"{param}: {annotations[param]}"'
-                )
                 additional_suggestions = _get_argument_suggestions(arg, annotations[param])
-                if additional_suggestions:
-                    error_message += f'\n{additional_suggestions}'
 
-                raise PyTAContractError(error_message)
+                raise PyTAContractError(
+                    f'{wrapped.__name__} argument {repr(arg)} did not match type annotation for parameter '
+                    f'"{param}: {annotations[param]}"' +
+                    (f'\n{additional_suggestions}' if additional_suggestions else '')
+                )
 
     # Check function preconditions
     preconditions = parse_assertions(wrapped)
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -153,9 +153,15 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
                 _debug(f'Checking type of parameter {param} in call to {wrapped.__qualname__}')
                 check_type(param, arg, annotations[param])
             except TypeError:
-                raise PyTAContractError(
+                error_message = (
                     f'{wrapped.__name__} argument {repr(arg)} did not match type annotation for parameter '
-                    f'"{param}: {annotations[param]}"')
+                    f'"{param}: {annotations[param]}"'
+                )
+                additional_suggestions = _get_argument_suggestions(arg, annotations[param])
+                if additional_suggestions:
+                    error_message += f'\n{additional_suggestions}'
+
+                raise PyTAContractError(error_message)
 
     # Check function preconditions
     preconditions = parse_assertions(wrapped)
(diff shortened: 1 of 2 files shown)


ckatsube:
Unsure of why the Travis-CI and Coverage checks did not start, but running the tests locally I get:

Tests:
- 427 passed, 2 warnings in 118.65s (0:01:58) 

Coverage:
- 195 misses out of 1890 statements for a 0.8969% coverage (unsure of how that compares to from before).

python_ta/contracts/__init__.py
```diff
@@ -176,6 +180,15 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
     return r
 
 
+def _get_argument_suggestions(arg: Any, annotation: type) -> str:
+    """Returns potential suggestions for the given arg and its annotation"""
+    suggestions = ""
```
david-yz-liu, line 185: I don't think we're going to accumulate suggestion texts very much, so I prefer just using an `if ... else ...` structure for this function.

python_ta/contracts/__init__.py
```diff
@@ -176,6 +180,15 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
...
+    if isinstance(arg, type) and annotation in arg.mro():
```
david-yz-liu, line 186: You can use the `issubclass` function (which also returns `True` if its two arguments are the same).

python_ta/contracts/__init__.py
```diff
@@ -176,6 +180,15 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
...
+        suggestions += f'Did you mean to pass in an instance of {arg}?'
```
david-yz-liu, line 187: I think we have this problem somewhere else, but it's better to use `arg.__name__`, which evaluates to just the name of the type.

I'm wondering if this message should be a little more explicit about how the code should change. It might be good to think about a concrete example and run the text by the other PyTA team members.

> Did you mean to pass in an instance of Player?

> Did you pass in Player instead of Player(...)?

> Did you forget the parentheses after Player?

etc. Do some brainstorming here!

tests/test_contracts.py
```diff
@@ -166,3 +166,55 @@ def test_search_invalid() -> None:
...
+    when changing the suggestion text.
+    """
+
+    # Ternary assignment tricks the PyCharm linter into not recognizing the faulty argument
+    player = HumanPlayer if True else CPU()
+    with pytest.raises(AssertionError) as excinfo:
+        _is_cpu(player)
+
+    msg = str(excinfo.value)
+    assert 'Did you mean to pass in an instance of' in msg
```
david-yz-liu, line 204: Whatever message we end up going with (see above comment), that full message text should be searched for in this test, not just the common prefix. That is, we want to make sure the name of the class is incorporated correctly as well!

Also, there should be a separate test for when the argument `is` the type annotation, not one of its subclasses. (In this case, that would mean passing in `Player`.

tests/test_contracts.py
```diff
@@ -166,3 +166,55 @@ def test_search_invalid() -> None:
...
+
+def test_type_not_instance_error() -> None:
+    """Test that the additional suggestion is added when the type of a class is passed in as an
+    argument instead of its instance
+
+    This test is coupled to the suggestion's arbitrarily chosen text, hence should be updated
+    when changing the suggestion text.
+    """
+
+    # Ternary assignment tricks the PyCharm linter into not recognizing the faulty argument
```
david-yz-liu, line 198: The PyCharm linter doesn't actually matter for this test, since we're deliberately testing "bad" code. So it's actually better to make the test more explicit just by calling `_is_cpu(HumanPlayer)` in the test. And same below.

python_ta/contracts/__init__.py
```diff
@@ -176,6 +180,14 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
     return r
 
 
+def _get_argument_suggestions(arg: Any, annotation: type) -> str:
+    """Returns potential suggestions for the given arg and its annotation"""
+    if isinstance(arg, type) and issubclass(arg, annotation):
```
david-yz-liu, line 185: One thing I forgot about: we need to protect from runtime errors if `annotation` isn't a valid second argument to `issubclass`. This can happen if the annotation isn't a valid type, *or* if it's a generic, like `list[int]` and `List[int]`.

The easiest way to handle this is to use a try-except block (`issubclass` will raise a `TypeError`).

tests/test_contracts.py
```diff
@@ -166,3 +166,62 @@ def test_search_invalid() -> None:
...
     msg = str(excinfo.value)
     assert 'all({n + m > 0 for n in numbers for m in numbers})' in msg
+
+
+class Player:
+    user: str
+
+
+class CPU(Player):
+    def __init__(self):
```
david-yz-liu, line 176: Add a return type annotation here

Issue #730, opened 2021-07-20
PyTA Error Report Redesign 


lorena-b:
This issue contains the changes to be made to the PyTA error report.

## Layout/Design:
- [x] Fit error report and TOC on the page 
- [x] Round corners of the error report 
- [x] Change the button design and add a hover effect
- [x] Update header and align to the left 
- [x] Add drop down animations (for dropdown code blocks)
- [x] Dock footer to the complete bottom of the page 
- [x] Add more information to footer

## Table of contents:
- [x] Align TOC to the left 
- [x] Display the errors and provide a link to the error on the report 
- [x] Hide/show table of contents 
- [x] Make TOC scrollable 

## Functionality: 
- [ ] #754
- [x] Make error link to documentation more obvious

PR #731, opened 2021-07-24
Added Table of Contents to template.html

ckatsube:
## Motivation and Context

Introducing the Table of Contents outlined in Issue #730.
This PR has very minimal changes to the current template and is aimed to lay the groundwork to create further PRs without causing unwanted merge conflicts between changes in the `<main>` and `<aside>`. 

Current ToC features:
- locks to top of screen when scrolling on the main element
- supports scrolling when the list of errors exceeds the size of the element
- maintains fixed width when resizing
- currently has full-screen height (probably a change that's up for debate or can be changed in a later PR)

## Your Changes

**Description**:
Changes can be broken up largely into two parts.
1. Added an `<aside>` element which essentially entirely contains the Table of Contents.
2. Wrapped the `<header>`, `<main>`, and new `<aside>` in a div that uses a grid layout.

The main motivation for the grid layout is to avoid using absolute or fixed positioning on the aside, as these two layouts may cause the ToC to overlap with the main component when changing the zoom unless the main component is also changed to be fixed. Grid seemed like a much cleaner and straight-forward approach to the column-like layout. 

## Testing

Mainly design changes so testing was limited to just running `python_ta.check_all()` on a number of local (untracked) files.
Main things tested:
- multiple files
- multiple messages in both Errors and Style categories

## Checklist

## SCREENSHOTS OF CHANGES

Images of a general multi-file and multi-message report
image
image
image
image

Image of scrollable ToC
image

### DIFF
--- a/python_ta/reporters/templates/stylesheet.css
+++ b/python_ta/reporters/templates/stylesheet.css
@@ -22,11 +22,6 @@ header {
     grid-template-columns: 20em 1fr;
 }
 
-aside {
-    padding-left: 16px;
-    padding-right: 16px;
-}
-
 .sidebar {
     margin-top: 0;
 
@@ -87,15 +82,6 @@ section {
     background: #141E30;
     background: linear-gradient(to left, rgba(36, 59, 85, 0.67), rgba(20, 30, 48, 0.77));
     background: -webkit-linear-gradient(right, rgba(36, 59, 85, 0.67), rgba(20, 30, 48, 0.77));
-}
-
-aside ul {
-    list-style-type: none;
-    padding-left: 20px
-}
-
-aside ul li {
-    margin: 3px 0;
 }
 
 .error-output {
--- a/python_ta/reporters/templates/stylesheet.css
+++ b/python_ta/reporters/templates/stylesheet.css
@@ -13,11 +13,13 @@ header {
     -webkit-font-smoothing: antialiased;
     color: #fff;
     text-align: center;
+
+    grid-column: 1 / span 2;
 }
 
 .main-grid {
     display: grid;
-    grid-template-columns: 20em 1fr
+    grid-template-columns: 20em 1fr;
 }
 
 aside {
@@ -25,12 +27,13 @@ aside {
     padding-right: 16px;
 }
 
-#sidebar {
+.sidebar {
     height: 100vh;
     width: 300px;
     display: block;
     margin-left: auto;
     margin-right: auto;
+    margin-top: 0;
 
     top: 0;
     position:sticky;
--- a/python_ta/reporters/templates/stylesheet.css
+++ b/python_ta/reporters/templates/stylesheet.css
@@ -15,6 +15,30 @@ header {
     text-align: center;
 }
 
+.main-grid {
+    display: grid;
+    grid-template-columns: 20em 1fr
+}
+
+aside {
+    padding-left: 16px;
+    padding-right: 16px;
+}
+
+#sidebar {
+    height: 100vh;
+    width: 300px;
+    display: block;
+    margin-left: auto;
+    margin-right: auto;
+
+    top: 0;
+    position:sticky;
+
+    overflow-y: auto;
+    overscroll-behavior-y: contain;
+}
+
 code {
     white-space: pre;
     font-size: 120%;
@@ -58,6 +82,25 @@ section {
     background: #141E30;
     background: linear-gradient(to left, rgba(36, 59, 85, 0.67), rgba(20, 30, 48, 0.77));
     background: -webkit-linear-gradient(right, rgba(36, 59, 85, 0.67), rgba(20, 30, 48, 0.77));
+}
+
+.section-heading {
+    display: flex;
+    padding: 0 15px;
+    color: rgba(255, 255, 255, 0.9);
+    text-shadow: 0px 1px 2px black;
+    background: #141E30;
+    background: linear-gradient(to left, rgba(36, 59, 85, 0.67), rgba(20, 30, 48, 0.77));
+    background: -webkit-linear-gradient(right, rgba(36, 59, 85, 0.67), rgba(20, 30, 48, 0.77));
+}
+
+aside ul {
+    list-style-type: none;
+    padding-left: 20px
+}
+
+aside ul li {
+    margin: 3px 0;
 }
 
 .error-output {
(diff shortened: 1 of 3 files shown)


ckatsube:
The Git difference looks really weird, but I broke up the commits to show that
First commit) added the `<aside>`, its appropriate css, and the wrapping `<div`> with grid layout WITHOUT changing the indentation of the wrapped components.
Second commit) Just indented the wrapped components. This is the one where the git diff has difficulty showing, but this commit is actually just "Highlight -> Tab"

python_ta/reporters/templates/stylesheet.css
```diff
@@ -15,6 +15,30 @@ header {
     text-align: center;
 }
 
+.main-grid {
+    display: grid;
+    grid-template-columns: 20em 1fr
```
david-yz-liu, line 20: Insert a semicolon here

python_ta/reporters/templates/template.html
```diff
@@ -8,138 +8,182 @@
...
-                                        <span class="slider">
-                                            <button style="float: right; background: #458B00;"> Expand/Collapse </button>
-                                        </span>
-                                    </h4>
-                                    {% if indiv.snippet != '' %}
-                                    <div class="collapsible">
-                                      <code>{{ indiv.snippet }}</code>
-                                    </div>
+        <div class="main-grid">
+            <header style="grid-column: 1 / span 2">
```
david-yz-liu, line 12: Don't use inline styles, instead you can add this to the CSS. You should fix this throughout this file.

python_ta/reporters/templates/template.html
```diff
@@ -8,138 +8,182 @@
...
+                <h1 class="title">PyTA Error Report</h1>
+                <time>{{ date_time }}</time>
+            </header>
+            <aside style="position: sticky">
```
david-yz-liu, line 16: You don't need an aside here. Just use the `section` below. In terms of positioning, I don't think you need to use `sticky` here. You should be able to use CSS grid layout to keep the table of contents on the left-hand side.

python_ta/reporters/templates/template.html
```diff
@@ -8,138 +8,182 @@
...
+                <section id="sidebar">
```
david-yz-liu, line 17: Use a class, not an id.

python_ta/reporters/templates/template.html
```diff
@@ -8,138 +8,182 @@
...
+                    <article class="section-heading"><h2>Table of Contents</h2></article>
```
david-yz-liu, line 18: Again, unnecessary `article` element. You don't need to style this heading either, just stick with the default `h2`.

python_ta/reporters/templates/template.html
```diff
@@ -8,138 +8,182 @@
...
+                    <article>
```
david-yz-liu, line 19: Same as above, you don't need the `article` here.

python_ta/reporters/templates/template.html
```diff
@@ -8,138 +8,182 @@
...
+                        <ul>
+                            {% for filename, messages in grouped_messages.items() %}
+                            <li>
+                                <a style="color: DimGray">
+                                    {{filename.split('\\')[-1]}}
```
david-yz-liu, line 24: This splitting won't work on Macs. You can use a Python built-in function from `os` to get the filename.

python_ta/reporters/templates/template.html
```diff
@@ -8,138 +8,182 @@
...
+                <time>{{ date_time }}</time>
+            </header>
+            <aside style="position: sticky">
+                <section id="sidebar">
+                    <article class="section-heading"><h2>Table of Contents</h2></article>
+                    <article>
+                        <ul>
+                            {% for filename, messages in grouped_messages.items() %}
+                            <li>
+                                <a style="color: DimGray">
```
david-yz-liu, line 23: Don't use an `a` tag, use a `span`. Also don't worry about setting a colour for now (the one you chose is too dim), just keep the default.

python_ta/reporters/templates/template.html
```diff
@@ -8,138 +8,182 @@
...
+                                    {{filename.split('\\')[-1]}}
+                                </a>
+                                <ul>
+                                    <li>
+                                        <a style="font-weight: bold">
```
david-yz-liu, line 28: Use `strong` instead of `a`, and don't use inline styles.

python_ta/reporters/templates/template.html
```diff
@@ -8,138 +8,182 @@
...
+                                        <ul>
+                                            {% for message_id in messages[0] %}
+                                            <li style="font-weight: normal">
+                                                {{'{} - {}'.format(message_id, messages[0][message_id][0].symbol)}}
+                                            </li>
+                                            {% endfor %}
+                                        </ul>
+                                    </li>
+                                    <li>
+                                        <a style="font-weight: bold">
```
david-yz-liu, line 40: Similar comments as above

python_ta/reporters/templates/stylesheet.css
```diff
@@ -15,6 +15,30 @@ header {
...
+    width: 300px;
+    display: block;
+    margin-left: auto;
+    margin-right: auto;
+
+    top: 0;
+    position:sticky;
+
+    overflow-y: auto;
+    overscroll-behavior-y: contain;
```
david-yz-liu, line 39: I don't think this is necessary, nor the `position: sticky`. See comments about using css grid below.

david-yz-liu (commented):
@ckatsube a good start. Don't worry about picking colours or other forms on styling in this PR, just focus on getting the grid positioning correct.

ckatsube:
@david-yz-liu thanks for the comments. I'll definitely keep the things like no inline-css in mind in the future.
Also a quick question; why use a class instead of an id for the `sidebar` element? I don't necessary expect that any other element might try using the same class unless we're expecting multiple sidebars.

ckatsube:
Realized I forgot that the aside selectors in the css were pointless since that was deleted. This latest commit should have wrapped up all the commented changes.

python_ta/reporters/templates/stylesheet.css
```diff
@@ -60,6 +74,16 @@ section {
     background: -webkit-linear-gradient(right, rgba(36, 59, 85, 0.67), rgba(20, 30, 48, 0.77));
 }
 
+.section-heading {
```
david-yz-liu, line 77: You can delete this as well

david-yz-liu:
@ckatsube in general it's better to have classes rather than ids so that the rules can be reused in the future. While in this particular application there isn't currently a need for more than one sidebar, that might change; or, we might want to reuse the CSS for a completely different file or project.

ckatsube:
Got it. I'll make sure to keep that in mind going forward, as well

PR #734, opened 2021-07-28
Add dropdown arrows

simonchenwastaken:
## Motivation and Context

Creates a cleaner, simpler, and more intuitive graphical way to show the expand and collapse feature of the sections.
Implements the dropdown arrows from issue #730 

## Your Changes

**Description**:
Changed the expand/collapse buttons into chevron SVG graphic. In addition, added functionality so that the graphic would flip based off of its collapsed state. (Faces up when closed, faces down when content is shown below.)

before:
image
 
after change:
image image
image

## Testing

Tested by running `python_ta.check_all()` to see how the final template ends up rendering with a file.

## Checklist

### DIFF
--- a/python_ta/reporters/templates/stylesheet.css
+++ b/python_ta/reporters/templates/stylesheet.css
@@ -197,6 +197,7 @@ section {
 .slider button {
     background: none;
     border: none;
+    cursor: pointer;
 }
 
 .slider button:hover {
--- a/python_ta/reporters/templates/stylesheet.css
+++ b/python_ta/reporters/templates/stylesheet.css
@@ -177,14 +177,17 @@ section {
 .slider {
     margin-left: auto;
 }
+.slider button {
+    background: none;
+    border: none;
+}
 
-.slider svg:hover {
+.slider button:hover {
     filter: brightness(120%);
 }
 
-svg.collapsed {
-    transform-origin: center;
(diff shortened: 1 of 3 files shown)


python_ta/reporters/templates/stylesheet.css
```diff
@@ -176,21 +178,13 @@ section {
...
-    font-size: 12px;
+.slider svg:hover {
+    filter: brightness(120%);
 }
-.slider button:hover {
-    text-decoration: underline;
+
+svg.collapsed {
+    transform-origin: center;
+    transform: rotateX(180deg) translateY(-6px);
```
david-yz-liu, line 187: Rather than use `rotate`, use `scaleY(-1)` to perform a vertical reflection. You shouldn't need the `translateY` either, if you've set up the viewbox correctly.

python_ta/reporters/templates/template.html
```diff
@@ -23,7 +23,12 @@ <h2> {{ filename }} </h2>
                     <h3 class="category-heading">
                         <span>{{ reporter.code_err_title }}</span>
                         <span class="slider">
-                            <button>Expand/Collapse Section</button>
```
david-yz-liu, line 26: It's better for accessibility to keep the button element and just put the svg inside. This will also allow you to remove the inline `cursor: pointer` rule. In general, you shouldn't use inline styles.

PR #735, opened 2021-08-02
Made reports element scroll independently instead of page

ckatsube:
## Motivation and Context

Addition to issue #730: modified scrolling through the reporters to be local to the area of the screen dedicated to the reports (rather than scrolling through the full page). Essentially the user can always see the header, toc, footer, and reports without having to zoom and the user scrolls through the reports independently.

## Your Changes

**Description**:

1) disabled the `body`'s scrolling to prevent double scrollbars when zooming.
2) fixed grid row heights to pixel-perfect fit the rest of the screen (the approach to this is definitely up for debate as currently it takes the screen height and subtracts the header, footer, and vertical padding/margin heights)
3) moved padding from `footer` to the grid
4) wrapped reporter sections in a div (given a newly introduced class `reporter-container`) because for some reason making applying the changes directly to `main` made the scrollable region tightly hug the bottom of the last report, cutting off the shadow. I was unable to make the main not hug the bottom even with padding and margins and opted to add the wrapping div which made it easier to style.
5) related to 4; made the wrapping div a flexible container in the vertical direction as that automatically centered the reports nicely.

## Testing

Manually created reports to visually look at (images attached at bottom).

## Checklist

Preview of change
image

Current bottom shadow with wrapping div
image

Bottom shadow when trying to make `main` the scrollable
image

### DIFF
--- a/python_ta/reporters/html_reporter.py
+++ b/python_ta/reporters/html_reporter.py
@@ -78,7 +78,8 @@ class HTMLReporter(PythonTaReporter):
 
         # Render the jinja template
         rendered_template = template.render(date_time=dt, reporter=self,
-                                            grouped_messages=grouped_messages)
+                                            grouped_messages=grouped_messages,
+                                            os=os)
 
         # If a filepath was specified, write to the file
         if self.out is not sys.stdout:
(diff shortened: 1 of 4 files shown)


ckatsube:
Reverted point 3 as that was part of the attempt to fix the hugging-shadow issue outlined in point 4.

david-yz-liu:
@ckatsube I'm not a fan of having the vertical scroll on the report sections, it loses the visual of the cards. However, you can keep the horizontal scrolling, but not on the containing card. Instead, put in a separate scroll for each code snippet.

ckatsube:
Uncommented the code's `width: 100%` since the scrollbar was strangely short without it.

Snippet of horizontal code scrolling 
image

Without `width: 100%`
image

PR #736, opened 2021-08-02
added collapse/expand buttons

ckatsube:
## Motivation and Context

Users may create reports for an entire directory of modules, creating an extraordinarily long report as well as table of contents. For easier navigation the table of contents lists should be collapsible.

This PR is a part of Issue #730 

## Your Changes

**Description**:
1) changed current ul style to have no default bullets
2) added a clickable button at each `<li>` which both replaces the removed default bullets and toggles its parent list element.
3) added corresponding css for button and inner svg (used for the button design)
4) added button's event handlers (added both clicking, space key, and enter key handlers)

## Testing

Created reports and observed elements in dev tools when activating events.
Images below

## Checklist

General image
image

Collapsing messages
image

Collapsing files
image

Focus border (when hitting tab)
image

Collapsing with keyboard
image

### DIFF
--- a/python_ta/reporters/templates/script.js
+++ b/python_ta/reporters/templates/script.js
@@ -2,17 +2,7 @@ $("body").on("click", ".slider", function () {
     $(this).parent().next().toggleClass("hide-and-maintain-width");
 });
 
-function toggleClosestCollapsible() {  
+$(".sidebar button").click(function () {  
     $(this).closest(".collapsible").children("ul").toggle();
     $(this).children("svg").toggleClass("collapsed");
-};
-
-$(".sidebar button").click(toggleClosestCollapsible);
-$(".sidebar button").keydown(function(event) {
-    const spacePressed = event.key === " ";
-    const enterPressed = event.key === "Enter";
-    if (spacePressed || enterPressed) {
-        event.preventDefault();
-        $(this).trigger("click");
-    }
-})
+});
--- a/python_ta/reporters/templates/script.js
+++ b/python_ta/reporters/templates/script.js
@@ -1,3 +1,18 @@
 $("body").on("click", ".slider", function () {
     $(this).parent().next().toggleClass("hide-and-maintain-width");
 });
+
+function toggleClosestCollapsible() {  
+    $(this).closest(".collapsible").children("ul").toggle();
+    $(this).children("svg").toggleClass("collapsed");
+};
+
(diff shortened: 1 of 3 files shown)


python_ta/reporters/templates/script.js
```diff
@@ -1,3 +1,18 @@
...
     $(this).parent().next().toggleClass("hide-and-maintain-width");
 });
+
+function toggleClosestCollapsible() {  
+    $(this).closest(".collapsible").children("ul").toggle();
+    $(this).children("svg").toggleClass("collapsed");
+};
+
+$(".sidebar button").click(toggleClosestCollapsible);
+$(".sidebar button").keydown(function(event) {
```
david-yz-liu, line 11: You don't need this. Button elements will automatically respond to Space/Enter keypresses.

python_ta/reporters/templates/stylesheet.css
```diff
@@ -29,6 +29,21 @@ header {
     overscroll-behavior-y: contain;
 }
 
+.sidebar ul {
```
david-yz-liu, line 32: Add to this section to reduce the amount of whitespace before each list element.

david-yz-liu (commented):
Currently, the sections shrink to fit the width of their contents. This is very noticeable now that the table of contents can be collapsed.

You can fix this by adding a `width: 100%` rule to the sections, and also you can add a margin around them to create a bit of a gap.

ckatsube:
Updated PR to resolve merge conflicts

david-yz-liu:
@ckatsube the issue with the collapsing width of the table of contents is still occurring. It looks like you reverted a change, but I'm not sure why. Please fix this.

ckatsube:
The `width: 100%` wasn't doing anything before because the div was already resizing to its grid column. What I didn't realize was that when pulling one of the changes to `master`, the column was changed from `20em` to `min-content`, which made the sidebar start resizing regardless of `width: 100%` or not. Since I want to keep the `min-content` as it makes it easier to make the elements disappear when doing `display: none` in #742, I decided to make the header determine the width column. This does cause the ToC to be a little wider than it really needs to be since its taking up the same width as the header, but imo the extra space isn't too bad and it can be fixed at a later time.

david-yz-liu:
@ckatsube that makes sense, thanks for resolving the issue.

PR #742, opened 2021-08-08
Make side elements disappear when zooming in 

ckatsube:
## Motivation and Context

When zooming in, the leftmost column consisting of the header and table of contents takes space priority over the actual reports. Users with poor eyesight may want to zoom in, but are ultimately unable to see the reports when attempting to.
This PR makes the table of contents and header disappear after zooming in a certain amount, allowing the reports to be the main content on the screen.

(Part of #730)
## Your Changes

**Description**:
Added media selector to make header and table of contents not display when the screen width goes under 38em (approx 600px on standard font size). 

## Testing

Manually created reports and observed behavior when zooming (also changed browser font size).

## Checklist

### DIFF
--- a/python_ta/reporters/templates/stylesheet.css
+++ b/python_ta/reporters/templates/stylesheet.css
@@ -21,7 +21,7 @@ header {
     grid-template-rows: min-content 1fr;
 }
 
-@media screen and (max-width: 38em) {
+@media screen and (max-width: 768px) {
     .sidebar {
         display: none;
     }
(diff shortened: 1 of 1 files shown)


python_ta/reporters/templates/stylesheet.css
```diff
@@ -21,6 +21,15 @@ header {
     grid-template-rows: min-content 1fr;
 }
 
+@media screen and (max-width: 38em) {
```
david-yz-liu, line 24: I like the idea a lot, but I'd rather have a fixed pixel width than use `em` here. Some common breakpoints are found [here](https://tailwindcss.com/docs/breakpoints), I think `768px` might be a good choice for this section.

PR #760, opened 2021-09-15
Report last line and last column of code snippet

mishaschwartz:
## Motivation and Context

The JSONReporter reports the line and column where an error starts in the text being analyzed but not the line and column where it ends. 

Both start and end positions are useful when using the output of the JSONReporter to identify problematic code.

## Your Changes

- calculated the values of line and column endings in a new method (JSONReporter._snippet_endings)
- Added a call to JSONReporter._snippet_endings in JSONReporter.display_messages to add this info to the reported json string.

**Description**:

## Checklist

### DIFF
--- a/python_ta/reporters/core.py
+++ b/python_ta/reporters/core.py
@@ -14,7 +14,7 @@ from pylint.reporters.ureports.nodes import BaseLayout
 from .node_printers import LineType, render_message
 
 # Extension of Pylint's Message class to incorporate astroid node and source code snippet
-NewMessage = namedtuple("NewMessage", Message._fields + ("node", "snippet"))
+NewMessage = namedtuple("NewMessage", Message._fields + ("node", "snippet", "line_end", "column_end"))
 
 # Messages without a source code line to highlight
 NO_SNIPPET = {"invalid-name"}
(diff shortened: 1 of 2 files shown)


david-yz-liu:
@mishaschwartz this is a really good catch, thanks! In terms of fixing the issue though, I'd prefer a simpler approach that specifically targets the `JSONReporter`. The key thing is that the `node` object in the message already contains the `end_lineno` and `end_col_offset` attribute, so you can change the

https://github.com/pyta-uoft/pyta/blob/2b622961b8a52f29002bfa1e49fdb6fb9e586fc4/python_ta/reporters/json_reporter.py#L30-L34

to something like

```python
        output = []
        for k, msgs in self.messages.items():
            output.append(
                {"filename": k,
                 "msgs": [{**msg._replace(node=None)._asdict(), 'line_end': msg.node.end_lineno, 'column_end': msg.node.end_col_offset} for msg in msgs]}
            )
```

with a special case for when `msg.node is None` (for certain message types with no node), and styling/refactoring as appropriate.

mishaschwartz:
Sure thing. It makes sense to move the code into the JSONReporter class if that's the only place this info will be needed.

PR #763, opened 2021-09-19
R1734 use list literal docs #1734

younwoochoi:
## Motivation and Context

Created documentation and an example for the Pylint checker R1734 use list literal.

## Your Changes

**Description**:

Added R1734 example file to `examples/pylint`

Added R1734 to `index.md`

**Type of change** (select all that apply):

- [x] Documentation update (change that modifies or updates documentation only)

## Checklist

- [x] I have performed a self-review of my own code.
- [x] I have verified that the CI tests have passed. 
- [x] I have reviewed the test coverage changes reported on Coveralls. 
- [x] I have added tests for my changes.

PR #764, opened 2021-09-19
R1735 use dict literal docs #1735

younwoochoi:
## Motivation and Context

Created documentation and an example for the Pylint checker R1735 use dict literal.

## Your Changes

Added R1735 example file to `examples/pylint`

Added R1735 to `index.md`

**Description**:

## Checklist

### DIFF
--- a/examples/pylint/R1735_use_dict_literal.py
+++ b/examples/pylint/R1735_use_dict_literal.py
@@ -1,3 +1,3 @@
-student_dict = dict() # Erro on this line.
+student_dict = dict() # Error on this line.
 
 student_dict["123456789"] = "John Doe"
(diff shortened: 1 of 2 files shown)


examples/pylint/R1735_use_dict_literal.py
```diff
@@ -0,0 +1,3 @@
+student_dict = dict() # Error on this line.
+
+student_dict["123456789"] = "John Doe"
```
david-yz-liu, line 3: I'd prefer showing an example using a loop, similar to what you had for R1734 for lists. Otherwise, the code is a bit simple, and in fact this example is misleading since our approved style would be `student_dict = {"123456789": "John Doe"}`, which defeats the purpose of the check.

david-yz-liu (commented):
@younwoochoi again good job, please look at my comments in #763 and apply them here as well. I also left one additional inline comment below.

david-yz-liu:
Thanks, @younwoochoi!

PR #765, opened 2021-09-19
Adding example and documentation for W1310 (Format string without interpolation)

rishivijayv:
## Changes
**Description**:
Adding a short example which contains the W1310 error along with supporting documentation which explains the error and contains a code sample that resolves W1310. The example can be found in the *Miscellaneous* section of `index.md`, where documentation for some other format string related pylint errors are located as well.

**Type of change**:

## Testing
- Manually verified that the W1310 error occurs in the example file documenting this error through pylint.
- Ran `python -m pytest tests/test_examples.py` after creating the W1310 example file, which succeeded as well.
- Verified through pylint that the corrected example mentioned in documentation does not include the W1310 error.
- Ran `make html` in the `docs` directory, verifying that the documentation is created for W1310.

## Checklist

### DIFF
--- a/docs/checkers/index.md
+++ b/docs/checkers/index.md
@@ -1829,6 +1829,30 @@ s = "{} and {}".format("first", "second")
 
 **See also**: [E1120](#E1120)
 
+### F-string without interpolation (W1309) [](#W1309)
+
+This error occurs when there are no interpolation variables present in an f-string. This might indicate that there is
+either a bug in the code (that is, there should be an interpolation variable in the f-string) or the f-string can be a
+normal string.
+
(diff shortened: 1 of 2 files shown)


david-yz-liu (commented):
@rishivijayv this is great! I just realized there's no documentation for a related check, W1309 (even though there's an example file for it).

Can you please add a new commit to this branch with documentation for W1309, and then use "See also" references to link between these two issues? (We use "See also" elsewhere in the documentation for this purpose.)

rishivijayv:
Thanks for the feedback @david-yz-liu ! Just added documentation for W1309 as well, while linking W1309 and W1310 using "See also"

PR #769, opened 2021-09-23
Updated `tests/test_examples` to use the actual Pylint module

ckatsube:
## Motivation and Context

Previously, `test_examples.py` would not access `python_ta` and throw an error until externally installing `python_ta` due to the usage of calling Pylint via a subprocess. This PR replaces the usage of the `subprocess` module with the actual `pylint` module, which resolves the issue.

## Your Changes

Replaced generating reports using `subprocess` by using `pylint.lint.Run`. It didn't seem like there was an option to provide a stdout string buffer so I manually replaced the system's stdout to capture the pylint report (and properly reverted the stdout).
Alternative notes:
* It might be possible to use a custom reporter as a buffer, but the approach didn't seem appealing to me.
* I considered using pylint.epylint but the method didn't seem to allow for passing in multiple files, which was the whole point of #713.

**Description**:

## Testing

Ran the test suite with and without the externally installed `python_ta` (haven't tried it on a clean installation yet since I currently have some untracked files I want to keep).

## Checklist

### DIFF
--- a/python_ta/patches/messages.py
+++ b/python_ta/patches/messages.py
@@ -16,7 +16,9 @@ def patch_messages():
     ):
         old_add_message(self, msg_id, line, node, args, confidence, col_offset)
         msg_info = self.msgs_store.get_message_definitions(msg_id)[0]
-        self.reporter.handle_node(msg_info, node)
+        # guard to allow for backward compatability with Pylint's reporters
+        if hasattr(self.reporter, "handle_node"):
+            self.reporter.handle_node(msg_info, node)
 
     MessagesHandlerMixIn.add_message = new_add_message
 
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -6,6 +6,9 @@ import re
 import pytest
 import json
 import itertools
+from pylint import lint
+from io import StringIO
+import sys
 
 
 _EXAMPLES_PATH = "examples/pylint/"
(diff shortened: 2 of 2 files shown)


ckatsube:
Hm I'll review why the tests aren't passing. I just tried a clean install on another computer and it worked fine (including both 3.8 and 3.9).

ckatsube:
I believe the issue has to do with a side effect of PyTA's where the linter's `add_message` method is replaced with the `new_add_message` from `python_ta/patches/messages`. The test is failing because `test_check` uses PyTA's `check_all`, which produces the side effect and then proceeds to test `test_examples` which uses Pylint's innate JSONReporter. The side effect appears to deny backward compatibility with Pylint's default reporters, as the linter's reporter requires an `handle_node` method, which is from PyTA. I see that I came across the issue before in #713 (the specific commit linked below), but opted to use the subprocess instead of the pylint module without realizing that it would cause the bug this issue is meant to fix.

Suggestions on how to proceed from here? I was thinking:
1) reset Pylint's linter to its default state after finishing running `check_all`
2) monkey patch the JSONReporter to either have a `handle_node` which does nothing, or give re-replace the `add_message` with its original one.

Aforementioned commit:
https://github.com/pyta-uoft/pyta/pull/713/commits/560cfd82b582cd375bdf48d444311cf8d9c90eba

david-yz-liu:
@ckatsube excellent investigation! I think you can fix this issue by forcing a reload of the pylint modules, e.g. with https://docs.python.org/3/library/importlib.html#importlib.reload.

PR #773, opened 2021-09-26
Dynamic postconditions check

rishivijayv:
## Motivation and Context
Postconditions provide an effective method to ensure that a function behaves as expected, and are a useful way for new programmers to be sure that a piece of code they have written is correct. This change allows programmers to add a list of checks that should be verified just after the function has executed - allowing them to specify checks on the function's return values, parameters, or other variables in the function's scope.

## Changes

**Description**:
For the most part I have reused the functionality present for checking function preconditions. One major difference has been an introduction of a special variable name for the function's return value, which I have called `$return_value`. Using the `$` symbol as the first character ensures that the variable name does not conflict with any other variable that might have been declared by users. 

When evaluating a postcondition check containing the `$return_value` variable - I first replace the assertion string containing this variable with the actual return value of the function. This is because assertions are evaluated in the `_check_assertions` function (present in the `__init__.py` file of `python_ta.contracts`) using the `eval` function. Thus, using the illegal variable name `$return_value` in this expression caused an exception, which is why the replacement was needed.

**Type of change**:

## Testing
I have added added 2 unit tests for each of the 3 categories below. The 2 tests check an erroneous and a correct implementation of a function - the former causes the postcondition check to fail, while the latter results in the check passing:

- Expression containing the `$return_value` variable
- Expression containing function's parameters
- Expressions using functions defined by the user that should be in scope for the function

These unit tests passed, along with the unit tests already present.

## Checklist

### DIFF
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -327,7 +327,7 @@ def _check_assertions(
 
 
 def parse_assertions(obj: Any, parse_token: str = "Precondition") -> List[str]:
-    """Return a list of preconditions/representation invariants parsed from the given entity's docstring.
+    """Return a list of preconditions/postconditions/representation invariants parsed from the given entity's docstring.
 
     Uses parse_token to determine what to look for. parse_token defaults to Precondition.
 
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -12,6 +12,7 @@ Set to True to display debugging messages when checking contracts.
 """
 
 _DEFAULT_MAX_VALUE_LENGTH = 30
+FUNCTION_RETURN_VALUE = "$return_value"
 
 
 class PyTAContractError(Exception):
@@ -183,6 +184,15 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
                 f"return type annotation {_display_annotation(return_type)}"
             )
 
+    # Check function postconditions
+    postconditions = parse_assertions(wrapped, parse_token="Postcondition")
+    _check_assertions(
+        wrapped,
+        {**function_locals, FUNCTION_RETURN_VALUE: r},
+        postconditions,
+        condition_type="postcondition",
+    )
+
     return r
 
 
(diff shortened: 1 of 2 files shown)


tests/test_contracts.py
```diff
@@ -250,6 +250,244 @@ def unary(arg: List[str]) -> None:
...
+    """
+    Incorrectly returns a number that is not twice the number passed as the argument
+
+    Postcondition: $return_value == num * 2
+    """
+    return (num * 2) + 1
+
+
+def test_get_double_valid() -> None:
+    """Calling the valid implementation of _get_double to get double the number"""
```
david-yz-liu, line 275: For all of your test docstrings:

1. Start with the phrase "Test that"
2. End the description with a period.
3. When the description spans more than one line, put the ending `"""` on a newline (not at the end of the previous line). This one really should be enforced by black (the formatter), but I think maybe it's not being run on the test files right now.

python_ta/contracts/__init__.py
```diff
@@ -183,6 +184,24 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
...
+        {**wrapped.__globals__, **function_locals}
+    )
+    postconditions = [
+        assertion.replace(FUNCTION_RETURN_VALUE, legal_return_val_var_name)
+        for assertion in pre_replacement_postconditions
+    ]
+
+    _check_assertions(
+        wrapped,
+        {**function_locals, legal_return_val_var_name: r},
```
david-yz-liu, line 199: The overall approach is great, but in terms of the implementation:

1. I think it's better to do the replacing inside `_check_assertions`. So rather than changing the `function_locals` and passing the identifier name, keep the original locals and pass the return value.
2. This way, in the error message, you can separate the arguments and the return value---this is important because the return value is not technically an "argument", and we want to make sure that PythonTA respects proper terminology. I could see an error message like `with arguments {...} and return value ...`
3. As a bonus, you'll have access to the original assertion in `_check_assertions`, so there won't be a need to do the reversing replacement.

python_ta/contracts/__init__.py
```diff
@@ -275,30 +294,55 @@ def _check_invariants(instance, klass: type, global_scope: dict) -> None:
...
         else:
             if not check:
                 arg_string = ", ".join(
-                    f"{k}: {_display_value(v)}" for k, v in function_locals.items()
+                    f"{k if legal_return_val_var_name != k else FUNCTION_RETURN_VALUE}: {_display_value(v)}"
+                    for k, v in function_locals.items()
                 )
                 arg_string = "{" + arg_string + "}"
+
+                # Replacing return_val_var_name (if used) with FUNCTION_RETURN_VALUE to display message in terms of
```
david-yz-liu, line 333: In general for comments, use imperative tense (so `Replacing` -> `Replace`

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Fix HTML report to display file even when no errors are found.
 - Fix pylint cache directory creation (backport of change from pylint 2.11)
 
+### Enhancements
```
david-yz-liu, line 17: Good, but you should actually add this to the "Unreleased" section above.

david-yz-liu:
@rishivijayv I re-ran the tests and they passed---not sure what the issue was, but let's just keep an eye out for it in the future.

PR #800, opened 2022-02-08
Add compile to contracts

peelstnac:
## Motivation and Context
Currently, checking preconditions, postconditions, and representation invariants requires 
1. parsing the assertions from the function/class docstring
2. calling eval on these assertions

every time a function is called/class instance is modified.

We can instead only parse these assertions once, `compile` them, and store the compiled object as an attribute. 

## Your Changes

**Description**:
Assertions are parsed only once, compiled, and stored as an attribute of the function/class it belongs to (note: we ignore any assertions that are not compilable ie. English statements).

## Testing

Code changes passed all tests. I've added two tests for postconditions in `test_class_contracts.py` because there weren't any before.

## Checklist


david-yz-liu:
@peelstnac great questions! Handling postconditions is definitely a bit tricky, you can see the PR for it at #773.

To specify a postcondition, users use a special identifier [`$return_value`](https://github.com/pyta-uoft/pyta/blob/master/python_ta/contracts/__init__.py#L15) to refer to the function's return value. A special identifier is necessary because we can't assume an existing variable name refers to the return value.

However, to evaluate the postcondition as a Python expression, we have to replace `$return_value` with another name, because `$return_value` is not a valid Python name (starts with a $). So we have to dynamically generate a new variable name *that does not conflict with any of the variables already in scope for the function*---that's what [this part](https://github.com/pyta-uoft/pyta/blob/master/python_ta/contracts/__init__.py#L327-L329) does.

To answer your second question, your current setup is very close to working, but just to clarify:

1. In the first phase of *parsing assertions*, for postconditions you should: (i) replace the `$return_value` with a valid identifier, (ii) compile the resulting expression, and (iii) store _both the compiled expression and the identifier for the return value_. You don't need to store the `return_val_dict` at this point.
2. Then when *checking assertions*, this is where you have access to the return value for the current function call. You should then (i) construct the `return_val_dict` using the identifier you saved and the return value, and then (ii) call `eval` on the compiled expression, passing in the `return_val_dict` along with the globals and locals.

So again you're really close, just need to separate the responsibilities of the compile vs. eval phases for handling return values.

peelstnac:
@david-yz-liu Thanks for the hints!  

- The correct return value is now used to check postconditions.

- Because bounded methods cannot have attributes added to them, I've decided to circumvent this issue with [this block of code](https://github.com/peelstnac/pyta/blob/d4d5815ba8ca69134cc04ba8ac8e22f8a84dcbd6/python_ta/contracts/__init__.py#L183-L193). In that block of code, I'm using a [dummy class](https://github.com/peelstnac/pyta/blob/d4d5815ba8ca69134cc04ba8ac8e22f8a84dcbd6/python_ta/contracts/__init__.py#L22-L23) which I can add attributes to. 

- Currently, test_class_contracts.py only has one test for preconditions of a class function and none for postconditions. My plan is to add a few more to make sure my code is correct.

david-yz-liu:
@peelstnac there's a simpler way to handle the bound methods. From a bound method like `a.foo`, you can access the underlying function object with `a.foo.__func__`, and can set attributes for that function in the same way as top-level functions.

david-yz-liu (commented):
@peelstnac great work! I've just left some comments to improve, but the overall approach is spot on.

peelstnac:
@david-yz-liu Thanks for the feedback! I've made all the suggested changes, but I'm not sure if the module docstring I put is what you're looking for.

python_ta/contracts/__init__.py
```diff
@@ -94,14 +107,24 @@ def add_class_invariants(klass: type) -> None:
         return
 
     # Update representation invariants from this class' docstring and those of its superclasses.
-    rep_invariants = set()
+    rep_invariants: List[Tuple[str, object]] = []
```
david-yz-liu, line 110: `object` is too broad a type here. Check the exact type in the Python console.

python_ta/contracts/__init__.py
```diff
@@ -1,7 +1,20 @@
+"""This module provides the functionality for PythonTA contracts.
+
+Representation invariants, preconditions, and postconditions are parsed, compiled, and stored.
+Below are some notes on how they are stored.
+    - Representation invariants are stored in the instance attribute __representation_invariants__
```
david-yz-liu, line 5: Don't use the word "instance" here. You can just say "...stored in a class attribute as a list ..."

python_ta/contracts/__init__.py
```diff
@@ -1,7 +1,20 @@
...
+    of the class instance as a list [(assertion, compiled)].
+    - Preconditions are stored in an attribute __preconditions__ of the function as a list
+    [(assertion, compiled)].
+    - Postconditions are stored in an attribute __postconditions__ of the function as a list
+    [(assertion, compiled, return_val_var_name)].
+    - For bounded methods, preconditions and postconditions are stored as an attribute of the method's
```
david-yz-liu, line 11: You can omit this point, it's implied by the previous two.

david-yz-liu (commented):
Good job, the module docstring is quite good. I just left a few more comments, but I think this is getting close

david-yz-liu (approved):

PR #803, opened 2022-02-28
Issue 799

### LINKED ISSUE #799: check_all_contracts should only modify functions and classes defined in the current module
Report from instructor: running the following file raises an error.

```python
blah = int

def divide(x: blah, y: blah) -> int:
    """Return x // y.

    """
    return x // y


if __name__ == '__main__':
    from python_ta.contracts import check_all_contracts
    check_all_contracts()

    divide(10, 'x') 
```

The cause is `check_all_contracts` attempts to add representation invariants to `blah` (a variable in the module scope), which refers to a built-in type (`int`) that can't be mutated. We should modify `check-all_contracts` to skip types that aren't defined in the current module (unless that module is specified as an argument to `check_all_contracts`).

peelstnac:
## Motivation and Context

Fixes #799.

## Your Changes

**Description**:

Skip contract checks for functions and classes which are not defined in a module whose name is passed as an argument to `check_all_contracts`.

## Testing

Added `tests/test_check_all_contracts.py` to test the code which produces this bug.

## Checklist


peelstnac:
@david-yz-liu I've moved the test for check_all_contracts into a folder within the tests folder. Please let me know if the naming is alright.

Thanks!

tests/test_contracts.py
```diff
@@ -503,3 +503,13 @@ def unary(arg: list[str]) -> None:
 
     with pytest.raises(AssertionError):
         unary(dict)
+
+
+def test_check_all_contracts_module_names_argument() -> None:
+    """Test that checks for classes and functions not declared as parameters to check_all_contracts
+    are skipped.
+    """
+    from tests.test_check_all_contracts.test_ignore_modules_not_in_arg import run
```
david-yz-liu, line 512: One thing I forgot to mention is that pytest will recursively look for folders and files that contain `test_` (or something like that), so we don't want to include `test_` in names of folders or files that contain input data for other tests.

Naming is a bit tricky here. You could do something like `tests/fixtures/contracts/modules_not_in_arg.py`. ("Fixtures" is a term for objects used to set up a test context, which can include external files.)

david-yz-liu (approved):

PR #810, opened 2022-05-08
Merge config

peelstnac:
## Motivation and Context

In many programs, when users provide their own configuration file, their configurations are merged with the default ones. We do the same for custom error messages.

## Your Changes

**Description**:

If user specifies their own messages config file, their custom error messages are merged into the default custom error messages.

## Testing

Manually tested by fielding a messages config file with different and new error messages.

## Checklist


python_ta/__init__.py
```diff
@@ -166,6 +166,24 @@ def _load_config(linter, config_location):
...
 
 
+def _load_messages_config(path: str, def_path: str) -> dict:
+    """Given path (potentially) specified by user and default def_path
+    of messages config file, merge the config files."""
+    if def_path == path:
+        return toml.load(path)
+    merge_into = toml.load(def_path)
+    merge_from = toml.load(path)
+    for i in merge_from:
```
david-yz-liu, line 176: This nesting is indicating some kind of assumed structure for the config file. Use meaningful variable names to convey what each level of structure means.

python_ta/__init__.py
```diff
@@ -166,6 +166,24 @@ def _load_config(linter, config_location):
     linter.load_config_file()
 
 
+def _load_messages_config(path: str, def_path: str) -> dict:
```
david-yz-liu, line 169: Since `def` has a different meaning in Python (and programming in general), it's not a good short form for `default`. Please change `def_path` to `default_path` both here and in `messages_config_def_path`.

david-yz-liu (commented):
@peelstnac this is good, but you should add tests for your changes. For example, provide a non-default configuration file, and check that (1) the messages in that file override the ones from the default file, and (2) the messages from the default file that _aren't_ overridden appear as expected.

python_ta/__init__.py
```diff
@@ -166,6 +166,24 @@ def _load_config(linter, config_location):
     linter.load_config_file()
 
 
+def _load_messages_config(path: str, default_path: str) -> dict:
+    """Given path (potentially) specified by user and default default_path
+    of messages config file, merge the config files."""
+    merge_into = toml.load(default_path)
+    merge_from = toml.load(path)
```
david-yz-liu, line 173: @peelstnac I thought about this a bit more carefully and have a couple more things I'd like you to update here. (And, investigating a runtime issue that @harshkpatel reported.)

1. If `path == default_path`, we should just return the default configuration, no need to reload the same file and merge.
2. If `path` cannot be found (because of an error in the configuration file), please output a warning message (similar to other PythonTA warnings) and return the default config. Right now `toml.load` will raise an error and halt execution altogether.
3. In the default `.pylintrc` file, please *comment out* the config value for the messages config (so that no value is used by default), and change the example value to just `messages.toml` or something similar, without the parent folders. The relative path structure is a bit misleading.

david-yz-liu (approved):

PR #823, opened 2022-06-03
Type annotation checker

peelstnac:
## Motivation and Context

Sometimes students make the mistake of assigning types instead of annotating them. 

```py
def add_two_numbers(x=int, y=int) -> int: 
    # type is assigned instead of annotated here, 
    # should be def add_two_numbers(x: int, y: int) -> int
    return x + y


class MyDataType:
    # type is assigned instead of annotated here
    x = int
    y = int
```

## Your Changes

**Description**:

We extend the `type_annotation_checker` to explicitly warn the student they are assigning types instead of annotating them.

## Testing

This change is tested with tests in `tests/test_custom_checkers/test_type_annotation_checker.py`. We run the checker on a new example we added `examples/e9995_type_is_assigned.py` and match its output.

## Checklist


david-yz-liu (commented):
@peelstnac great work. Here are a couple of overall comments for you to work on:

1. The error should _not_ trigger if there is already a type annotation for the parameter/attribute. So for example, `x: type = int` shouldn't trigger the error, but also `x: int = int` shouldn't either, even though the default value doesn't match the type annotation (this is checked by a different checker).
2. Right now the check is too specialized; it doesn't handle subscript expressions like `x = list[int]`, or types from somewhere other than `builtins`. Instead of checking the literal name, try calling `.inferred()` on the node; you'll get a list of possible values that the node evaluates to. Access the first item in the list and check whether it's a `type` or not. (If the list is empty, don't trigger the error---this means the inference failed, and so we should be conservative and not report errors in this case.)
3. Make sure to add additional tests for all of the functionality described above.
4. Add the checker to the HTML documentation, and to the Changelog (since it's a user-facing change).

peelstnac:
It seems like what I really wanted was just to check if inferred node is a class definition. This should cover all bases 🎉.

david-yz-liu (approved):

PR #825, opened 2022-06-14
updated documentation and usage for current version of new accumulation table feature

harshkpatel:
This change is to add a new feature in PyTA that allows students to better debug for loops that deal with accumulating a certain object such as a list. This will allow future students to see each loop iteration and its respective value formatted nicely in a table. So far, there has been a new data structure created `AccumulationTable` that is used to print the desired table output. The data structure was made and turned into a context manager to simplify the lines of code the students would have to write in their code. Furthermore, the documentation was updated, which includes sample usage.

## Motivation and Context

This pull request was created to provide future students with a new way to debug. A part of CSC110/111 was going through the loop accumulation tables manually by hand to understand how a for loop functioned and by providing students with this new debugging feature, it will not only strengthen their understanding but it will also help them debug any accumulation issues they may have in loops they write for projects, assignments and other tasks in CSC110/111.

## Your Changes

**Description**: I created a new data structure, turned it into a context manager and created documentation to illustrate an Accumulation Table. I then went further with the context manager and used the `inspect` module to dynamically access a frame in the call stack while the code is running in order to obtain the name of the loop variable as well as the values of the accumulator variables.

## Testing

I tested this data structure on 3 different accumulation function each with a different quantity of accumulation variables and received the desired table output for each case.

I have added comments to the file `print_table` about two issues I ran into. I am unsure about the first one that is dealing with a duplicate loop variable name in the same scope, however the issue with the class attributes, I tried fixing something by looking at if the function is a class function vs a stand alone function, but I felt that there would be other issues in the case that the class function had its own accumulator variables in the scope of the function instead of class attributes.

## Checklist


sample_usage/print_table.py
```diff
@@ -0,0 +1,20 @@
+from python_ta.debug import AccumulationTable
+
+
+def my_func(numbers: list) -> None:
+    """Print the Accumulation Table to show the tracking of the respective variables."""
+    sum_so_far = 0
+    list_so_far = []
+    avg_so_far = "N/A"
+    with AccumulationTable.AccumulationTable(
```
david-yz-liu, line 9: @harshkpatel I think this is a typo (two `AccumulationTable`s)

python_ta/debug/AccumulationTable.py
```diff
@@ -0,0 +1,72 @@
...
+    _num_iterations: int
+    _accumulation_names: list
+    _loop_accumulation_values: dict
+    _loop_var_name: str
+    _loop_variable: list
+
+    def __init__(self, accumulation_names: list):
+        self._accumulation_names = accumulation_names
+        self._num_iterations = 0
+        self._loop_variable = ["N/A"]
```
david-yz-liu, line 17: it's okay if there's no loop variables specified (you can make this an empty list)

python_ta/debug/AccumulationTable.py
```diff
@@ -0,0 +1,72 @@
...
+        for i in range(len(self._accumulation_names)):
+            self._loop_accumulation_values[self._accumulation_names[i]].append(val[i])
+        self._num_iterations += 1
+
+    def _convert_table(self) -> dict:
+        table_dict = {
+            "iterations": list(range(self._num_iterations + 1)),
+            "loop variable (" + self._loop_var_name + ")": self._loop_variable,
+        }
+        for key, pair in self._loop_accumulation_values.items():
```
david-yz-liu, line 30: You can actually just add `**self._loop_accumulation_values` to the previous dict literal expression to unpack it and get all of the key-value pairs added to the dictionary.

python_ta/debug/AccumulationTable.py
```diff
@@ -0,0 +1,72 @@
...
+        return table_dict
+
+    def _tabulate_data(self) -> None:
+        print(
+            tabulate.tabulate(
+                self._convert_table(), headers="keys", colalign=("left", "left", "left")
+            )
+        )
+
+    def _loop_string(self, nconverted: str) -> str:
```
david-yz-liu, line 41: You should improve this function in two ways:

1. Make use of the *current line number* to determine where in the source code to start
2. Check the *indentation of all non-blank lines* to determine where in the source code to *end* (remember that you only care about the block of code that's *indented within* the `with` statement).

python_ta/debug/AccumulationTable.py
```diff
@@ -0,0 +1,72 @@
...
+                break
+
+        cut_s = list_str[temp_index:]
+        curr_str = ""
+        for elem in cut_s:
+            curr_str += elem + "\n"
+
+        return curr_str
+
+    def _add_keys(self, ast_node, prev) -> None:
```
david-yz-liu, line 56: Add type annotations here, and pick better variable names.

python_ta/debug/AccumulationTable.py
```diff
@@ -0,0 +1,72 @@
...
+        self._loop_var_name = ast_node.target.name
+        self._loop_accumulation_values = {}
+        for name in self._accumulation_names:
+            if name in prev.f_locals:
```
david-yz-liu, line 60: If this check fails for some variables but not others, the stored lists will have different lengths, which doesn't make sense for printing the table.

I think it's okay to just raise an error if this check fails---assume that the loop variable is local to the enclosing function.

python_ta/debug/AccumulationTable.py
```diff
@@ -0,0 +1,72 @@
...
+                self._loop_accumulation_values[name] = [prev.f_locals[name]]
+
+    def __enter__(self):
```
david-yz-liu, line 63: Include a return type here

python_ta/debug/AccumulationTable.py
```diff
@@ -0,0 +1,72 @@
...
+        prev = inspect.getouterframes(inspect.currentframe())[1].frame
```
david-yz-liu, line 64: overall you should pick much better variable names in this function

python_ta/debug/AccumulationTable.py
harshkpatel: For this specific thing it throws off the formatting of the table: it makes it so the loop variable section is now missing the zeroth iteration, so either I can leave it as it is or I can use '\n' instead

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,115 @@
...
+
+    """
+
+    _num_iterations: int
+    _accumulation_names: list[str]
+    _loop_accumulation_values: dict[str, list]
+    _loop_var_name: str
+    _loop_var_val: list[Union[int, str]]
+
+    def __init__(self, accumulation_names: list) -> None:
```
david-yz-liu, line 30: specify the contained type in the `list` (and check the other type annotations in this file)

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,115 @@
...
+        _loop_var_name: the name of the loop variable
+        _loop_var_val: the value of the loop variable during each iteration
+
+    """
+
+    _num_iterations: int
+    _accumulation_names: list[str]
+    _loop_accumulation_values: dict[str, list]
+    _loop_var_name: str
+    _loop_var_val: list[Union[int, str]]
```
david-yz-liu, line 28: Oh, this can actually just be `list` because the loop variable can have any type (remember that you can iterate over arbitrary collections)

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __init__(self, accumulation_names: list) -> None:
+        self._accumulation_names = accumulation_names
+        self._num_iterations = 0
+        self._loop_var_val = ["N/A"]
```
david-yz-liu, line 33: Regarding your comment, you should move the initialization into the `__enter__` method, because that's where the accumulator values get their initial values. This keeps the data structures parallel to each other throughout the method calls.

Also, you should initialize all instance attributes here, so you need to add the missing ones here.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,115 @@
...
+        for name in self._accumulation_names:
+            if name in func_frame.f_locals:
+                self._loop_accumulation_values[name] = [func_frame.f_locals[name]]
+            else:
+                raise NameError
+
+    def __enter__(self) -> AccumulationTable:
+        func_frame = inspect.getouterframes(inspect.currentframe())[1].frame
+        func_string = inspect.cleandoc(inspect.getsource(func_frame))
+        lst_from_with_stmt = func_string.splitlines()[
```
david-yz-liu, line 101: Don't call `splitlines` more than once (a bit inefficient), save the value in a variable here

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def _tabulate_data(self) -> None:
+        """Print the values of the accumulator and loop variables into a table"""
+        print(
+            tabulate.tabulate(
+                self._create_iteration_dict(), headers="keys", colalign=("left", "left", "left")
+            )
+        )
+
+    def _no_of_indents(self, with_line: str) -> int:
```
david-yz-liu, line 61: This is definitely the right idea, but I think needs to be a bit more general. The issue is that Python allows different sizes for indentation (e.g. 2 spaces instead of 4) and allows tabs (`'\t'`) as well.

So you can improve your approach to just count the number of whitespace characters rather than the number of indents. The same logic you're using in `__enter__` will work, you just need to compare the the number of whitespace characters rather than the number of indentations

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,115 @@
...
+            if name in func_frame.f_locals:
+                self._loop_accumulation_values[name] = [func_frame.f_locals[name]]
+            else:
+                raise NameError
+
+    def __enter__(self) -> AccumulationTable:
+        func_frame = inspect.getouterframes(inspect.currentframe())[1].frame
+        func_string = inspect.cleandoc(inspect.getsource(func_frame))
+        lst_from_with_stmt = func_string.splitlines()[
+            inspect.getlineno(func_frame) - func_frame.f_code.co_firstlineno + 1 :
```
david-yz-liu, line 102: Similarly save this index (or this index - 1) in a variable and give it a good name

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,115 @@
...
+        """Convert the list of lines that are strings from the beginning of the for loop
+        to the end of the for loop"""
+        endpoint = len(full_string)
+        for i in range(len(full_string)):
+            if full_string[i] != "" and (no_indents + 1) * "    " not in full_string[i]:
+                endpoint = i
+                break
+
+        end_loop = full_string[:endpoint]
+        loop_str = ""
```
david-yz-liu, line 82: You can simplify this part by using the `str.join` method

python_ta/debug/accumulation_table.py
david-yz-liu: @harshkpatel I think you missed this comment

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,133 @@
...
+            for accumulator in self._accumulation_names:
+                if accumulator in local_vars:
+                    curr_vals.append(local_vars[accumulator])
+            if self._loop_var_name in local_vars:
+                self._curr_values = [curr_vals, local_vars[self._loop_var_name]]
+                self._add_iteration(curr_vals, local_vars[self._loop_var_name])
+
+    def __enter__(self) -> AccumulationTable:
+        self._loop_var_val.append("N/A")
+        func_frame = inspect.getouterframes(inspect.currentframe())[1].frame
```
david-yz-liu, line 115: The code from here to line 122 should be moved into a (well-documented) helper function

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,133 @@
...
+    _loop_var_val: list
+    _curr_lineno: int
+    _curr_values: list
+
+    def __init__(self, accumulation_names: list) -> None:
+        self._num_iterations = 0
+        self._accumulation_names = accumulation_names
+        self._loop_accumulation_values = {}
+        self._loop_var_name = ""
+        self._loop_var_val = []
```
david-yz-liu, line 39: Don't use separate attributes for the accumulation variable values and the loop variable values. Keep one dictionary that stores all values.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,133 @@
...
+    Class used as a form of print debugging to analyze different loop and
+    accumulation variables during each iteration in a for loop
+
+    Private instance attributes:
+        _num_iterations: the number of for loop iterations
+        _accumulation_names: the variable names of the different accumulation variables
+        _loop_accumulation_values: a mapping between the accumulation variables and their value during each iteration
+        _loop_var_name: the name of the loop variable
+        _loop_var_val: the value of the loop variable during each iteration
+        _curr_lineno: the current line number of the loop
```
david-yz-liu, line 21: This attribute name and description need to be updated

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+from __future__ import annotations
+
+import inspect
+import sys
+
+import astroid
+import tabulate
+
+
+def no_of_whitespaces(with_line: str) -> int:
```
david-yz-liu, line 14: Rename to `num_whitespaces`. (In general "no" is not a good short form for "number", since "no" also means... "no")

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+    """Return the number of whitespaces that are in the line of code
```
david-yz-liu, line 15: This description should be more generic, and not assume that the line of code contains a with statement (or in fact is a line of code at all). Something like "number of whitespace at the start of <param name>" is better. Also choose a better parameter name.

And finally, all docstring statements should end with a period (make sure to do this throughout).

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+    containing the with statement
+    """
+    blank_chars = 0
+    for char in with_line:
+        if char == " ":
```
david-yz-liu, line 20: Use `str.isspace` (to allow for tabs)

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+            blank_chars += 1
+        else:
+            break
+
+    return blank_chars
+
+
+def loop_string(full_string: list[str], no_whitespaces: int) -> str:
```
david-yz-liu, line 28: Similar to above comment, rename this function and reword the description to capture the full generality of the behaviour. This function returns the lines up to the first line...

Rename `full_string` (to something like `lines`) and `no_whitespaces` (see above comment about "no").

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+    endpoint = len(full_string)
+    for i in range(len(full_string)):
+        if full_string[i].strip() != "" and not full_string[i][no_whitespaces].isspace():
+            endpoint = i
+            break
+
+    return "\n".join(full_string[:endpoint])
+
+
+def setup_table(table: AccumulationTable) -> None:
```
david-yz-liu, line 41: This should be an instance method of the class, as its primary focus is on something the class instance is responsible for.

In terms of function design, this function is doing a bit too much and should be split up. A natural part to extract is the code that returns the `For` node.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    Get the frame of the code containing the with statement, cut down the source code
+    such that it only contains the with statement and the accumulator loop and set up
+    the trace function to track the values of the accumulator variables during each iteration
+    """
+    class_frame = inspect.getouterframes(inspect.currentframe())[1].frame
```
david-yz-liu, line 47: I'm not sure why this line is necessary. Seems like you should be able to get the `func_frame` directly from the call stack.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+    _zero_iteration: bool
+
+    def __init__(self, accumulation_names: list) -> None:
+        self.loop_accumulators = {accumulator: [] for accumulator in accumulation_names}
+        self.loop_var_name = ""
+        self.loop_var_val = []
+        self._loop_lineno = 0
+        self._zero_iteration = True
+
+    def get_lineno(self, line_no: int) -> None:
```
david-yz-liu, line 94: This doesn't need to be a separate method. After moving `setup_table` into this class you can then set `_loop_lineno` directly.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+        """Get the line number of the for loop nested in the with statement"""
+        self._loop_lineno = line_no
+
+    def _add_iteration(self, val: list, var: Any) -> None:
```
david-yz-liu, line 98: Rename "add" to "record", which better describes the purpose of this method. Don't forget to also update the docstring.

david-yz-liu: Also, pick better names for `val` and `var`.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+        """Add the values of the accumulator and loop variables of an iteration"""
+        self.loop_var_val.append(var)
+        for index, key in enumerate(self.loop_accumulators):
```
david-yz-liu, line 101: Pick a better name for `key`

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+        self.loop_var_val = []
+        self._loop_lineno = 0
+        self._zero_iteration = True
+
+    def get_lineno(self, line_no: int) -> None:
+        """Get the line number of the for loop nested in the with statement"""
+        self._loop_lineno = line_no
+
+    def _add_iteration(self, val: list, var: Any) -> None:
+        """Add the values of the accumulator and loop variables of an iteration"""
```
david-yz-liu, line 99: Check the singular/plural forms for "accumulator" and "loop variables". How many of each kind do we track?

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+        self.loop_var_val.append(var)
+        for index, key in enumerate(self.loop_accumulators):
+            self.loop_accumulators[key].append(val[index])
+
+    def _create_iteration_dict(self) -> dict:
+        """Use the values of each iteration and return a dictionary that maps each accumulator
```
david-yz-liu, line 105: Remember that docstrings should describe what a function does, but not how it works. You don't need the phrase "use the values of each iteration" here

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+        and loop variable to its respective value during each iteration
+        """
+        return {
+            "iteration": list(range(len(self.loop_accumulators[list(self.loop_accumulators)[0]]))),
```
david-yz-liu, line 109: This can be simplified if you use `self.loop_var_val` instead

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+    def _tabulate_data(self) -> None:
+        """Print the values of the accumulator and loop variables into a table"""
+        iteration_dict = self._create_iteration_dict()
+        print(
+            tabulate.tabulate(
+                iteration_dict, headers="keys", colalign=(*["left"] * len(iteration_dict),)
+            )
+        )
+
+    def add_zero_iter(self, func_node: types.NodeNG, func_frame: types.FrameType) -> None:
```
david-yz-liu, line 123: You don't need a separate method here. You can split up this logic as follows:

1. Initialize `self.loop_var_name` in `setup_table`
2. Do the updating of `self.loop_accumulators` and `self.loop_var_val` in `add_iteration`. Note that you can distinguish between the 0-th and subsequent iterations by checking for the length of `self.loop_var_val`. This means you can remove the `_zero_iteration` attribute.
    
    (As a note, also move the `self.loop_var_val.append("N/A")` into `add_iteration`.)

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+        self.loop_var_name = func_node.target.name
+        for name in self.loop_accumulators:
+            if name in func_frame.f_locals:
+                self.loop_accumulators[name] = [func_frame.f_locals[name]]
+            else:
+                raise NameError
+
+    def trace_loop(self, frame: types.FrameType, event: str, arg: Any) -> None:
+        """Trace through the for loop and store the values of the
+        accumulators during each iteration
```
david-yz-liu, line 134: You should mention "loop variable" here as well.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+                self._add_iteration(curr_vals, local_vars[self.loop_var_name])
+
+            self._zero_iteration = False
+
+    def __enter__(self) -> AccumulationTable:
+        self.loop_var_val.append("N/A")
+        setup_table(self)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
```
david-yz-liu, line 155: Add type annotations here

sample_usage/print_table.py
```diff
@@ -0,0 +1,63 @@
+"""
+Examples of how to use the AccumulationTable Class with its
```
david-yz-liu, line 2: You can remove the phrase "with its respective outputs", I'm not sure exactly what you mean. The main block is good and speaks for itself.

tests/test_debug/test_accumulationtable.py
```diff
@@ -0,0 +1,58 @@
...
+    assert table.loop_var_val == ["N/A", 10, 20, 30]
+    assert table.loop_accumulators == {
+        "sum_so_far": [0, 10, 30, 60],
+        "list_so_far": [[], [10], [10, 20], [10, 20, 30]],
+    }
+    assert table.loop_var_name == "number"
+
+
+class MyClass:
+    items = list
```
david-yz-liu, line 39: this type annotation uses incorrect syntax (cc @peelstnac)

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,158 @@
...
+                    curr_vals.append(local_vars[accumulator])
+                else:
+                    raise NameError
+
+            if self.loop_var_name in local_vars and not self._zero_iteration:
+                self._add_iteration(curr_vals, local_vars[self.loop_var_name])
+
+            self._zero_iteration = False
+
+    def __enter__(self) -> AccumulationTable:
```
david-yz-liu, line 150: Include a docstring to specify what happens when the "with context" is entered. You may wish to combine some parts of `setup_table` with this.

python_ta/debug/accumulation_table.py
david-yz-liu: Also, iInclude a docstring to specify what happens when the "with context" is exited. You may wish to combine some parts of `setup_table` with this.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,155 @@
...
+
+import inspect
+import sys
+
+import astroid
+import tabulate
+
+
+def num_whitespaces(start_of_loop: str) -> int:
+    """Number of spaces at the beginning of the accumulation loop"""
```
david-yz-liu, line 15: Generally the function description should state what the function does, written in the imperative tense (e.g., "Return ...")

So here you should reword this to "Return the number of spaces..."

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,155 @@
...
+        if char.isspace():
+            blank_chars += 1
+        else:
+            break
+
+    return blank_chars
+
+
+def get_loop_lines(lines: list[str], num_whitespace: int) -> str:
+    """Function that returns the lines from the start to the end
```
david-yz-liu, line 27: Similarly here, reword to "Return the lines..."

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,155 @@
...
+    endpoint = len(lines)
+    for i in range(len(lines)):
+        if lines[i].strip() != "" and not lines[i][num_whitespace].isspace():
+            endpoint = i
+            break
+
+    return "\n".join(lines[:endpoint])
+
+
+def get_for_node(frame: types.FrameType) -> types.NodeNG:
```
david-yz-liu, line 39: `types` is an actual Python module (see its [documentation](https://docs.python.org/3/library/types.html)). You need to import it in order for type checkers (including PyCharm) to verify the actual type.

For the return type, the correct type is `For`, not `NodeNG`. This is also imported from `astroid`, not `types`.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,155 @@
...
+
+    return astroid.parse(loop_lines).body[0]
+
+
+class AccumulationTable:
+    """
+    Class used as a form of print debugging to analyze different loop and
+    accumulation variables during each iteration in a for loop
+
+    Private instance attributes:
```
david-yz-liu, line 56: You can just call this `Instance attributes`, since not all of them are private.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,155 @@
...
+    _loop_lineno: int
+
+    def __init__(self, accumulation_names: list[str]) -> None:
+        self.loop_accumulators = {accumulator: [] for accumulator in accumulation_names}
+        self.loop_var_name = ""
+        self.loop_var_val = []
+        self._loop_lineno = 0
+
+    def _record_iteration(
+        self, accumulator_values: list, loop_variable_value: Any, frame: types.FrameType
```
david-yz-liu, line 77: Okay this function is definitely better, but can be simplified even further. The only argument you need to pass in here is the `frame`. The current accumulator values and loop variable value can be computed from `frame.f_locals`; currently, you're doing this outside of the function (in `_trace_loop`), but really you can always do it in this function.

Note that if you do so, the code for updating the *accumulator values* should be the same for the zero-th and every other iteration; the only part that requires branching is choosing the value to append to `self.loop_var_val`.

python_ta/debug/accumulation_table.py
```diff
@@ -0,0 +1,155 @@
...
+        if event == "line" and frame.f_lineno == self._loop_lineno:
+            local_vars = frame.f_locals
+            curr_vals = []
+            for accumulator in self.loop_accumulators:
+                if accumulator in local_vars:
+                    curr_vals.append(local_vars[accumulator])
+                else:
+                    raise NameError
+
+            if self.loop_var_name in local_vars and len(self.loop_var_val) > 0:
```
david-yz-liu, line 125: This check isn't necessary. It's okay to have `self._record_iteration` execute every time the loop line is reached, including the very first time ("iteration 0"). This means you do not have to call `self._record_iteration` in the `_setup_table` method, just here.

PR #830, opened 2022-07-07
Code cleanup startup logic

tupi2001:
## Motivation and Context
Code cleanup to automatically generate a list of checkers rather than writing them down manually.

## Your Changes

Used the absolute path for the checkers files to generate the list.

**Description**:

## Testing

After changes were made, tested to make sure functionality stayed the same.

## Checklist


python_ta/__init__.py
```diff
@@ -272,21 +273,28 @@ def reset_linter(config=None, file_linted=None):
         ),
     )
 
+    absolute_checkers_path = Path("python_ta//checkers").absolute()
```
david-yz-liu, line 276: Using `python_ta/checkers` still produces a path that's relative to where the user is running PythonTA, so that's not what we want. You should use the special `__file__` variable to get the location of this file, and then go from there.

python_ta/__init__.py
```diff
@@ -272,21 +273,28 @@ def reset_linter(config=None, file_linted=None):
...
-        "python_ta.checkers.invalid_range_index_checker",
-        "python_ta.checkers.one_iteration_checker",
-        "python_ta.checkers.type_annotation_checker",
-        "python_ta.checkers.unnecessary_indexing_checker",
-        "python_ta.checkers.shadowing_in_comprehension_checker",
-        "python_ta.checkers.redundant_assignment_checker",
-        "python_ta.checkers.invalid_for_target_checker",
-        "python_ta.checkers.missing_space_in_doctest_checker",
-        "python_ta.checkers.pycodestyle_checker",
+        ("python_ta.checkers." + f[:-3])
```
david-yz-liu, line 279: You can make the `f[:-3]` a bit more readable by using [os.path.splitext](https://docs.python.org/3/library/os.path.html#os.path.splitext).

python_ta/__init__.py
```diff
@@ -272,21 +273,28 @@ def reset_linter(config=None, file_linted=None):
...
+        for f in listdir(path_as_string)
+        if f != "__init__.py" and f != "__pycache__"
```
david-yz-liu, line 281: This condition isn't totally precise. You should use `os.path.splitext` here as well, and check whether the extension is `.py` or not, and if the file name is `__init__.py` (which is the one `.py` file we want to ignore).

python_ta/__init__.py
```diff
@@ -272,21 +273,28 @@ def reset_linter(config=None, file_linted=None):
...
     ]
+    #     [
```
david-yz-liu, line 283: Please delete all commented out code before sending your work for review. Remember that you can always access previous versions of files using git by either going back to previous commits, or other branches.

python_ta/__init__.py
```diff
@@ -319,8 +327,8 @@ def reset_linter(config=None, file_linted=None):
                 linter.set_option(key, config[key])
 
     # Custom checker configuration.
-    if linter.config.pyta_type_check:
-        linter.load_plugin_modules(["python_ta.checkers.type_inference_checker"])
+    # if linter.config.pyta_type_check:
```
david-yz-liu, line 330: In addition to deleting this commented code, you should also delete the comment immediately above it. Also you should delete the `pyta_type_check` option from the configuration file.

python_ta/config/.pylintrc
```diff
@@ -71,6 +71,7 @@ disable=
     W1406,
     E1507,
     E1700, E1701,
+    E9900,
```
david-yz-liu, line 74: This is good; you should also change the custom `.pylintrc` files we use for testing (search in the `tests` folder for other `.pylintrc` files).

python_ta/checkers/forbidden_io_function_checker.py
```diff
@@ -57,16 +57,16 @@ def visit_call(self, node):
                 if (
                     isinstance(scope_parent, nodes.ClassDef)
                     and isinstance(scope, nodes.FunctionDef)
-                    and (scope_parent.name + "." + scope.name) not in self.linter.config.allowed_io
+                    and (scope_parent.name + "." + scope.name) not in self.config.allowed_io
```
david-yz-liu, line 60: I think this is leftover from the merge conflict: you should revert the changes you see here (using `self.linter.config` rather than `self.config` throughout)

python_ta/__init__.py
```diff
@@ -272,20 +273,13 @@ def reset_linter(config=None, file_linted=None):
...
-        "python_ta.checkers.type_annotation_checker",
-        "python_ta.checkers.unnecessary_indexing_checker",
-        "python_ta.checkers.shadowing_in_comprehension_checker",
-        "python_ta.checkers.redundant_assignment_checker",
-        "python_ta.checkers.invalid_for_target_checker",
-        "python_ta.checkers.missing_space_in_doctest_checker",
-        "python_ta.checkers.pycodestyle_checker",
+        ("python_ta.checkers." + os.path.splitext(f)[0])
+        for f in os.listdir(parent_dir_path + "/" + python_ta_dir[checkers_index])
+        if f != "__init__.py" and os.path.splitext(f)[1] != ""
```
david-yz-liu, line 282: Check for the extension being `"py"` rather than `!= ""` (because the former is more specific)

python_ta/__init__.py
```diff
@@ -272,20 +273,13 @@ def reset_linter(config=None, file_linted=None):
         ),
     )
 
+    parent_dir_path = os.path.dirname(__file__)
+    python_ta_dir = os.listdir(parent_dir_path)
+    checkers_index = python_ta_dir.index("checkers")
```
david-yz-liu, line 278: This works but is a bit redundant. You don't need to use `checkers_index` at all, since you're just looking for the directory `checkers`, which you know exists. So you can just append `/checkers` directly to the parent directory

david-yz-liu (approved):

PR #832, opened 2022-07-25
Astroid transformer

peelstnac:
## Motivation and Context

We lay the groundwork to extend the capabilities of pylint by translating assertions found in docstrings of astroid nodes (currently only FunctionDef and preconditions) into z3 expressions. We do so with an astroid visitor which visits the nodes in question, parses the assertions, performs the translations, and appends them to the node they belong to.

## Your Changes

**Description**:

The changes come mainly from two core files: `python_ta/transforms/ExprWrapper.py` and `python_ta/transforms/z3_visitor.py`. The former houses a class `ExprWrapper` which is used to translate astroid Expr nodes into z3 ExprRef expressions. The latter is the aforementioned visitor class.

## Testing

Tests are included in `tests/test_z3_visitor.py`. Preconditions are checked by solving `actual == expected` with z3: they pass if the result is satisfiable. 

## Checklist


python_ta/transforms/ExprWrapper.py
```diff
@@ -0,0 +1,139 @@
...
+        if typ == "float":
+            return z3.Real(name)
+        if typ == "bool":
+            return z3.Bool(name)
+
+        return None
+
+    def parse_compare(self, node: astroid.Compare) -> Optional[z3.ExprRef]:
+        left, ops = node.left, node.ops
+        # ERROR: left, right can be None at any point in the execution
```
david-yz-liu, line 55: Right, so using exceptions will remove this comment. (Basically we're assuming that it's just `self.reduce` and `self.apply_bin_op` that raise errors, but this method itself doesn't *assuming* that the others don't. This method can't introduce new errors.)

python_ta/transforms/ExprWrapper.py
```diff
@@ -0,0 +1,139 @@
...
+        # ERROR: see apply_bool_op ERROR comments
+        if left is None:
+            return None
+        try:
+            if op == "not":
+                return z3.Not(left)
+            else:
+                return None
+        except z3.z3types.Z3Exception:
+            return None
```
david-yz-liu, line 74: Here's an example where (based on earlier comment) you can just let this exception bubble up.

python_ta/transforms/ExprWrapper.py
```diff
@@ -0,0 +1,139 @@
...
+                return left <= right
+            elif op == ">=":
+                return left >= right
+            elif op == "<":
+                return left < right
+            elif op == ">":
+                return left > right
+            else:
+                return None
+        except TypeError:
```
david-yz-liu, line 106: Okay, so here I think it would be good to raise a custom exception from inside the `except`. E.g. `raise Z3CustomException(f'variables of incompatible types {left} and {right} for operation {op}')`

python_ta/transforms/z3_visitor.py
```diff
@@ -0,0 +1,43 @@
+import astroid
+from astroid import Uninferable, nodes
+from astroid.transforms import TransformVisitor
+
+from ..contracts import parse_assertions
+from .ExprWrapper import ExprWrapper
+
+
+class Z3Visitor:
```
david-yz-liu, line 9: Please add a docstring here. (This is especially important because this is the main entrypoint into your Z3 work!)

python_ta/transforms/z3_visitor.py
```diff
@@ -0,0 +1,43 @@
...
+        docstring = node.doc
+        preconditions = parse_assertions(docstring, parse_token="Precondition")
+        # Get z3 constraints
+        z3_constraints = []
+        for pre in preconditions:
+            pre = astroid.parse(pre).body[0]
+            ew = ExprWrapper(pre, types)
+            transformed = ew.transformed
+            if transformed is not None:
+                z3_constraints.append(transformed)
```
david-yz-liu, line 40: Right so to be clear, this loop body is where I think the "main" try-except should go.

setup.cfg
```diff
@@ -40,6 +40,8 @@ dev =
     pytest-cov
     sphinx
     sphinx-rtd-theme
+z3 =
+    z3-solver
```
david-yz-liu, line 44: See my comment on Slack, but the `test.yml` file also needs to be updated with this dependency change.

tests/test_z3_visitor.py
```diff
@@ -0,0 +1,54 @@
...
+    function_def = mod.body[0]
+    # Declare variables
+    x = z3.Int("x")
+    y = z3.Int("y")
+    z = z3.Real("z")
+    # Construct expected
+    expected = z3.And([x**2 + y**2 == z**2, z3.And([x > 0, y > 0, z == 0])])
+    actual = z3.And(function_def.z3_constraints)
+    solver = z3.Solver()
+    solver.add(actual == expected)
```
david-yz-liu, line 28: Okay in terms of testing methodology, I'm okay with this general approach and using the `solver.check()` at the end. *However* to make the testing a bit more granular and robust against the "contradiction" error we talked about in our meeting, please loop over the `z3_constraints` attribute and check each generated constraint separately. (Rather than the current approach, which `And`s all the constraints together.)

david-yz-liu (commented):
@peelstnac great start! I left some general comments for you to work on. Please let me know if you have any questions!

python_ta/contracts/__init__.py
```diff
@@ -430,7 +430,10 @@ def parse_assertions(obj: Any, parse_token: str = "Precondition") -> List[str]:
        line is of the form "- <cond>". Each line is considered a separate condition.
        The lines can be separated by blank lines, but no other text.
     """
-    docstring = getattr(obj, "__doc__") or ""
+    if isinstance(obj, str):
```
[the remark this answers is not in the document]
peelstnac: The only problem that I'm running into is that the original function takes `node.__doc__`, but I want `node.doc`.

python_ta/transforms/ExprWrapper.py
```diff
@@ -0,0 +1,139 @@
...
+import astroid
+import z3
+from astroid import nodes
+
+
+class ExprWrapper:
+    node: astroid.NodeNG
+    types = {}
+
+    def __init__(self, expr: nodes.Expr, types={}):
```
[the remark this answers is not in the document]
peelstnac: I am in disbelief. I've been doing this for so long and never noticed that default arguments only evaluated once...

python_ta/contracts/__init__.py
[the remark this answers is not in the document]
david-yz-liu: @peelstnac ah great point. Still, I'd rather try to access both variables here to support passing in astroid nodes, rather than passing in a string directly. Something like `getattr(obj, "__doc__") or getattr(obj, "doc") or ""`.

PR #846, opened 2022-09-22
Added venv to gitignore and changed install command for different CLI

ZhiYingSun:
## Motivation and Context
It is easier to manage the dependencies of python projects using a virtual environment and we should add support for it in the gitignore file. 

When installing dependencies using macOS, the command did not work on zsh and it took time to figure out the issue. I propose a new command that should work for bash, zsh, Command shell, and PowerShell. This should work for most operating systems.

## Your Changes

**Description**:
- added venv to gitignore
- changed install command for different CLI

## Testing

Tested the command on zsh and this [discussion](https://github.com/mu-editor/mu/issues/852) tested it for other CLIs.  
## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ tests/test_type_inference/png*
 
 # Sphinx documentation build files
 docs/_build
+
+# Environments
+venv/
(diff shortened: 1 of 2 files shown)


david-yz-liu (approved):

david-yz-liu:
Thanks @ZhiYingSun!

PR #852, opened 2022-10-08
IOFunctionChecker Modification to check for calls at top-level

ZhiYingSun:
## Motivation and Context

`IOFunctionChecker`: this function checks for using `print` and similar input/output functions within function definitions. (Most commonly used to flag students who use print for debugging inside a function.) The change here is to modify the checker so that it also checks for calls to `print` (and similar functions) written at the top-level of a module (not within a function body), but outside of the main block. The global variables checker has a `is_in_main` function that you might find useful.

## Your Changes

- Added logic for checking for top-level module calls for IO functions that is not in the main block
- Added one test for a call to a IO function in the top-level
- Added one test for a call to a IO function at the top-level but in a main block

**Description**:

## Testing

- added `test_message_global_allowed_main_block`
- added `test_message_global`

## Checklist

- [ ] I have performed a self-review of my own code.
- [ ] I have updated the CHANGELOG.md file.

### DIFF
--- a/python_ta/checkers/forbidden_io_function_checker.py
+++ b/python_ta/checkers/forbidden_io_function_checker.py
@@ -5,6 +5,8 @@ from astroid import nodes
 from pylint.checkers import BaseChecker
 from pylint.checkers.utils import only_required_for_messages
 
+from python_ta.checkers.global_variables_checker import is_in_main
+
 FORBIDDEN_BUILTIN = ["input", "print", "open"]
 
 
@@ -52,7 +54,6 @@ class IOFunctionChecker(BaseChecker):
             if not (name in node.frame() or name in node.root()):
                 scope = node.scope()
                 scope_parent = scope.parent
-                # TODO: Only FunctionDefs are checked. Include global scope?
 
                 if (
                     isinstance(scope_parent, nodes.ClassDef)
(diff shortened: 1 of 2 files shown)


python_ta/checkers/forbidden_io_function_checker.py
```diff
@@ -5,6 +5,8 @@
 from pylint.checkers import BaseChecker
 from pylint.checkers.utils import only_required_for_messages
 
+from python_ta.checkers.global_variables_checker import is_in_main
```
david-yz-liu, line 8: It's not common for one checker to import another. Instead, please move the `is_in_main` function into the `utils` module, and then import it into both checkers.

python_ta/checkers/forbidden_io_function_checker.py
```diff
@@ -68,6 +69,13 @@ def visit_call(self, node):
                 ):
                     if name in self.linter.config.forbidden_io_functions:
                         self.add_message("forbidden-IO-function", node=node, args=name)
+                elif (
+                    isinstance(scope_parent, type(None))
```
david-yz-liu, line 73: This isn't necessary, it's enough to check `isinstance(scope, nodes.Module)`

david-yz-liu (commented):
@ZhiYingSun looks great! I left a few comments, and also yes please update the documentation with further examples and clarification of the different cases.

python_ta/checkers/forbidden_io_function_checker.py
ZhiYingSun: Makes sense! Addressed.

python_ta/checkers/forbidden_io_function_checker.py
ZhiYingSun: Makes sense! Moved to the utils module.

ZhiYingSun:
Updated documentation with the one case that allows for an IO call that is not explicitly allowed.
image

david-yz-liu (approved):
@ZhiYingSun great work!

PR #857, opened 2022-10-24
Modified invalid for target checker to handle comprehensions

yibinglu:
## Motivation and Context
The motivation was to modify invalid-for-target-checker to check for subscripting in the target of a comprehension and for loops.

## Your Changes

**Description**:
I added a new visit_comprehensions method to the main checker file, added comprehension versions of all the tests in the test file, and updated the PyTa documentation and examples to include mention of comprehensions.

## Testing

I added comprehension versions of all the tests that were already present in the testing file that checked for loops and then added one more that tests for slicing. I didn't add any tests that checked for edge cases in the body of the comprehension or possible if-conditions since this checker seems to only focus on the target but please let me know if those are necessary.

I noticed that the name for the checker in the code (invalid for target), doesn't match up with what's in the documentation (for target subscript), should they be modified to be consistent? Also, the title is specific to for loops so I'm not sure if we want to change that to be more general as well.

## Checklist

### DIFF
--- a/docs/checkers/index.md
+++ b/docs/checkers/index.md
@@ -2112,7 +2112,7 @@ Comprehension:
 
 ```python
 def example7(lst: List[int]) -> List[int]:
-    return [lst[i] for i in lst]  # Fixed
+    return [number for number in lst]  # Fixed
 ```
 
 (E9969)=
--- a/docs/checkers/index.md
+++ b/docs/checkers/index.md
@@ -2049,10 +2049,10 @@ def print_sum(lst1: List[int], lst2: List[int]) -> None:
 
 ### For Target Subscript (E9984)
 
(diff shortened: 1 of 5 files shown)


docs/checkers/index.md
```diff
@@ -2098,6 +2108,13 @@ def example1(lst: List[int]) -> int:
     return s
 ```
 
+Comprehension:
+
+```python
+def example7(lst: List[int]) -> List[int]:
+    return [lst[i] for i in lst]  # Fixed
```
david-yz-liu, line 2115: This correction doesn't really make sense. The variable `i` isn't a good variable name, as `i` doesn't refer to an index, but an element of the list.

david-yz-liu (approved):

PR #858, opened 2022-10-30
Issue-831: Added check in pyta contracts to differentiate types float and int

### LINKED ISSUE #831: Contract checker bug?
Hi, 
I was testing out the `check_all_contracts` function on a basic example as follows:

```python
def f(a: int, b: int) -> float:
    """Return the result when dividing a by b.

    >>> f(10, 5)
    2.0
    """
    return a // b


if __name__ == "__main__":
    import python_ta.contracts
    python_ta.contracts.check_all_contracts()
    print(type(f(10, 5)))
    print(python_ta.__version__)
```
running this gives the output

```shell
<class 'int'>
2.2.0
```

I would think this should get caught as a violation of the type contract because the return type is `int`, but the type annotation indicates `float`. But this does not get caught as an `AssertionError`. Am I overlooking something super basic here ? The python version is `3.10.4`.

Thanks,
Amanjit

yibinglu:
## Motivation and Context
Resolves #831: Pyta contracts is unable to detect a type error when the expected type is `float` but got `int` instead.

## Your Changes

I added a check within `_check_function_contracts` in `contracts\__init__.py` that checks if the return type is `float` but returned `int` instead.

**Description**:

## Testing

I added three tests that are checking return type: one that simply checks that a return type float passes, one for when the return type is `float` and got `int`, and one for when the return type is `int` and got `float`. I also added a few tests that checked for function parameter types. I added a function that was the same as `_my_sum` but the parameter type is a list of floats. To avoid confusion, I changed the previous test to `_my_sum_int` and the new one to `_my_sum_float`.

## Checklist


python_ta/contracts/__init__.py
```diff
@@ -184,6 +184,11 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
         if param in annotations:
             try:
                 _debug(f"Checking type of parameter {param} in call to {wrapped.__qualname__}")
+                if type(arg) is int and annotations[param] is float:
```
david-yz-liu, line 187: It's not great to duplicate this code in two different places. Instead, create a new helper function that wraps both the check you added and the call to `check_type` itself. The new function should take the same arguments as `check_type` itself.

yibinglu: I've now updated the helper function to raise a TypeError.

python_ta/contracts/__init__.py
```diff
@@ -259,6 +259,13 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
     return r
 
 
+def check_type_strict(argname: str, value, expected_type) -> None:
```
david-yz-liu, line 262: Please put in type annotations for `value` (`Any`) and `expected_type` (`type`).

david-yz-liu (approved):

PR #861, opened 2022-11-18
Added instance attributes to rep invariant error message 

yibinglu:
## Motivation and Context

The motivation for this pull request was to modify the error message for a violated representation invariant to include the instance attributes' current values.

## Your Changes

**Description**:

In `python_ta/contracts/__init__.py`, I modified the check so that after the check returns false (invariant was violated), it would then access the current instance attributes and include those in the error message raised. I also fixed two typos in the test file. One that said "valid" when it was testing for an invalid instance and another whose docstring value for an attribute didn't match up with what was in the precondition (100 instead of 150).

## Testing

Because this change only affected the error message that was produced I just modified the existing representation invariant test for `SetWrapper` to check if the arguments were included as well in the assertion. I did add two more tests that were also checking for this with the `Pizza` class, one for valid and one for invalid. The `Pizza` tests included two invariants and it should raise an error when one of them is violated.

## Checklist

### DIFF
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -345,7 +345,12 @@ def _check_invariants(instance, klass: type, global_scope: dict) -> None:
             _debug(f"Warning: could not evaluate representation invariant: {invariant}")
         else:
             if not check:
-                raise PyTAContractError(f'Representation invariant "{invariant}" was violated')
+                arg_string = instance.__dict__
+
+                raise PyTAContractError(
+                    f'Representation invariant "{invariant}" was violated for'
+                    f" arguments {arg_string}"
+                )
 
 
 def _get_legal_return_val_var_name(var_dict: dict) -> str:
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -352,11 +352,15 @@ def _check_invariants(instance, klass: type, global_scope: dict) -> None:
             _debug(f"Warning: could not evaluate representation invariant: {invariant}")
         else:
             if not check:
-                arg_string = instance.__dict__
+                curr_attributes = ", ".join(
+                    f"{k}: {_display_value(v)}" for k, v in vars(instance).items()
+                )
+
+                curr_attributes = "{" + curr_attributes + "}"
 
                 raise PyTAContractError(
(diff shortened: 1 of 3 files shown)


tests/test_class_contracts.py
```diff
@@ -217,6 +217,27 @@ def test_circle_area_invalid() -> None:
...
+def test_pizza_invalid() -> None:
+    """
+    Test the Pizza representation invariant on an invalid instance.
+    """
+    with pytest.raises(AssertionError) as excinfo:
+        Pizza(radius=10, ingredients=[])
+    msg = str(excinfo.value)
+    assert (
+        "\"len(self.ingredients) > 0\" was violated for arguments {'radius': 10, 'ingredients': []}"
+        in msg
```
david-yz-liu, line 237: You shouldn't use the `in` operator, you can check for exact string equality

yibinglu: I noticed that the `in` operator is used in a few of the old tests as well, should I change them all to use `==` instead?

david-yz-liu: @yibinglu the other tests use `in` because they're just checking that the correct assertion expression is in the message. But in this case, you're writing out the full message, so `==` is more appropriate.

python_ta/contracts/__init__.py
```diff
@@ -352,7 +352,16 @@ def _check_invariants(instance, klass: type, global_scope: dict) -> None:
...
             if not check:
-                raise PyTAContractError(f'Representation invariant "{invariant}" was violated')
+                curr_attributes = ", ".join(
+                    f"{k}: {_display_value(v)}" for k, v in vars(instance).items()
+                )
+
+                curr_attributes = "{" + curr_attributes + "}"
+
+                raise PyTAContractError(
+                    f'Representation invariant "{invariant}" was violated for'
```
david-yz-liu, line 362: @yibinglu as I was reviewing this I realized we can improve the error message further by writing the class name here as well (as we do on line 348, except use `instance.__class.__name__` instead of `.__qualname__`). So the message would read something like `Person representation invariant "..." was violated for ...`.

Please make that change, and update the changelog entry to mention that we're adding both the class name and instance attribute values to the message.

david-yz-liu (approved):

PR #862, opened 2022-11-24
Type contracts differentiate bool and int

yibinglu:
## Motivation and Context

This made the Pyta type check stricter by raising an `AssertionError` when the expected type is `int` but got `bool` instead.

## Your Changes

**Description**:
Added another condition to `check_type_strict()` that checks if the expected type is `int` and got `bool` instead. If it meets that condition, raise a `TypeError`.

## Testing
I added three tests: one that checks for the correct return type `bool`, one that is expecting an `int` and got `bool`, and one that is expecting `bool` and got `int`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Bug Fixes
 
 - Fixed Issue #831: Contract Checker Bug. Now raises `AssertionError` when the expected type is `float` but got `int` instead.
-- Now raises `AssertionError` when the expected type is `int` but got `bool` instead.
+- PyTA contracts' type checking now raises `AssertionError` when the expected type is `int` but got `bool` instead.
 
 ## [2.3.3] - 2022-09-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Bug Fixes
 
 - Fixed Issue #831: Contract Checker Bug. Now raises `AssertionError` when the expected type is `float` but got `int` instead.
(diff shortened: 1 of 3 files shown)


python_ta/contracts/__init__.py
```diff
@@ -260,8 +260,18 @@ def _check_function_contracts(wrapped, instance, args, kwargs):
...
-    """Ensure that ``value`` matches ``expected_type``. Differentiates between types float and int."""
-    if type(value) is int and expected_type is float:
+    """Ensure that ``value`` matches ``expected_type``.
+
+    Differentiates between:
+        - float vs. int
+        - bool vs. int
+    """
+    if (
+        type(value) is int
```
david-yz-liu, line 270: This condition broken up into multiple lines is a bit hard to read, especially since the precedence rules of `and` vs. `or` aren't well-known.

I would rewrite this condition as

```
if ((type(value) is int and expected_type is float) or
    (type(value) is bool and expected_type is int):
    ...
```

and then let the `black` pre-commit hook format the code from there.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Bug Fixes
 
 - Fixed Issue #831: Contract Checker Bug. Now raises `AssertionError` when the expected type is `float` but got `int` instead.
+- Now raises `AssertionError` when the expected type is `int` but got `bool` instead.
```
david-yz-liu, line 18: You need to mention contract checking here, otherwise it's not clear what the subject of "raises" is. (This was implied by include the issue title in the previous entry.)

tests/test_contracts.py
```diff
@@ -36,6 +36,16 @@ def nullary() -> Dict[str, int]:
     nullary()
 
 
+def test_nullary_return_bool() -> None:
```
david-yz-liu, line 39: This is good, but you should add some tests for parameters as well.

david-yz-liu (approved):

PR #863, opened 2022-12-03
Added ENABLE_CONTRACT_CHECKING 

yibinglu:
## Motivation and Context

This change was made so that the contract checking for preconditions and representation invariants would only be checked if this new constant `ENABLE_CONTRACT_CHECKING` was set to `True`, otherwise, skip it.

## Your Changes

**Description**:

I added the new constant at the top of the `__init__.py` file with its default value set to `True`. Then I found each location in the code that called `check_assertion` for the precondition and `check_invariant` for the representation invariant and added a condition that `ENABLE_CONTRACT_CHECKING` must be `True` for the methods to run.

## Testing

I added two tests, one for precondition and one for representation invariant. These tests set the value of `ENABLE_CONTRACT_CHECKING` to `False` and provide an invalid input that would usually raise an error, but in this case, do not. I did not have any tests that set `ENABLE_CONTRACT_CHECKING` to `True` since that is the default value for all the other tests. I also didn't include the case where `ENABLE_CONTRACT_CHECKING` is `False` and the input is valid since there would be no error raised regardless of `ENABLE_CONTRACT_CHECKING`.

## Checklist


python_ta/contracts/__init__.py
```diff
@@ -18,6 +18,7 @@
 import wrapt
 from typeguard import check_type
 
+ENABLE_CONTRACT_CHECKING = True
```
david-yz-liu, line 21: Include a docstring below this variable, like `DEBUG_CONTRACTS`. That will be included when we generate docs for this module (you can check this by running `make docs`).

david-yz-liu (commented):
@yibinglu Thanks! I left one comment, and also you should modify the `check_type_strict` function as well (and add tests for that).

david-yz-liu (commented):
@yibinglu oh I just realized one more thing, in `_check_function_contracts` there's a section near the bottom that checks postconditions (you may not have realized PythonTA can check those too). You should add a check there as well, and a test for it. 

You can see the documentation for postcondition checking here: https://www.cs.toronto.edu/~david/pyta/contracts/index.html#functions-custom-postconditions

python_ta/contracts/__init__.py
david-yz-liu: @yibinglu I don't think you checked this on the HTML file properly. It turns out you need to add this to the `index.md` file to get it to appear. Please add an entry there as well.

david-yz-liu (approved):

PR #864, opened 2022-12-04
Extend accumulation table loop detection

ZhiYingSun:
## Motivation and Context
Our current limitation for the AccumulationTable is that for loop/while loop must be the first statement within the with block. In this pull request, we extend the loop detection to find the **first loop** within the with statement.

```python
def f():
    with AccumulationTable(['sum_so_far', 'prod_so_far', 'blah']):
        sum_so_far = 0
        for x in [1, 2, 3]:
            sum_so_far += x

    return sum_so_far
```

## Your Changes

**Description**:
- renamed the function `get_loop_lines` to `get_with_lines` since it gets all the statements within the with statement
- instead of returning the first node, we will iterate through all of the nodes inside of the with statement until we find the **first loop**
- changed the initialization of `self._loop_lineno` to use the line number of the with statement +  the line number of the loop with respect to the with statement

## Testing

`test_three_different_loop_lineno`
- checks loop detection works with code statements before the loop

`test_four_different_loop_lineno`
- checks loop detection works with code statements after the loop

**Questions**:
- Should I add this change to the Changelog or specify only the first loop will be accumulated in the documentation?

**Comments**:
- This limitation is not included in the [Current limitation section](https://www.cs.toronto.edu/~david/pyta/debug/index.html#current-limitations) of the documentation.
- I thought about simplifying `get_loop_lines` further to directly loop through the full function, but then we would have to find the with statement and iterate inside of it.

## Checklist


ZhiYingSun:
Question: Should I add this change to the changelog or specify only the first loop will be accumulated in the documentation?

python_ta/debug/accumulation_table.py
```diff
@@ -26,7 +26,7 @@ def num_whitespaces(start_of_loop: str) -> int:
     return blank_chars
 
 
-def get_loop_lines(lines: list[str], num_whitespace: int) -> str:
+def get_with_lines(lines: list[str], num_whitespace: int) -> str:
```
david-yz-liu, line 29: This is good, but you should update the docstring as well.

python_ta/debug/accumulation_table.py
```diff
@@ -46,9 +46,11 @@ def get_loop_node(frame: types.FrameType) -> Union[astroid.For, astroid.While]:
     lst_str_lines = func_string.splitlines()
     lst_from_with_stmt = lst_str_lines[with_stmt_index + 1 :]
     num_whitespace = num_whitespaces(lst_str_lines[with_stmt_index])
-    loop_lines = get_loop_lines(lst_from_with_stmt, num_whitespace)
+    with_lines = get_with_lines(lst_from_with_stmt, num_whitespace)
 
-    return astroid.parse(loop_lines).body[0]
+    for node in astroid.parse(with_lines).body:
```
david-yz-liu, line 51: I realized that we can extend this even further by using the astroid `nodes_of_class` method. `node.nodes_of_class((astroid.For, astroid.While))` returns a _generator_ for the nodes of those types contained in `node`.

You can call `next` on a generator to obtain the first item.

This will allow `AccumulationTable` to detect loops that aren't top-level statements in the `with`, but nested (e.g. within an if statement).

```python
with AccumulationTable(...):
    if something:
        for x in numbers:
            ...
```

Please add some tests for this as well!

david-yz-liu (commented):
@ZhiYingSun great work! I left a few comments. In response to your question, I just had a look at the documentation page and I think it's okay to leave it as is.

david-yz-liu:
@ZhiYingSun sorry I meant to _also_ say, please add a Changelog entry under "enhancements".

david-yz-liu (approved):

PR #878, opened 2023-02-14
Fixed bug where webpage opens on bad paths

sushimon:
## Motivation and Context

Calls to `check_all` and `check_errors` generate an error report when we pass in a path that doesn't exist or is unreadable. 

## Your Changes

**Description**:
- PythonTA no longer opens a webpage under these conditions but still prints the error message to the console.
- Added some logic to ensure at least one file has been checked before generating reports.
- Updated `CHANGELOG.md` to document this bug fix.

## Testing

- Ran the tests in the test suite and verified they passed.
- Ran `python` in `pyta` repository and called `check_all` and `check_errors` (in reality, they both use the same helper method) with erroneous inputs (both files and directories) and verified that an error message was displayed but no webpage opened. Also tested with different PythonTA reporters.
- Also tested the `check_all` and `check_errors` methods with valid inputs and verified this change did not affect the messages being displayed (e.g. removing some by accident).
- Previewed the `CHANGELOG.md` file to verify it now includes information related to this bug fix.

- As of now, the PR for the `ForbiddenPythonSyntaxChecker` hasn't been merged so once it does, there will might be a merge conflict with the `CHANGELOG.md` files so that needs to be dealt with

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,9 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
-## [2.4.3] - 2023-02-13
-
 ### Bug fixes
 
-- Fixed bug where `check_errors` and `check_all` open a webpage when a undefined or unreadable path is passed as an argument.
+- Fixed bug where `check_errors` and `check_all` opens a webpage when a undefined or unreadable path is passed as an argument.
 
 ## [2.4.2] - 2023-1-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+## [2.4.3] - 2023-02-13
+
+### Bug fixes
+
+- Fixed bug where `check_errors` and `check_all` open a webpage when a undefined or unreadable path is passed as an argument.
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+## [2.4.3] - 2023-02-13
```
david-yz-liu, line 10: Please remove this line (similar to the other PR)

python_ta/__init__.py
```diff
@@ -109,7 +109,9 @@ def _check(
 
     # Try to check file, issue error message for invalid files.
     try:
-        for locations in _get_valid_files_to_check(module_name):
+        files_to_check = list(_get_valid_files_to_check(module_name))
```
david-yz-liu, line 112: This isn't exactly the right place to do this because `file_to_check` might be a sequence of *directories*, not just files, and I'd like us to handle the case when the folder exists but it contains no files (also not showing the HTML report in this case). That's why there's an inner loop `for file_pay in get_file_paths(locations)` below.

I'd suggest using a boolean flag like "file found" or something and using that to indicate whether at least one file was actually checked or not.

CHANGELOG.md
sushimon: Fixed!

david-yz-liu (approved):

PR #879, opened 2023-02-23
Augmented CFG-related classes to support labelling edges

sushimon:
## Motivation and Context

Currently, the use of CFGs in PyTA is limited to checkers. This change should hopefully be the start of allowing them to be used for educational purposes.

## Your Changes

**Description**:
- Augmented the CFGEdge class to have a `label` attribute that is used to label each edge in the graph.
- Modified the `visit_if` method to label the edges transitioning from the if condition.
- Modified `create_block` and `link_or_merge` to include optional arguments for passing the edge label.

## Testing

- Added some tests to check whether labels were updated correctly.
- Ran the test suite and verified all tests passed.
- Created some CFGs using `draw_cfg` and made sure the changes didn't break anything.
- Ran a modified version of `draw_cfg` with the debugger and verified that the labels were correctly updated in `visit_if`.

- Is there a better way to check if the updating of labels works? What I've done is that I modified the `draw_cfg` method and use the debugger so I can view the state of the local variables when I run it in the Python console. I figure there has to be a better way though.

## Checklist

### DIFF
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -84,10 +84,8 @@ class CFGVisitor:
         node.cfg_block = self._current_block
         old_curr = self._current_block
 
-        # Handle "then" branch.
-        then_block = self._current_cfg.create_block(old_curr)
-        # Label the edge to the "then" branch
-        self._current_cfg.label_edge(old_curr, then_block, True)
+        # Handle "then" branch and label it.
+        then_block = self._current_cfg.create_block(old_curr, edge_label="True")
         self._current_block = then_block
         for child in node.body:
             child.accept(self)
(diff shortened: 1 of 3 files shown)


python_ta/cfg/visitor.py
```diff
@@ -86,6 +86,8 @@ def visit_if(self, node: nodes.If) -> None:
 
         # Handle "then" branch.
         then_block = self._current_cfg.create_block(old_curr)
+        # Label the edge to the "then" branch
+        self._current_cfg.label_edge(old_curr, then_block, True)
```
david-yz-liu, line 90: So overall this approach is pretty good, but has the downside that it requires a bit of iteration through existing edges. This isn't necessary in all the cases you've added here, which explicitly match a specific source-target pair.

So a better approach is to *pass the label as an optional argument to `create_block`*, and from there pass it down to `link_or_merge`, and to the Edge initializer. This would make the above code (on line 88) look like

```python
then_block = self._current_cfg.create_block(old_curr, edge_label='True')
```

or something. I would also pass in `'True'` rather than `True` to avoid confusion (but it's fine to leave the label as `Any`, I could see storing more information there in the future perhaps).

david-yz-liu (approved):

PR #883, opened 2023-03-16
Updated CFGVisitor with an options attribute to configure the visitor

sushimon:
## Motivation and Context

The CFGs produced aren't that user friendly because if conditions would get merged with preceding statements. Although the statements are still executed in sequence, it may be misleading so add an option to configure the visitor to not merge the condition with the previous statements.

## Your Changes

**Description**:

- Added an `options` dict to the `CFGVisitor` class which may be passed into the initializer to prevent merging if conditions with preceding statements.
- Added new test cases to verify that, when configured to do so, a new block is created to separate the if condition from preceding statements.

## Testing

- Ran the test suite and verified all the test cases passed.
- Verified the newly added test cases passed.
- Verified that the CFGs produced from `sample_usage.draw_cfg` were correct depending on how the visitor was initialized.

## Checklist

### DIFF
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from astroid import nodes
 
@@ -8,6 +8,11 @@ from .graph import CFGBlock, ControlFlowGraph
 class CFGVisitor:
     """An astroid visitor that creates a control flow graph for a given Python module.
 
+    Initialize with an options dict to configure the visitor.
+
+    Supported Options:
+      - "separate-condition-blocks": bool
+
     Private Attributes:
     _control_boundaries: A stack of the boundaries the visitor is currently in.
         The top of the stack corresponds to the end of the list.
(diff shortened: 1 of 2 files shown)


python_ta/cfg/visitor.py
```diff
@@ -8,20 +8,30 @@
 class CFGVisitor:
     """An astroid visitor that creates a control flow graph for a given Python module.
 
+    Initialize with an options dict to configure the visitor.
+
+    Supported Options:
+      - "separate-condition-blocks": bool
```
david-yz-liu, line 14: This is good, but underneath (and indented) please write a short English description of the purpose of this option.

python_ta/cfg/visitor.py
```diff
@@ -8,20 +8,30 @@
...
     _current_cfg: Optional[ControlFlowGraph]
     _current_block: Optional[CFGBlock]
     _control_boundaries: List[Tuple[nodes.NodeNG, Dict[str, CFGBlock]]]
 
-    def __init__(self) -> None:
+    def __init__(self, options: Optional[Dict[str, Any]] = None) -> None:
         super().__init__()
         self.cfgs = {}
+        if options is not None:
+            self.options = options
```
david-yz-liu, line 32: A more robust approach would be to define a class-level "default options" dict, and then *merge it* with the passed in options dict. This will ensure the options dict always has the expected keys.

david-yz-liu (approved):

Issue #882, opened 2023-03-14
check_all dynamic typing issues


skrukwa:
Hello,

I have noticed that things like `for` loops and `while` loops that change the types of variables can cause pyta to report incorrect warnings.

An example is as follows.
```python
def f() -> int:
    """while loop test function"""
    var = None
    while True:
        assert var is None or isinstance(var, tuple)

        if var is None:
            var = 1, 2
        else:
            return var[0]  # pyta reports E1136 (unsubscriptable-object) warning


if __name__ == '__main__':
    import python_ta
    python_ta.check_all()
```

I assume this is probably not an easy fix, and is probably already a known issue, but thought I would post it here formally. Feel free to do with thread as you see fit.

skrukwa:
On a side note, another (easier to fix) issue results from pyta not treating `raise` like `return`.

An example is as follows.

```python
def g(a: int) -> int:
    """if statements test function"""
    if a % 3 == 0:
        result = 3
    elif a % 2 == 0:
        result = 2
    else:
        raise NotImplementedError

    return result  # pyta reports E9969 (possibly-undefined) warning
```

david-yz-liu:
Thanks for reporting these issues. The first issue no longer is triggered (fixed by pylint) and the second has been fixed in #884.

PR #884, opened 2023-03-22
Fixed bug related to missing raise handling in CFG implementation

sushimon:
## Motivation and Context

These changes aim to fix the bug introduced in this issue's [comment](https://github.com/pyta-uoft/pyta/issues/882#issuecomment-1476670044). In addition, the current CFG implementation does not seem to handle `raise` statements (which should be similar to `return` statements) so these changes also fix this gap in the implementation.

## Your Changes

**Description**:
- `raise` statements are now being handled as if they were jumps (because they are very similar to a `return` statement).
- Added the [bug](https://github.com/pyta-uoft/pyta/issues/882#issuecomment-1476670044) as a test case.
- Added a new test case to ensure that `raise` statements are correctly being handled as a jump.

## Testing

- Ran the test suite and verified that they all passed.
- Added the bug as a test case and ensured that the previously buggy code is no longer flagged by the `possibly-undefined` checker.
- Added a new test case to verify that `raise` statements are now being handled as a jump. Verified that this new test passes.

- This bug was not the fault of the `possibly-undefined-checker`, but rather, the current CFG implementation did not support and handle `raise` statements correctly. Since they are very similar to `return` statements, I handled them in a similar manner and considered them a jump in these changes.
- Although this change fixes the original bug, there is a new bug involving `raise` statements inside `tryexcept` blocks. For example, the following snippet:
```
def f() -> None:
    try:
        raise NotImplementedError
    except NotImplementedError:  # CFG indicates it is unreachable
        pass
```
With this new change, the code above connected the `raise` statement directly to the end but in actuality, it should connect to the `ExceptHandler` block.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Fixed bug in possibly-undefined checker where a comprehension variable is falsely flagged as possibly undefined.
 - Fixed bug where `check_errors` and `check_all` opens a webpage when a nonexistent or unreadable path is passed as an argument.
+- Fixed the CFG implementation of pyta to resolve a bug in the possibly-undefined checker where variables were falsely flagged as possibly undefined when the code conditionally raises an exception and the variable was referenced afterwards.
 
 ### New checkers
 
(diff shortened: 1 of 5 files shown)


skrukwa:
I really don't know anything about how pyta works, but is it possible that the `raise` being under a `try except` would be another edge case where the raise might not behave like `return`. I have no idea, so up to your discretion.

david-yz-liu (approved):

PR #886, opened 2023-04-01
Updated raise handling in tryexcept

sushimon:
## Motivation and Context

The changes made in #884 introduced a new bug with code containing raise statements inside tryexcept blocks. These changes are meant to fix this bug.

## Your Changes

**Description**:

**NOTE: Builds on top of #884**
- Modifies `visit_tryexcept` so that it can link the raise statement with the corresponding `ExceptHandler` block if the same exception is raised and handled.
- Added new test cases to verify correctness of changes in different scenarios with raising an exception inside a tryexcept block.
- Updated the `CHANGLOG.md` file to document this bugfix.

## Testing

- Ran the test suite and verified the tests passed
- Generated a CFG with the previously buggy code and verified that the except block is not highlighted as unreachable
- Added test cases for the following scenarios: original buggy snippet with raise and except, overlapping exception clauses, and general catch-all (both when it handles the error and when a more specific clause handles it)

- Depends on changes made in #884. Do not merge this until after #884 gets merged to master and this branch gets merged/rebased.
- NOTE: there is a limitation where it won't detect any exceptions raised that are children of a parent exception class that is being handled
- NOTE: another limitation is that the cfg won't be able to identify raise statements where the expression is a variable or function call (i.e. `raise var` or `raise f()`). They will only identify them of the form `raise`, `raise <exception_class>`, or `raise <exception_instance>`

## Checklist

### DIFF
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -235,11 +235,15 @@ class CFGVisitor:
     ) -> None:
         old_curr = self._current_block
         for boundary, exits in reversed(self._control_boundaries):
+            # Linked specific Raise statements to corresponding except handler
             if isinstance(node, nodes.Raise) and f"{nodes.Raise.__name__} {node.exc.name}" in exits:
                 self._current_cfg.link(old_curr, exits[f"{nodes.Raise.__name__} {node.exc.name}"])
                 old_curr.add_statement(node)
                 break
-            elif type(node).__name__ in exits:
+
+            # Linking Raise to general catch-all except handler and other nodes to corresponding
+            # blocks
+            if type(node).__name__ in exits:
                 self._current_cfg.link(old_curr, exits[type(node).__name__])
                 old_curr.add_statement(node)
                 break
@@ -264,15 +268,20 @@ class CFGVisitor:
         end_block = self._current_cfg.create_block()
 
         after_body = []
-        for handler in node.handlers:
+        # Construct blocks in reverse to give precedence to the first block in overlapping except
+        # branches
+        for handler in reversed(node.handlers):
             h = self._current_cfg.create_block()
             self._current_block = h
             handler.cfg_block = h
 
-            # Iterate through all the exceptions this ExceptHandler handles, adding each one to the
-            # control boundary
-            for exception in _extract_exceptions(handler):
-                # Store both the 'Raise' and the exception class name
+            exceptions = _extract_exceptions(handler)
+            # Edge case: catch-all except clause (i.e. except: ...)
+            if exceptions == []:
+                self._control_boundaries.append((node, {nodes.Raise.__name__: h}))
+
+            # General case: specific except clause
+            for exception in exceptions:
                 self._control_boundaries.append((node, {f"{nodes.Raise.__name__} {exception}": h}))
 
             if handler.name is not None:  # The name assigned to the caught exception.
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -268,6 +268,7 @@ class CFGVisitor:
         end_block = self._current_cfg.create_block()
         # Case where Raise is not handled in tryexcept
         self._control_boundaries.append((node, {nodes.Raise.__name__: end_block}))
+        cbs_added = 1
 
         after_body = []
         # Construct blocks in reverse to give precedence to the first block in overlapping except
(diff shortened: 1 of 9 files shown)


david-yz-liu (commented):
@sushimon good work. You're running into a current limitation of the "control boundaries" because right now the "key" is the class name (`'Raise'`) but we need a way to distinguish between exception types for except blocks. However, your current approach breaks the natural recursive structure of the code (and modifies `_control_boundaries` a lot more than necessary).

A cleaner way of resolving this problem is to generalize the "control boundaries" so that they store both the '`Raise'` and the exception class name, and then add all exception handlers to `_control_boundaries` at the start, before recursing into the body statements.

So something like 

```python
try:
    ...
except ValueError:
    ...
except IndexError:
    ...
```

would cause `control_boundaries` to look like

```python
[('Raise ValueError', ...), ('Raise IndexError', ...)]
```

that does make `_visit_jump` more complicated, but I think overall it'll simplify the code.

python_ta/cfg/visitor.py
```diff
@@ -235,7 +235,11 @@ def _visit_jump(
     ) -> None:
         old_curr = self._current_block
         for boundary, exits in reversed(self._control_boundaries):
-            if type(node).__name__ in exits:
+            if isinstance(node, nodes.Raise) and f"{nodes.Raise.__name__} {node.exc.name}" in exits:
```
david-yz-liu, line 238: It's important to be careful when going deeper into AST structure. In particular, we shouldn't assume `node.exc` is a `Name` node or has a `name` attribute. Consider

```python
raise ValueError

raise ValueError(10)
```

This might require a helper function to be more general (and I think this can just default to `'Raise'` if the `exc` is of an unexpected type).

python_ta/cfg/visitor.py
```diff
@@ -253,17 +257,24 @@ def visit_tryexcept(self, node: nodes.TryExcept) -> None:
...
         for handler in node.handlers:
             h = self._current_cfg.create_block()
             self._current_block = h
             handler.cfg_block = h
+
+            # Iterate through all the exceptions this ExceptHandler handles, adding each one to the
+            # control boundary
+            for exception in _extract_exceptions(handler):
+                # Store both the 'Raise' and the exception class name
+                self._control_boundaries.append((node, {f"{nodes.Raise.__name__} {exception}": h}))
```
david-yz-liu, line 276: This is the right idea, but I realized there's a subtlety if we want to handle overlapping except branches:

```python
try:
    raise ValueError
except ValueError:
    ...
except ValueError:
    ...
```

In this case, it's the first block that should be "linked", not the second. I think this will be reversed in the current implementation because of the call to `reversed` in `_visit_jump`. Please try that out.

As a general comment, it would be great to add some additional test cases for the changes you're making, as they are fairly subtle!

david-yz-liu: Oh yeah, and have you tried a "bare except" (bad programming practice, but we should support it since it's valid Python code):

```python
try:
   ...
except:
   ...
```

That should act as a catch-all.

python_ta/cfg/visitor.py
```diff
@@ -292,3 +326,36 @@ def visit_with(self, node: nodes.With) -> None:
...
+    exception that this Raise node throws.
+
+    Preconditions:
+        - the raise statement is of the form 'raise' or 'raise <exception_class>'
+    """
+    exceptions = node.nodes_of_class(nodes.Name)
+
+    # Return the formatted name of the exception or the just 'Raise' otherwise
+    try:
+        return f"{nodes.Raise.__name__} {exceptions.__next__().name}"
```
david-yz-liu, line 359: You can call `next(exceptions)`

python_ta/cfg/visitor.py
```diff
@@ -281,6 +303,18 @@ def visit_tryexcept(self, node: nodes.TryExcept) -> None:
...
 
+        # Construct the try body so reset current block to this node's block
+        self._current_block = node.cfg_block
+
+        for child in node.body:
+            child.accept(self)
+        end_body = self._current_block
+
+        # Remove each control boundary that we added in this method
+        for _ in range(len(node.handlers) + 1):
```
david-yz-liu, line 314: This isn't the correct number if there are multiple exceptions for the same handler

```python
except (RuntimeError, TypeError, NameError):
     ...
```

(note that you have a `for exception in exceptions` loop above)

david-yz-liu (approved):

PR #893, opened 2023-05-09
Added example and documentation for W3301 nested-min-max

sushimon:
## Motivation and Context

The `nested-min-max` error is not documented in the pyta docs nor does it have an example showing what may cause this error. This PR aims to fix that.

## Your Changes

**Description**:

- Created an example that causes the `nested-min-max` error
- Updated the `docs/checkers/index.md` file to document this error

## Testing

- Ran python_ta on the example file to ensure that the intended error was being reported
- Ran python_ta on the corrected version of the example to ensure that it fixes the error
- Generated the docs to verify that the documentation was properly updated

## Checklist

### DIFF
--- a/docs/checkers/index.md
+++ b/docs/checkers/index.md
@@ -906,6 +906,24 @@ for student in students_info:
         cs_student_dict[student[0]] = student[1]
 ```
 
+(W3301)=
+
+### Nested min-max (W3301)
+
+This error occurs when there are nested calls of min or max instead of using a single min/max call.
+
+```{literalinclude} /../examples/pylint/w3301_nested_min_max.py
+
+```
+
+Corrected version:
+
+```python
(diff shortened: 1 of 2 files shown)


docs/checkers/index.md
```diff
@@ -906,6 +906,24 @@ for student in students_info:
         cs_student_dict[student[0]] = student[1]
 ```
 
+(W3301)=
+
+### Nested min-max (W3301)
+
+This error occurs when there are nested calls of min or max instead of using a single min/max call.
```
david-yz-liu, line 913: Use backticks around min and max to make them appear as `code font`, because they are the names of functions.

It would be good to provide more explanation here to explain why this is okay (you can pass in an arbitrary number of arguments to the functions).

david-yz-liu (approved):

PR #903, opened 2023-05-22
Modified the _is_constant_assignment function to allow assignment to type aliases

Bruce-8:
## Motivation and Context

This pull request extends the _is_constant_assignment function (from TopLevelCodeChecker) to allow not just assignment to "constant" variables like MY_CONST, but also to type aliases like MyType = list[int].

## Your Changes

**Description**: In the top_level_code_checker.py file, added an extra import statement to access the default pattern for the naming convention of type aliases and modified the _is_constant_assignment function to check for type aliasing assignments using the type alias pattern.

## Testing

- Used the Python Console to test re.match() on lists of string names of different naming styles to ensure that only string names that follow the type alias and constant naming styles are accepted.
- Added test cases to check that the function detects valid and invalid type alias assignments. 

Although it was mentioned in Slack that type aliases' naming convention is PascalCase, I noticed in the "Predefined Naming Patterns" section of this page (https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/invalid-name.html) requires type alias naming to be more strict, such as not allowing 'Type' to be the beginning of a name. I also found the exact strict pattern/rgx of type-aliases in this page (https://github.com/pylint-dev/pylint/blob/main/pylint/checkers/base/name_checker/checker.py#L40), and so used it in my code through an import statement.

## Checklist


david-yz-liu (commented):
@Bruce-8 good work. Three overall comments:

1. Change the name of the helper function (perhaps "`_is_allowed_assignment`")
2. Add test cases for your change
3. Update the Changelog

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Enhancements
 
 - Can now create control flow graphs to visualize the execution of your program.
+- The `TopLevelCodeChecker` can now check for the proper naming style usage for type alias assignment statements.
```
david-yz-liu, line 13: This is a good start but a bit misleading as written. Really the point of this check is that `TopLevelCodeChecker` now *allows* type alias assignment statements at the top level, i.e., *doesn't* flag such lines of code as an error

david-yz-liu (approved):

PR #906, opened 2023-05-24
Modify how cluster ids are generated to make them deterministic

sushimon:
## Motivation and Context

Cluster ids were previously generated by using object ids (via Python's `id` function) so they were not reproducible. This change opens up the ability to compare the content of `graphviz` files for testing.

## Your Changes

**Description**:

- Added a new instance attribute to `ControlFlowGraph` to keep track of this cfg's id.
- Updated `visitor.py` to initialize each CFGs idea using a state counter.
- Modified each subgraph to be named using the corresponding cfg's id instead of the node's object id.
- Added snapshot testing for the generated CFGs.
- Added `pytest-snapshot` as a dependency to the test environment.
- Modified cfg test case to no longer use the `tempfile` library but uses file fixtures instead.

## Testing

- Ran the test suite locally and verified all tests passed
- Created cfgs of code snippets (including ones with multiple modules within the same file) and verified that `cfg_count` is used to name each subgraph.
- Ran the new test case and verified it passed.

## Checklist


david-yz-liu (commented):
Hi @sushimon, I'd like these cfg "ids" to be a new (?) instance attribute of the `ControlFlowGraph` class itself, as we may want to use them for other purposes outside of `_display` in the future.

tests/test_cfg/snapshots/sample_cfg_output.txt
```diff
@@ -0,0 +1,25 @@
+digraph my_file {
```
david-yz-liu, line 1: Good! Usually for snapshot tests the convention is to name the file analogously to the input. So here that might be `my_file.gv` (see <https://en.wikipedia.org/wiki/DOT_(graph_description_language)>)

tests/test_cfg/test_cfg_generator.py
```diff
@@ -2,60 +2,55 @@
...
+    # Create the graphviz files using my_file.py
+    cfg.generate_cfg(mod=script_name, auto_open=False)
 
-    # Change directory to the temporary directory so the graphviz files will be created there
-    os.chdir(dir_name)
+    # Check if the graphviz files were created
+    assert os.path.isfile(dot_file_path) and os.path.isfile(svg_file_path)
 
-    return script.name, file_path, svg_file_path
+    # Teardown: remove the graphviz generated files
```
david-yz-liu, line 30: Make these separate assertions, as that makes the tests more granular.

tests/test_cfg/test_cfg_generator.py
```diff
@@ -2,60 +2,55 @@
...
+    os.remove(dot_file_path)
```
david-yz-liu, line 31: It's not ideal to put cleanup code at the end of a test body, since (for example) if an earlier assertion fails, the code won't run. Instead this is a good time to use a `yield` fixture: see https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization.

*Note*: the easiest way to use such a fixture is to move the logic of the specific filenames (`script_name`, `dot_file_path`, `svg_file_path`) into the fixture and pass them into the test case. This will also remove some code duplication across your two tests.

tests/test_cfg/test_cfg_generator.py
```diff
@@ -2,60 +2,55 @@
...
+    dot_file_path = os.path.splitext(os.path.basename(script_name))[0]
+    svg_file_path = dot_file_path + ".svg"
 
+    # Create the graphviz files using my_file.py
     cfg.generate_cfg(mod=script_name, auto_open=False)
 
-    # Check if the graphviz files were created
-    assert os.path.isfile(file_path) and os.path.isfile(svg_file_path)
+    # Open the actual graphviz file for reading
+    actual = open(dot_file_path, "r")
```
david-yz-liu, line 45: The `"r"` is redundant (default value) and can be omitted.

david-yz-liu (approved):

PR #908, opened 2023-05-28
Update UnnecessaryIndexingChecker to allow for node.target types other than AssignName

sarahsonder:
## Motivation and Context
The `_is_unnecessary_indexing` function in `UnnecessaryIndexingChecker` currently assumes that `node.target` is an `AssignName`. However, since `node.target` is of type `NodeNG | None`, it is not restricted to `AssignName`. My task was to update the checker to accommodate for different types of `node.target`s.

## Your Changes

1. I updated the helper function `_is_unnecessary_indexing` to identify all `AssignName` and `Name` nodes in `node.target` and then check whether those nodes are redundant.
2. I updated the functions `visit_for` and `visit_comprehension` to use the first `AssignName`/`Name` node in `node.target` as the `args` argument in the output message. This change was made because previously these two functions also assumed that `node.target` is of type `AssignName`.
3. I added a test where `node.target` is a `subscript` to demonstrate the functionality of the updated code.

**Description**:

## Testing

All the pytests pass and PythonTA also detects the unnecessary indexing with the new example.

I originally had an example where I converted my for loop test into a comprehension. However, since the target variable is a parameter of the function, the scope of the target variable (the function) doesn't match that of the comprehension and as a result, `_index_name_nodes` does not recognize instances of the target variables in the comprehension. I would be more than happy to implement any suggestions for examples that would run with comprehensions.

## Checklist


david-yz-liu (commented):
@sarahsonder you're definitely on the right track! Here are a few responses to your comments/questions:

1. For loop targets can also be subscript/attribute nodes, basically anything that can be on the left-hand side of a `=` in an assignment statement. This is bad practice, but possible.
2. However, I realized that you can also have nested loop targets, e.g. `for (..., ...), ... in ...`. So rather than enumerating all possible types of loop target, why don't you instead use the `nodes_of_class` recursive method, which will let you iterate over all `AssignName` nodes in the target, regardless of which type it is.
3. Regarding your test cases, my advice is to save your test code in files and try running PythonTA on each file. This way you can see whether the error is being detected, or not detected, successfully in each case.

sarahsonder:
Hi Prof. Liu! As per your suggestions, I have updated the `_is_unnecessary_indexing` function to use the `nodes_of_class` method, and the function now works as intended. I also realized that the reason why my previous test examples didn't work is because `_iterable_if_range` function in `UnnecessaryIndexingChecker` only accommodates for `range()` and not for `zip()`. Hence, I have created a new test that includes `range()` and targets node of type `subscript`. However, this is where I run into more problems:

1. The test code works properly when I save it in a separate file run it through the Python console, but PythonTA raises an astroid error (which is probably why the github tests failed as well). Do you know what could be causing this?
2. Although the syntax for `nodes_of_class` is `nodes_of_class(nodes.AssignName, nodes.Name)`, my test only works in pytest with `nodes_of_class(Union[nodes.AssignName, nodes.Name])` which I find to be odd.
3. In the `visit_for` function of `UnnecessaryIndexingChecker`, I need to accommodate the fact that `target` might not be `AssignName`, so similar to `_is_unnecessary_indexing`, I have changed the code to `nodes_of_class(nodes.AssignName, nodes.Name)`, but since `nodes_of_class` returns a generator object, I can't access the nodes directly. Is there a way I can access the nodes without using `__next__()`? Thank you.

sarahsonder:
@david-yz-liu Hi Prof. Liu! I just wanted to let you know that I have made the PR ready for review and I have also updated the problem description at the top. Thank you :)

david-yz-liu (commented):
@sarahsonder good job! In addition to the comments I left, please also make sure to update the Changelog to describe the change you made.

david-yz-liu (approved):

PR #909, opened 2023-05-29
Modified GlobalVariablesChecker to allow for type alias assignment

Bruce-8:
## Motivation and Context

This pull request extends the GlobalVariablesChecker functionality to allow not just assignment to "constant" variables like MY_CONST, but also to type aliases like MyType = list[int].

## Your Changes

**Description**: In the global_variables_checker.py file, added extra import statements to access the default pattern for the naming convention of type aliases as well as a pylint function to determine if a name is a python built-in name, and modified the _get_child_disallowed_global_var_nodes function to ignore type alias Name and AssignName nodes.

## Testing

-Ran PythonTA on sample lines of code, which were global variable assignment statements.
-Added test cases to check that GlobalVariablesChecker detects valid and invalid type alias assignments.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Can now create control flow graphs to visualize the execution of your program.
 - The `TopLevelCodeChecker` now allows type alias assignment statements at the top level, i.e., doesn't flag such lines
   of code as an error.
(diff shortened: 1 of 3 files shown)


david-yz-liu (commented):
@Bruce-8 good job, but it would also be good to handle cases like `datetime.date` where the type is imported from another module, and is not a built-in. I believe I mentioned the function `is_defined_in_scope`, please try using that for this purpose.

david-yz-liu (approved):

PR #911, opened 2023-06-04
Improved message display for C0303 trailing-whitespace

Bruce-8:
## Motivation and Context

This pull request solves two problems. The first problem was that error C0303 trailing-whitespaces wasn't appearing in the HTML report. The second was that after getting the trailing whitespaces to display, the entire erroneous line of code was highlighted, rather than the trailing whitespaces.

## Your Changes

**Description**: Modified test.pylintrc to allow error C0303 trailing-whitespaces to appear in the HTML report. Wrote a custom rendering function to allow only for trailing whitespaces to be highlighted, rather than the entire erroneous line of code. Modified html_reporter.py and core.py to make sure the line of code is not stripped of trailing whitespaces and to ensure the correct placement of highlighting on the line of code.

- [X] Other (please specify): The _colourify helper function in html_reported.py was modified to ensure that the highlighting placement was in the correct position.

## Testing

I have ran my code through PythonTA to ensure the HTML Report produces the desired outcome. 

I wasn't too sure whether I should add any test cases because I couldn't find a file nor folder that is designated for the type of change I made. I also went through the GitHub history of node_printers.py as well as the tests folder to see if any related test cases were made in the past. However, I wasn't able to find any.
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   of code as an error.
 - The `GlobalVariablesChecker` now allows type alias assignment statements at the top level, i.e., doesn't flag such
   lines of code as an error.
-- The HTML Report now displays the error message for C0303 `trailing-whitespace`, and only trailing whitespaces are
-  highlighted rather than the entire line of code.
+- For the message display of C0303 `trailing-whitespace`, trailing whitespaces are now appearing in the reporters,
+  and only those spaces are highlighted rather than the entire line of code.
 
 ### New checkers
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   of code as an error.
 - The `GlobalVariablesChecker` now allows type alias assignment statements at the top level, i.e., doesn't flag such
   lines of code as an error.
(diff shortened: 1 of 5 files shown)


tests/test_messages_config/test.pylintrc
```diff
@@ -124,8 +124,6 @@ pycodestyle-ignore =
     E711,  # pylint C0121
     E712,  # pylint C0121
     E722,  # pylint W0702
-    W291,  # pylint C0303
```
david-yz-liu, line 127: you shouldn't delete these, they are used to disable redundant checks from `pycodestyle` (it doesn't impact C0303)

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   of code as an error.
 - The `GlobalVariablesChecker` now allows type alias assignment statements at the top level, i.e., doesn't flag such
   lines of code as an error.
+- The HTML Report now displays the error message for C0303 `trailing-whitespace`, and only trailing whitespaces are
```
david-yz-liu, line 17: The first part isn't correct, the error message was always being triggered. (If you were having trouble triggering it it's likely due to having your IDE automatically strip trailing whitespace.)

This change also isn't just for the HTML reporter, but the colour reporter as well. So you don't need to specify the reporter type.

david-yz-liu (approved):
Thanks @Bruce-8! Regarding your question in the PR description, yes this is awkward to add test cases for, so I'm fine if you don't add them here.

PR #912, opened 2023-06-06
Add the option to render cfgs for functions only

sushimon:
## Motivation and Context

Currently, the pyta API only allows you to create a cfg for an entire module. This change now gives the user the ability to restrict the cfg creation to functions that they specify.

## Your Changes

**Description**:

- Updated `generate_cfg` to use the options dictionary to configure how it creates cfgs.
- Added new tests to verify that cfgs created with this newly added parameter are correct.
- Updated the pyta documentation to mention this new functionality.
- Updated the `CHANGELOG.md` file to mention this new functionality.

## Testing

- Ran the test suite and verified that the test cases pass.
- Ran the newly added tests and verified they passed as well.
- Used the pyta API to create sample cfgs using the new parameter and verified that the only cfgs created were the functions that were specified.
- Previewed the pyta documentation and the `CHANGELOG.md` file and verified my changes were now being displayed.

## Checklist


python_ta/cfg/visitor.py
```diff
@@ -68,7 +81,14 @@ def visit_classdef(self, node: nodes.ClassDef) -> None:
             child.accept(self)
 
     def visit_functiondef(self, func: nodes.FunctionDef) -> None:
-        self._current_block.add_statement(func)
+        # If user specifies to only render functions, check if the current function is listed
+        if self._functions_to_render is not None and func.name not in self._functions_to_render:
```
david-yz-liu, line 85: It would be good here to also handle methods using their qualified name (`Class.method_name`). From a `FunctionDef` node you can try to get the enclosing scope to see whether it's in a class. The "forbidden IO" checker does something similar.

python_ta/cfg/visitor.py
```diff
@@ -48,9 +54,16 @@ def __getattr__(self, attr: str):
...
     def visit_generic(self, node: nodes.NodeNG) -> None:
         """By default, add the expression to the end of the current block."""
-        self._current_block.add_statement(node)
+        if self._current_block is not None:
+            self._current_block.add_statement(node)
 
     def visit_module(self, module: nodes.Module) -> None:
+        # Modules shouldn't be considered if user specifies functions to render
+        if self._functions_to_render is not None:
+            for child in module.body:
```
david-yz-liu, line 63: The overall approach is good but a little more on the permissive side than the efficient side in terms of how it searches for functions.

1. When iterating over the module `body`, add a check so that we only recurse on `FunctionDef`s (with a matching name) and `ClassDef`s (to search for methods).
2. When iterative over a class def `body`, add a check so that we only recurse on `FunctionDef`s with a matching name.

In particular, you don't need to worry about finding `FunctionDef`s that are nested inside other types of syntax, e.g. if statements or even other `FunctionDef`s. It's fine to just look for function defs at the top-level of a module or class definition.

david-yz-liu (commented):
@sushimon good work. I left some comments, but also please make sure to update both the Changelog and the documentation page.

david-yz-liu (approved):

PR #913, opened 2023-06-11
Highlighted blank lines in between functions for the message display of E9989 pep8-errors

Bruce-8:
## Motivation and Context

In the PEP8 error check for the number of blank lines in between functions, the highlighted line is the line that the second function starts on, rather than the blank lines that precede it, which can be misleading. This pull request solves this problem by creating a new custom renderer that will highlight the blank lines in between functions to make it clear what the error is referring to.

## Your Changes

**Description**: Created a new custom renderer that will highlight the blank lines in between functions to make it clear what the error is referring to.

- [X] Other (please specify): Since this renderer is for all pep-8 error checkers, this change may modify the highlighting for other error messages in the reporters.

## Testing

I have ran my code through PythonTA to ensure the HTML Report produces the desired outcome.

I know that the specific PEP8 error checker for this situation is E302, but I believe that there is only one single PythonTA checker for all PEP8 errors. Thus, I wasn't sure how to customize this rendering function for only E302 because the 'msg' parameter only referred to a general PEP8 message (i.e., it's 'msg_id' attribute being E9989).
## Checklist

### DIFF
--- a/python_ta/reporters/node_printers.py
+++ b/python_ta/reporters/node_printers.py
@@ -141,7 +141,7 @@ def render_pep8_blank_lines_in_between_functions(msg, _node, source_lines=None):
         )
         yield from render_context(msg.line, msg.line + 2, source_lines)
     else:
-        render_generic(msg, _node, source_lines)
+        yield from render_generic(msg, _node, source_lines)
 
 
 CUSTOM_MESSAGES = {
--- a/python_ta/reporters/node_printers.py
+++ b/python_ta/reporters/node_printers.py
@@ -3,6 +3,8 @@ from enum import Enum
 
 from astroid import nodes
 
+NEW_BLANK_LINE_MESSAGE = "# INSERT NEW BLANK LINE HERE"
+
 
 def render_message(msg, node, source_lines):
     """Render a message based on type."""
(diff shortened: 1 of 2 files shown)


python_ta/reporters/node_printers.py
```diff
@@ -128,6 +128,19 @@ def render_missing_space_in_doctest(msg, _node, source_lines=None):
...
 
 
+def render_pep8_blank_lines_in_between_functions(msg, _node, source_lines=None):
+    """Render a PEP8 blank lines in between functions message."""
+    line = msg.line - 1
+    while source_lines[line - 1] == "\n":
+        line -= 1
+    yield from render_context(line - 1, line + 1, source_lines)
+    yield from (
+        (curr_line, slice(None, None), LineType.ERROR, " ")
```
david-yz-liu, line 138: Based on the point about visibility that was raised in Monday's meeting, one simple idea is to put in a larger number of spaces here (e.g. 20) for rendering.

david-yz-liu (commented):
@Bruce-8 great work. Regarding how to distinguish between the "number of blank lines" and other pycodestyle errors, there isn't an elegant way to do it. I'd recommend parsing the error message itself (from the `msg` object). You can call the default renderer in the case that some other pycodestyle message is being reported.

david-yz-liu (commented):
@Bruce-8 you've found the right idea to do the check, but as I was reviewing I realized there may be other messages from pycodestyle that contain the text "blank line".

Indeed, if you look at their [documentation page](https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes) under "E3" you'll see a few different ones. So, I'd like to expand the scope of this PR to include good rendering for *all* "blank line" messages that pycodestyle displays. This means you'll need to be able to reproduce each one to test; note that the lines may need to be highlighted differently.

Additionally, this could be a good place for some custom "instruction" strings, similar to the missing module/docstring custom renderers. Why don't you play around with that and try to figure out what looks good (and run your ideas by the PythonTA team).

python_ta/reporters/node_printers.py
```diff
@@ -128,6 +128,65 @@ def render_missing_space_in_doctest(msg, _node, source_lines=None):
...
+                    slice(None, None),
+                    LineType.ERROR,
+                    '"""INSERT NEW BLANK LINE HERE"""',
+                )
+                for _ in range(0, 2)
+            )
+        else:
+            line -= 1
+            yield from render_context(line - 1, line + 1, source_lines)
+            yield (line + 1, slice(None, None), LineType.ERROR, " " * 32)
```
david-yz-liu, line 163: You can make a helper like `render_blank_line` to pull out this repeated functionality.

david-yz-liu (commented):
Good work, I did some testing and found one case that isn't being properly highlighted:

image

(This is occurs for methods within a class, which pycodestyle expects just one blank line between.)

david-yz-liu (approved):

PR #915, opened 2023-06-14
Update number of error occurrences reported + Bug fix in HTML reporter

sarahsonder:
## Motivation and Context

Currently, PythonTA limits the maximum number of the times the same error can shown at 5 errors. However, students have requested for all error occurrences to be shown. Furthermore, there was a bug in the HTML reporter which displayed all error occurrences despite the reporter stating that only a limited number of occurrences were being shown.

## Your Changes
**Description**:
1. Changed the default value of `pyta-number-of-messages` to 0. This by default displays all instances of the same error.
2. Updated the behaviour of `plain_reporter` so that it limits the number of error occurrences ONLY when `pyta-number-of-messages` is not 0.
3. Added limiting behaviour to `html_reporter` so that its behaviour mirrors that of `plain_reporter` (previously `html_reporter` was not limiting the number of errors being shown).
4. Updated the conditions for "First <x errors> shown" to be displayed only when `pyta-number-of-messages` is not 0 and the number of error messages exceeds `pyta-number-of-messages`.
5. Fixed typo in the Configuration page of the PythonTA documentation.

## Testing

I tested my code by running both `plain_reporter` and `html_reporter` on three cases:
1. Setting `pyta-number-of-messages = 0` to check that all error occurrences are being shown
2. Setting `pyta-number-of-messages = 2` and running PythonTA on a file that has 5 of the same errors. This was to check that PythonTA was limiting the number of messages shown.
3. Setting `pyta-number-of-messages = 6` and running PythonTA on a file that has 5 of the same errors. This was to check that PythonTA would still display all error occurrences.

## Questions and Comments
I noticed that the [Configuration page](https://www.cs.toronto.edu/~david/pyta/usage/configuration.html?highlight=page) in the PythonTA documentation is under construction. Is there anything that I can add to the page?

## Checklist


CHANGELOG.md
```diff
@@ -45,6 +46,7 @@ list only contains the Pylint checkers enabled by default in PythonTA.
 - Fixed bug where `check_errors` and `check_all` opens a webpage when a nonexistent or unreadable path is passed as an argument.
 - Fixed the CFG implementation to resolve a bug in the possibly-undefined checker where variables were falsely flagged as possibly undefined when the code conditionally raises an exception and the variable was referenced afterwards.
 - Fixed bug where the generated CFGs will highlight the except block as unreachable if the same exception it is handling was raised in the body of the tryexcept.
+- Fixed bug where the HTML reporter would display all error occurrences of the same type despite stating that only a limited number was being shown.
```
david-yz-liu, line 49: You added this line in the incorrect place, it's under `[2.5.0]`. You can instead add a new `Bug fixes` section under `[unreleased]`.

python_ta/__init__.py
```diff
@@ -244,7 +244,7 @@ def reset_linter(
         (
             "pyta-number-of-messages",
             {
-                "default": 5,
+                "default": 0,
                 "type": "int",
                 "metavar": "<number_messages>",
                 "help": "Display a certain number of messages to the user, without overwhelming them.",
```
david-yz-liu, line 250: Please add a note here about what "0" means.

python_ta/config/.pylintrc
```diff
@@ -3,7 +3,7 @@
 # Make sure to register custom options tuple first in `python_ta/__init__.py`
 # ===========================================================
 # Default max amount of messages for reporter to display.
```
david-yz-liu, line 5: Add a similar comment here about what "0" means.

david-yz-liu (commented):
@sarahsonder great work! I just left a few minor comments, in addition to my earlier feedback of adding some test cases. 😄 

That said, I also realized that I said to use the JSON reporter for testing, but based on the changes you made, I'm not sure whether this reporter is actually using this configuration at all? If not, it would be great if you fixed that in this PR as well.

sarahsonder:
@david-yz-liu Hi Prof. Liu! I have updated the JSON reporter behaviour to limit the number of error messages and have added tests for my changes. The tests themselves work and the JSON reporter also working when I run PyTA on other files. However, the git tests are not passing due to a type error ('type' object is not subscriptable) caused by the helper function that I wrote in the JSON reporter. Could I please get some advice on how to resolve this error? Thank you!

sarahsonder:
@david-yz-liu Issue fixed! I needed to use List and Dict instead of list and dict.

python_ta/reporters/json_reporter.py
```diff
@@ -33,8 +33,36 @@ def display_messages(self, layout: BaseLayout) -> None:
...
+        all_msg_names = [msg.message.msg_id for msg in msgs]
+
+        unique_msg_names = []
+        [
+            unique_msg_names.append(msg_id)
+            for msg_id in all_msg_names
+            if msg_id not in unique_msg_names
+        ]
+
+        for msg_id in unique_msg_names:
```
david-yz-liu, line 55: This code is okay but is less efficient than it could be. Instead try this approach:

- loop over just `msgs`
- use a `dict` accumulator mapping error `msg_id` to the count of that message type "seen so far" in the loop. Note that you can initialize the accumulator using all of the unique keys (see above comprehension) with initial corresponding value 0
- then inside the loop use the accumulator to decide whether to append a message to `output_lst` or not
- after that loop, use a second loop over `output_lst` that sets the `number_of_occurrences` for each output message---we don't need to bother with updating this for messages that aren't included in the output!

david-yz-liu (approved):

david-yz-liu:
@sarahsonder I left a review but accidentally selected "Approve" instead of "Comment". So just to be clear, please address the review comments I left and then re-request a review. Thanks! 😄

python_ta/reporters/json_reporter.py
```diff
@@ -33,8 +33,28 @@ def display_messages(self, layout: BaseLayout) -> None:
...
+                    "msgs": self._output_messages(msgs),
                 }
             )
 
         self.writeln(json.dumps(output, indent=4))
+
+    def _output_messages(self, msgs: List[NewMessage]) -> List[Dict]:
+        """Returns a list of dictionaries containing formatted error messages."""
+        max_messages = self.linter.config.pyta_number_of_messages
+        unique_msg_names = {msg.message.msg_id for msg in msgs}
```
david-yz-liu, line 45: You're on the right track, but can simplify this further by using a dictionary comprehension instead of a set comprehension

david-yz-liu (approved):

PR #927, opened 2023-07-12
Removed code snippet in the error message for config file parsing errors

sushimon:
## Motivation and Context

From #920, it is apparent that pylint does not accurately display the lines containing the error for config file parsing errors. This PR aims to prevent the reporter from displaying the incorrect lines as the source of the error and simply show no code instead.

## Your Changes

**Description**:

- Added certain config file parsing error ids to `NO_SNIPPET` which prevents the reporter from displaying incorrect lines as the sources of error.
- Generalized some test cases to include the newly blacklisted config file parsing errors.
- Added a new test case to verify that no snippet was built for the blacklisted config file parsing errors.
- Updated `_override_config` to add E0015 to the reporter instead of printing to `stderr`.

## Testing

- Ran the test suite and verified they all passed.
- Ran the modified and new tests and verified they passed.

- ~~Did not include the `F0011` error in the updated tests since it seems to cause all other config file errors to be excluded. I ran the updated `test_config_parsing_errors` but included `F0011` in `CONFIG_ERRORS_TO_CHECK`, but after debugging, it seems to not report the `W0012` and `R0022` even though those errors are in the config file. Maybe since it's a "fatal" error, it stops pylint from checking for other errors?~~
- ~~`F0011` and `R0022` were the only errors that I was able to reproduce. The other errors whose id started with "00" were probably either command line or inline errors.~~
- `E0015` was not showing up in the reporter because the message was being printed to `stderr` instead inside of `_override_config`. I've updated it to now add the message to the reporter.

## Checklist

### DIFF
--- a/python_ta/__init__.py
+++ b/python_ta/__init__.py
@@ -248,7 +248,8 @@ def _override_config(linter: PyLinter, config_location: AnyStr) -> None:
     try:
         linter._parse_configuration_file(config_args)
     except _UnrecognizedOptionError as exc:
-        print(f"Unrecognized options: {', '.join(exc.options)}", file=sys.stderr)
+        unrecognized_options_message = ", ".join(exc.options)
+        linter.add_message("unrecognized-option", args=unrecognized_options_message, line=0)
 
     # Everything has been set up already so emit any stashed messages.
     linter._emit_stashed_messages()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Fixed bug in how PythonTA reports error messages that occur when parsing configuration files.
 - Fixed bug where the HTML reporter would display all error occurrences of the same type despite stating that only a limited number was being shown.
 - Fixed bug where the JSON reporter was not limiting the number of error occurrences displayed with respect to `pyta-number-of-messages`.
+- Ensured some config file parsing errors no longer display incorrect lines of code as the source of the error.
 
 ### New checkers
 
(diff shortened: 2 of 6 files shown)


david-yz-liu (commented):
@sushimon good work!

- For testing, you should create a separate test case and file for the config parse error.
- I think you missed a fewer errors that aren't in the "W" category: E0015 and R0022 at least.

sushimon:
> @sushimon good work!
> 
> * For testing, you should create a separate test case and file for the config parse error.
> * I think you missed a fewer errors that aren't in the "W" category: E0015 and R0022 at least.

In the latest commit I've:
- Created a separate test case for `F0011`. I've structured it so that non-fatal errors can be checked inside `test_config_parsing_errors` and fatal errors will have to be checked separately in their own individual test case and config file.
- `R0022` was already being checked in the very first commit but `E0015` has been added.

python_ta/__init__.py
```diff
@@ -244,11 +244,15 @@ def _override_config(linter: PyLinter, config_location: AnyStr) -> None:
         print(ex, file=sys.stderr)
         sys.exit(32)
 
+    unrecognized_options_message = None
     # Override the config options by parsing the provided file.
     try:
         linter._parse_configuration_file(config_args)
     except _UnrecognizedOptionError as exc:
-        print(f"Unrecognized options: {', '.join(exc.options)}", file=sys.stderr)
+        unrecognized_options_message = ", ".join(exc.options)
```
david-yz-liu, line 252: It's a bit strange to just set this variable here. You can instead call `linter.add_message` directly in this `except` block.

david-yz-liu (approved):

PR #928, opened 2023-07-12
New function validate-invariants

sarahsonder:
## Motivation and Context
To explicitly check that the representation invariants of an object are satisfied.

## Your Changes

**Description**:
Created a new function called `validate_invariants` which takes in an object `obj` and checks that the representation invariants of `obj` are satifised.

**Type of change**

## Testing

Created testing suite `test_validate_invariants.py`.

## Checklist


david-yz-liu (commented):
@sarahsonder you're definitely on the right track here! In addition to writing tests (of course), my main comment is you've correctly found some existing code on which to base you work. But now that you've done so, the best approach is to *refactor* your code to remove the duplication, so that the common code is used in existing place and in the new function you've added. This might mean defining new helper methods as appropriate.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - For the contract checking `new_setattr` function, any variables that depend only on `klass` are now defined in the
   outer function, efficiency of code was improved, and the attribute value is now restored to the original value if the
   `_check_invariants` call raises an error.
+- Can now check explicitly for whether the representation invariants of an object are satisfied.
```
david-yz-liu, line 32: You should name the function explicitly: "Added `check_invariants` function that...

python_ta/contracts/__init__.py
```diff
@@ -603,3 +583,37 @@ def _debug(msg: str) -> None:
...
+
+    setattr(klass, "__representation_invariants__", rep_invariants)
+
+
+def check_invariants(obj: object) -> None:
+    """Check that the representation invariants of obj are satisfied."""
+    klass = obj.__class__
+    klass_mod = _get_module(klass)
+
+    _set_invariants(klass)
```
david-yz-liu, line 618: Okay I thought about this and I want `check_invariants` to be purely non-mutating, and so it should only check invariants that have already been parsed/added, and *not* call `_set_invariants` here. So please remove this call, but you can keep the helper function refactoring you made.

This also has an implication for the test you added, since you'll need to add the `@check_contracts` decorator to your class. This means that you can't violate the representation invariant when you initially create the object (since an `AssertionError` will be raised), but instead will need to create a valid object then mutate it to violate an RI. I suspect doing something like `self.age -= ...` will work, but if not, you can create an attribute that's a mutable object and then *mutate* that attribute value's object to violate an RI, e.g. `self.something.append(...)`. PythonTA doesn't check RIs after one of its attribute objects is mutated.

tests/test_check_invariants.py
```diff
@@ -0,0 +1,41 @@
...
+        self.age = age
+
+
+def test_no_errors() -> None:
+    """Checks that check_invariants does not raise an error when representation invariants are satisfied."""
+    person_obj = Person("Jim", 50)
+
+    try:
+        check_invariants(person_obj)
+    except Exception:
```
david-yz-liu, line 32: Only catch errors that you're looking for: in this case, that would be `AssertionError`

tests/test_check_invariants.py
```diff
@@ -0,0 +1,41 @@
...
+        assert False
```
david-yz-liu, line 33: A better version of this is to use [`pytest.fail`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-fail)

david-yz-liu (commented):
@sarahsonder good work. I left some inline comments, but also make sure to add this new function to the contract checking documentation page.

tests/test_check_invariants.py
sarahsonder: Hi Prof. Liu! Regarding the error type, I realize that `_check_invariants` only raises `PyTAContractErrors`, whereas `_instance_method_wrapper` (which is called when setting attributes) raises `AssertionErrors`. Is there a way I could change the behaviour of  `check_invariants` or `_check_invariants` to raise `AssertionErrors`?

david-yz-liu: Good observation! Don't change the helper, but you should catch the PyTA error and then raise an AssertionError from the new function. Similar to what existing code does.

Think of the PyTA error as private, and AssertionError as the public error we should show users.

docs/contracts/index.md
```diff
@@ -48,10 +48,11 @@ AssertionError: divide argument '2' did not match type annotation for parameter
 
 ## API
 
-The `python_ta.contracts` module offers two functions for enabling contract checking.
+The `python_ta.contracts` module offers three functions for enabling contract checking.
```
david-yz-liu, line 51: You're on the right track here, but the documentation could be improved further. The new function `check_invariants` isn't used to *enable* contract checking, but rather to manually check the representation invariants of an object.

You can keep the generated docstring where it is, but move the intro blurb ("The third, `check_invariants`, ...") to directly above it, mimicking the style of how the constants `ENABLE_CONTRACT_CHECKING` etc. are introduced.

python_ta/contracts/__init__.py
```diff
@@ -603,3 +583,39 @@ def _debug(msg: str) -> None:
...
+                    _debug(
+                        f"Warning: representation invariant {assertion} could not be parsed as a valid Python expression"
+                    )
+                    continue
+                rep_invariants.append((assertion, compiled))
+
+    setattr(klass, "__representation_invariants__", rep_invariants)
+
+
+def check_invariants(obj: object) -> None:
```
david-yz-liu, line 613: Okay, so reviewing the documentation made me realize that `check_...` is actually a bad name for this function, because it's too similar to `check_contracts`, and they do fundamentally different things. I actually slightly regret choosing `check_contracts` as the name of the decorator, since that function is *enabling*, but not *checking*, RIs. But we're stuck with that.

Instead please rename this function to `validate`, which is another common verb used for this type of action.

tests/test_check_invariants.py
david-yz-liu: Now that you've made this change, you still need to change `Exception` to `AssertionError` here

david-yz-liu (approved):

PR #929, opened 2023-07-14
Improved the behavior of the contract checking new_setattr function

Bruce-8:
## Motivation and Context

For the `new_setattr` function in `python_ta/contracts/__init__.py`, this pull request re-organizes certain variable definitions, changes the use of certain `inspect` functions, and fixes a bug with attribute value restoration when representation invariants are violated.
## Your Changes

**Description**:
- Moved any variables that depend only on `klass`, but not `self/name/value`, to be defined in the outer function rather than inside `new_setattr`.
- Replaced `inspect.getouterframes` with the `f_back` to improve efficiency of code.
- The attribute value is restored to the original value if the `_check_invariants` call raises an error. 

## Testing

- Used an example in another file to ensure that `f_back` represents the frame that's currently being checked in the code.
- Added test case for the restoration of the attribute value to its original value if the `_check_invariants` call raises an error.
- Added test case for the attribute not existing if the `_check_invariants` call raises an error.

For the 10 tests that didn't pass in `test_class_contracts`, I believe it was due to my change of moving `ENABLE_CONTRACT_CHECKING` to the top of the function body, so I don't think there's a way to make this test pass unless I don't don't move the constant. Additionally, for the 6 tests that didn't pass in `test_contracts`, the tests all passed in PyCharm when I ran them on my own branch, and I'm not too sure why that's the case because I double checked that the tests were the same both here and on my branch.
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,10 +26,9 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - When running `check_contracts` on a class with type aliases as type annotations for its attributes, the `NameError`
   that appears (which indicates that the type alias is undefined) is now resolved.
 - The default value of `pyta-number-of-messages` is now 0. This automatically displays all occurrences of the same error.
-- For the contract checking `new_setattr` function, the check for `ENABLE_CONTRACT_CHECKING` is now at the top of the
-  function body, any variables that depend only on `klass` are now defined in the outer function, the use of
-  `inspect.getouterframes` is now changed to improve efficiency, and the attribute value is now
-  restored to the original value if the `_check_invariants` call raises an error.
+- For the contract checking `new_setattr` function, any variables that depend only on `klass` are now defined in the
+  outer function, efficiency of code was improved, and the attribute value is now restored to the original value if the
+  `_check_invariants` call raises an error.
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - When running `check_contracts` on a class with type aliases as type annotations for its attributes, the `NameError`
   that appears (which indicates that the type alias is undefined) is now resolved.
 - The default value of `pyta-number-of-messages` is now 0. This automatically displays all occurrences of the same error.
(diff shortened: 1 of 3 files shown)


david-yz-liu (commented):
@Bruce-8 good work. Revert the changes you made to the `ENABLE_CONTRACT_CHECKING` constant and keep everything else and I'll take another look. You can (afterwards) do that part in a separate PR.

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - When running `check_contracts` on a class with type aliases as type annotations for its attributes, the `NameError`
   that appears (which indicates that the type alias is undefined) is now resolved.
 - The default value of `pyta-number-of-messages` is now 0. This automatically displays all occurrences of the same error.
+- For the contract checking `new_setattr` function, the check for `ENABLE_CONTRACT_CHECKING` is now at the top of the
```
david-yz-liu, line 29: Update this to remove the reference to `ENABLE_CONTRACT_CHECKING`. Also you don't need to go into details about the efficiency improvements: reword to just say "improve efficiency" in the changelog entry.

python_ta/contracts/__init__.py
```diff
@@ -176,18 +176,21 @@ def new_setattr(self: klass, name: str, value: Any) -> None:
...
-        frame_locals = callframe[1].frame.f_locals
+        frame_locals = inspect.currentframe().f_back.f_locals
         if self is not frame_locals.get("self"):
             # Only validating if the attribute is not being set in a instance/class method
-            klass_mod = _get_module(klass)
             if klass_mod is not None and ENABLE_CONTRACT_CHECKING:
                 try:
                     _check_invariants(self, klass, klass_mod.__dict__)
                 except PyTAContractError as e:
+                    if original_attr_value is None:
```
david-yz-liu, line 190: This check doesn't work if the original attribute value was `None`, which is a different case from the original attribute note existing at all. You can use a separate variable to track whether there was an original attribute. And make sure to add a test for this case as well.

tests/test_contracts_attr_value_restoration.py
```diff
@@ -0,0 +1,19 @@
+from python_ta.contracts import check_contracts
+
+
+def test_class_attr_value_restores_to_original_if_violates_rep_inv() -> None:
```
david-yz-liu, line 4: Always make sure to add a docstring describing the purpose of each test case.

david-yz-liu (approved):

PR #930, opened 2023-07-19
Moved ENABLE_CONTRACT_CHECKING to top of function body.

Bruce-8:
## Motivation and Context

This pull request is made so that any contract or type checks are not executed if `ENABLE_CONTRACT_CHECKING` is False.
## Your Changes

**Description**: Moved the check for `ENABLE_CONTRACT_CHECKING` to the top of the function so that the `new_setattr` function doesn't execute any contract or type checks if the constant is False.

## Testing

- Added test case to ensure that no error is raised if a type check is violated (in the case that `ENABLE_CONTRACT_CHECKING` is set to False).
## Checklist


david-yz-liu (commented):
@Bruce-8 instead of indenting all of the existing code, you should use the *guard* pattern, which makes use of an early return:

```python
if not ENABLE_CONTRACT_CHECKING:
    ...
    return
```

The tests are likely failing due to `ENABLE_CONTRACT_CHECKING` being set to `False` in one test and then never being reset back to `True` in other tests. You should look up different ways of handling this type of situation robustly using `pytest`.

PR #931, opened 2023-07-21
Created InvalidNameChecker

sushimon:
## Motivation and Context

Pylint's `invalid-name` error explains how a variable name violates conventions but does so using regular expressions. Since beginners are unlikely to know what regular expressions are, this PR aims to "extend" the behaviour of the `invalid-name` checker by providing more beginner-friendly error messages.

## Your Changes

**Description**:

- Created a new checker.
- Created tests for this new checker.
- Updated the `CHANGELOG.md` file.
- Updated the documentation to include this new checker.
- Created examples for this new checker.
- Disabled `invalid-name` in the config file to avoid redundant checks.

## Testing

- Ran the test suite and verified they passed.
- Ran the newly added tests and verified they passed.
- Previewed the `CHANGELOG.md` and verified the newly added documentation is there.
- Generated the documentation and verified that the newly added section on the checker is there (with the example).

- ~~The failing test on GitHub Actions passes locally. The appropriate message gets added properly locally.~~

## Checklist


python_ta/checkers/invalid_name_checker.py
```diff
@@ -0,0 +1,525 @@
...
+        self._check_name("module", node.name.split(".")[-1], node)
+
+    @only_required_for_messages("pep8-name-violation")
+    def visit_classdef(self, node: nodes.ClassDef):
+        """Visit a Class node to check for any name violations.
+
+        Taken from pylint.checkers.base.name_checker.checker."""
+        self._check_name("class", node.name, node)
+        # Check any instance attributes that were first defined in this class.
+        for attr, anodes in node.instance_attrs.items():
```
david-yz-liu, line 363: Don't check this here. Instead, check only when instance attributes are "declared", which by our convention is in the class body as type annotations.

python_ta/checkers/invalid_name_checker.py
```diff
@@ -0,0 +1,525 @@
...
+
+        Snippets taken from pylint.checkers.base.name_checker.checker."""
+        if node.is_method():
+            if utils.overrides_a_method(node.parent.frame(future=True), node.name):
+                return
+
+        self._check_name(
+            _determine_function_name_type(node, config=self.linter.config), node.name, node
+        )
+        # Check argument names
```
david-yz-liu, line 379: Instead of putting this here, figure out what `visit_` method gets called on each parameter name, and do the check there. (`_recursive_check_names` isn't a very useful helper method, and should be removed)

python_ta/checkers/invalid_name_checker.py
```diff
@@ -0,0 +1,525 @@
...
+    def visit_functiondef(self, node: nodes.FunctionDef) -> None:
+        """Visit a FunctionDef node to check for any name violations.
+
+        Snippets taken from pylint.checkers.base.name_checker.checker."""
+        if node.is_method():
+            if utils.overrides_a_method(node.parent.frame(future=True), node.name):
+                return
+
+        self._check_name(
+            _determine_function_name_type(node, config=self.linter.config), node.name, node
```
david-yz-liu, line 377: What does ` _determine_function_name_type` do? I have a hard time imagining there are different cases we want to cover here.

docs/checkers/index.md
```diff
@@ -1118,6 +1118,29 @@ str                 sum                 super               tuple
 vars                zip
 ```
 
+(C9103)=
+
+### PEP8 name violation (C9103)
+
+This error is an extension of C0103 (invalid-name) which aims to provide a more detailed and beginner-friendly error message about how a variable name violated the PEP8 naming conventions.
```
david-yz-liu, line 1125: Don't reference C0103 here, just describe the checker as is. Then at the bottom of this entry you can add as a `*Note*` that this checker is intended to replace C0103.

python_ta/checkers/invalid_name_checker.py
```diff
@@ -0,0 +1,525 @@
...
+                    ):
+                        self._check_name("class_const", node.name, node)
+                        break
+                else:
+                    self._check_name("class_attribute", node.name, node)
+
+    def _check_name(self, node_type: str, name: str, node: nodes.NodeNG) -> None:
+        """Check for a name that violates the PEP8 naming conventions."""
+        name_check = NAME_CHECK[node_type]
+        # Update node_type to be more readable
```
david-yz-liu, line 474: It isn't necessary to do any string parsing, we don't need to follow what pylint does. Just pick the (human-friendly) name that you want for each node type and use it directly.

tests/test_custom_checkers/test_invalid_name_checker.py
```diff
@@ -0,0 +1,354 @@
...
+import astroid
+import pylint.testutils
+from astroid import nodes
+
+from python_ta.checkers.invalid_name_checker import InvalidNameChecker
+
+
+class TestInvalidNameChecker(pylint.testutils.CheckerTestCase):
+    CHECKER_CLASS = InvalidNameChecker
+    CONFIG = {
```
david-yz-liu, line 14: This points to a deeper problem. You shouldn't have to add a config option that isn't used anywhere in the class; find a better solution.

tests/test_custom_checkers/test_invalid_name_checker.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def test_module_name_violation(self) -> None:
+        """Test that the checker correctly reports an invalid module name."""
+        src = """
+        i = "hi"
+        """
+        mod = astroid.parse(src)
+        module_node, *_ = mod.nodes_of_class(nodes.Module)
+        module_node.name = "l"
+        print(module_node.name)
```
david-yz-liu, line 31: Remove all `print` calls

python_ta/reporters/core.py
```diff
@@ -44,6 +44,7 @@ def to_dict(self) -> Dict:
 NO_SNIPPET = {
     "invalid-name",
     "unknown-option-value",
+    "pep8-name-violation",
```
david-yz-liu, line 47: Overall it seems like we should actually be able to show the snippet, at least in every case except possibly the module name. In the module name case, you can pass in `node=None` and ensure that no snippet is shown when the node is `None`.

python_ta/checkers/invalid_name_checker.py
```diff
@@ -0,0 +1,525 @@
...
+
+
+def _is_in_upper_case_with_underscores(name: str) -> bool:
+    """Returns whether `name` is in UPPER_CASE_WITH_UNDERSCORES.
+
+    `name` is in `UPPER_CASE_WITH_UNDERSCORES if:
+      - each word is in uppercase, and
+      - words are separated by an underscore.
+    """
+    pattern = "([A-Z_][A-Z0-9_]*)$"
```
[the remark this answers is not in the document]
sushimon: I've tested a few examples of constants with leading underscores, i.e. `_CONSTANT`, and it already allows this

python_ta/checkers/invalid_name_checker.py
sushimon: It re-labels function definitions with certain decorators as attributes instead. I've updated to just classify each as either a method or a function.

python_ta/checkers/invalid_name_checker.py
sushimon: `visit_assignname` is called and we can use the existing variable `assign_type` to check if it was assigned in an `Arguments` node.

tests/test_custom_checkers/test_invalid_name_checker.py
sushimon: Removed the function that would cause the error so this shouldn't be a problem anymore

python_ta/reporters/core.py
sushimon: When I try to set `node=None` in the case of the module name, I get the following errors:

Here's the error that PythonTA reports.
image

Here's the error that Python reports.
image

docs/checkers/index.md
```diff
@@ -1118,6 +1118,39 @@ str                 sum                 super               tuple
 vars                zip
 ```
 
+(C9103)=
+
+### Naming convention violation (C9103)
+
+This error aims to provide a more detailed and beginner-friendly error message about how a variable name violated the PEP8 naming conventions.
```
david-yz-liu, line 1125: PEP8 -> Python

python_ta/checkers/invalid_name_checker.py
```diff
@@ -0,0 +1,476 @@
...
+    """
+
+    name = "naming_convention_violation"
+    # Template of displayed message will include the following:
+    #   1. Describe the variable name type (e.g. `module`, `const`, etc.) and the name itself.
+    #   2. Explanation of how the name violates the naming convention and suggest a
+    #   correction(s), if appropriate.
+    msgs = {
+        "C9103": (
+            'Found %s name "%s" in violation of Python naming conventions. %s',
```
david-yz-liu, line 286: In terms of the overall message: I like what you've done with the explanations, they're more detailed. Because of this, I don't think the first boilerplate sentence ("Found %s name ...") is necessary, since each message can be read standalone and clearly conveys the problem. By deleting the first sentence you'll make each error message a bit shorter, and therefore clearer.

Also, to make the above messages a bit clearer, change "is not in" to "should be in". For example, "Class name "{name}" should be in PascalCase format..."

python_ta/checkers/invalid_name_checker.py
```diff
@@ -0,0 +1,476 @@
...
+        bad_name_msg = _is_bad_name(name)
+        if bad_name_msg:
+            error_msgs.append(bad_name_msg)
+
+        name_length_msg = _is_within_name_length(node_type, name)
+        if name_length_msg:
+            error_msgs.append(name_length_msg)
+
+        for msg in error_msgs:
+            msg_id = (
```
david-yz-liu, line 436: Both this variable and `line_no` can be defined outside the for loop.

david-yz-liu (approved):

PR #948, opened 2023-09-15
Added option configuration to forbidden import checker

AinaMerch:
## Motivation and Context
This pull request provides additional documentation for the forbidden imports checker. It tells the users how to add optional configuration with examples, making it easier for them to use the checkers in the complete capacity.

## Your Changes

**Description**:
Added relevant documentation for the forbidden imports checker to add information about the optional configuration. Edited the changes based on the reviews.

## Testing
Generated the HTML document using the make html command

I think I have managed to revert all the "reformat code" changes back to the original. 

## Checklist


docs/checkers/index.md
```diff
@@ -1169,6 +1176,27 @@ lines: 1-3
 ---
 ```
 
+PythonTA allows you to specify all the modules that you wish to allow for a particular file using the
+allowed-import-modules configuration option in the check_all function
```
david-yz-liu, line 1180: Write the option name (here and below) in backticks (`) so that they are rendered in code font.

david-yz-liu, line 1180: Remove "in the check_all function" and write a colon after "option"

docs/checkers/index.md
```diff
@@ -1169,6 +1176,27 @@ lines: 1-3
...
+
+```python
+import python_ta
+python_ta.check_all(..., config={'allowed-import-modules': ["random"]})
+```
+
+You can specify any additional modules you want to allow for import using the extra-imports
+configuration option, which you can set in a call to python_ta.check_all or in a configuration file.
```
david-yz-liu, line 1188: as above, you can omit the part after the comma and just use a colon to introduce the subsequent example

docs/checkers/index.md
```diff
@@ -1169,6 +1176,27 @@ lines: 1-3
...
+
+```python
+import python_ta
+python_ta.check_all(..., config={'extra-imports': ["math", "tkinter"]})
+```
+
+```python
```
david-yz-liu, line 1195: It's great that you've considered this usage. Introduce this code example with a sentence or two. Note that both of the above options can be specified in this section, so your example should include both options.

david-yz-liu, line 1195: Also, the language for the configuration file is `toml` not `python`.

docs/checkers/index.md
```diff
@@ -1169,6 +1176,27 @@ lines: 1-3
...
+[FORBIDDEN IMPORT]
+extra - imports = math, tkinter
```
david-yz-liu, line 1197: There shouldn't be spaces around `-`

david-yz-liu (commented):
@AinaMerch good work. Regarding the reformatting, this is indeed not something you should ever do and so you'll need to revert these changes. There's no easy way to do it while also preserving the changes you want to keep (which is why you should not commit large or unnecessary changes), so my advice would be to copy the changes you want to keep into a separate file and then revert all changes.

Also, please add your name to the list of contributors at the bottom of the README.

david-yz-liu (approved):

PR #953, opened 2023-09-21
seperate __init__.py for handling config related functions

umututku03:
## Motivation and Context

In the `python_ta/__init__.py` file, we currently have functions that handle configuration-related tasks. However, we also have a dedicated `config` file within the `python_ta` directory, specifically designed for managing configuration-related features in PythonTA. It would be beneficial to organize our code in a more structured manner by placing a `__init__.py` file under the proper submodule (`python_ta.config`). This `__init__.py` file would encapsulate all configuration-related functions for PythonTA, enhancing code organization and clarity.

## Your Changes

**Description**:

Firstly, I created a new `.py` file under the submodule `python_ta.config` named `__init__.py`. Subsequently, I transferred all configuration-related functions, **along with the required import statements**, from `python_ta/__init__.py` to this new file. I also removed the unused import statements from `python_ta/__init__.py`, which were previously utilized by the aforementioned functions. Additionally, I noticed a minor typo in the `README.md` file and promptly corrected it.

## Testing

After applying these changes, I imported `python_ta` and called its user-utilized functions (e.g., `python_ta.check_all`) on multiple examples. Everything seems to be working smoothly!

I have two specific questions related to the code-base:

1. I'm curious about our approach to accessing protected members of certain modules (e.g., `pylint`). While Python allows flexibility in terms of "privacy", is this considered a best practice?
2. I've observed that our import statements are organized and ordered in a specific manner. Is there a particular convention that we are adhering to?

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Simeon Krastnikov,
 Ryan Lee,
 Hayley Lin,
 Bruce Liu,
-Merrick Liu
+Merrick Liu,
 Wendy Liu,
 Yibing (Amy) Lu,
 Shweta Mogalapalli,
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Simeon Krastnikov,
 Ryan Lee,
 Hayley Lin,
 Bruce Liu,
-Merrick Liu,
+Merrick Liu
 Wendy Liu,
 Yibing (Amy) Lu,
 Shweta Mogalapalli,
--- a/python_ta/__init__.py
+++ b/python_ta/__init__.py
@@ -22,10 +22,10 @@ import builtins
 from pylint.lint import PyLinter
 
 from python_ta.config import (
-    _find_local_config,
-    _load_config,
-    _load_messages_config,
-    _override_config,
+    find_local_config,
+    load_config,
+    load_messages_config,
+    override_config,
 )
 
 from .reporters.core import PythonTaReporter
--- a/python_ta/__init__.py
+++ b/python_ta/__init__.py
@@ -40,6 +40,7 @@ import os
 import sys
 import tokenize
 import webbrowser
+from builtins import FileNotFoundError
 from os import listdir
 from typing import AnyStr, Generator, List, Optional, TextIO, Union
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -88,7 +88,7 @@ Simeon Krastnikov,
 Ryan Lee,
 Hayley Lin,
 Bruce Liu,
-Merrick Liu
+Merrick Liu,
```
david-yz-liu, line 91: You shouldn't make this change in this PR since it's unrelated to the work you're doing. You would be welcome to make it in a separate PR!

python_ta/config/__init__.py
```diff
@@ -0,0 +1,101 @@
+""" Python Teaching Assistant - Configuration
+
+Within the config submodule, this .py file encompasses functions responsible
+ for managing all configuration-related tasks.
+"""
+__version__ = "2.6.2.dev"  # Version number
```
david-yz-liu, line 6: This part shouldn't be copied over from the `python_ta/__init__.py` file. Same with the `HELP_URL` and `sys.version_info` check a bit further below.

david-yz-liu (commented):
@umututku03 good work! In response your questions:

1. You're absolutely correct that we ideally shouldn't be import protected names from pylint; removing those imports would be a good future project.
2. I believe we're using [isort](https://pycqa.github.io/isort/) for the 
import order checking.

An *overall comment*, related to the question of public/private: now that you've moved the config helper functions into a separate file, let's make them public (by dropping the leading underscore).

python_ta/config/__init__.py
umututku03: Hi there @david-yz-liu , hope you are doing well. 

I've removed everything except the module docstring. I believe it's an effective way to inform users about the purpose of the .py file. What are your thoughts on this approach?

python_ta/__init__.py
```diff
@@ -19,10 +19,15 @@
 # First, remove underscore from builtins if it has been bound in the REPL.
 import builtins
 
-from pylint.config.config_file_parser import _ConfigurationFileParser
-from pylint.config.exceptions import _UnrecognizedOptionError
 from pylint.lint import PyLinter
 
+from python_ta.config import (
```
david-yz-liu, line 24: You should be able to do a relative import here, something like `from .config import ...` (https://www.geeksforgeeks.org/absolute-and-relative-imports-in-python/)

david-yz-liu (approved):

Issue #945, opened 2023-09-07
Allow import of student defined modules (E9999)


ch-iv:
Importing a student-defined module, such as any module located outside of the global python environment or the virtual environment should not raise an E9999 (forbidden-import). 

Students may choose to separate code into modules for code organization purposes. This does not violate the restriction for using external modules, since all code is written by the student.

The solution is to check the import directory location and conditionally raise E9999 if the module is not user defined or is a disallowed built-in module.

PR #959, opened 2023-09-25
Don't throw e9999 for local imports

ChristopherLi05:
## Motivation and Context

Addresses #945 

## Your Changes

**Description**:

I added the `get_local_files` method to `python_ta/checkers/forbidden_import_checker.py#ForbiddenImportChecker` which gets all of the python files in the same directory as the file that the linter is currently checking. I then modified `visit_import`, `visit_importfrom`, and `visit_call` to use this function so the checker won't throw an error when importing local files.

I updated `examples/custom_checkers/e9999_forbidden_import.py` to add the line `import e9999_forbidden_import_local  # No error`.

I created `examples/custom_checkers/e9999_forbidden_import_local.py` to demonstrate local file importing. 

The checker will no longer throw an e9999 error when importing a local file.

## Testing

I ran `python -m pytest tests` to make sure I didn't mess the preexisting tests up.

I then ran the following code to make sure that python_ta wouldn't throw an e9999 error on line 4.

```python
import python_ta
python_ta.check_all("examples/custom_checkers/e9999_forbidden_import")
```

Where is the test function/file for the `forbidden_import_checker`?

## Checklist


david-yz-liu (commented):
Hi @ChristopherLi05! Thanks for making this PR, and sorry for the veyr long delay in considering it.

Ultimately, the current behaviour of PythonTA is the preferred default. However, I would support *adding a new boolean configuration option*, `allow-local-imports`, which provides users the option of allowing local imports. If you're interested in pursuing this, you can take a look at the other custom configuration options PythonTA supports; see, for example, https://github.com/pyta-uoft/pyta/pull/973.

david-yz-liu (approved):
Thanks very much, @ChristopherLi05!

PR #964, opened 2023-09-29
Add config option to choose whether to overwrite Pylint error messages

vsahni3:
## Motivation and Context

Currently, all Pylint error messages are overwritten by custom Python TA error messages to make them more reader-friendly. However, in case the user wants to see the default Pylint messages instead, they have no flexibility. Now, users can simply add a config option if they don't want the Pylint messages to be overwritten.

## Your Changes

**Description**:
Added a new custom configuration option representing whether the Pylint error messages should be overwritten (default is True). Add logic to only overwrite the error messages when the config option is True. Also added documentation specifying the purpose and usage of  the new config option.

## Testing

I ran python_ta.check_all without the config option, with the config option as True, and with it as False. In all cases, the expected output was received. This was also to make sure that the checkers still worked and that the new changes didn't create bugs. I ran the test cases for pyta as well to ensure everything was working as expected.

If I have a pending pull request and made new changes to my fork, should I also pull in those changes when making a second pull request?

## Checklist

### DIFF
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -94,6 +94,19 @@ python_ta.check_all(..., output='pyta_output.txt')
 This options is compatible with all of PythonTA's reporter types, but we do not recommend its use with ColorReporter,
 as this reporter uses terminal-specific characters to colourize text displayed on your screen.
 
+## Overwrite Error Messages
+
+By default, PythonTA overwrites Pylint's error messages with its own to make them more reader-friendly. You can specify whether
+you want Pylint's default error messages instead using the `overwrite-error-messages` option. For example, use the following configuration to
+see Pylint's error messages:
+
+```python
+import python_ta
+python_ta.check_all(config={
+    "overwrite-error-messages": False
+})
+```
+
 ## Forbidden Imports
 
 By default, PythonTA has a list of modules that are allowed to be imported. You can specify any additional modules using the `extra-imports` configuration option, which you can set in a call to `python_ta.check_all` or in a configuration file.
--- a/python_ta/__init__.py
+++ b/python_ta/__init__.py
@@ -120,10 +120,12 @@ def _check(
     messages_config_path = linter.config.messages_config_path
     messages_config_default_path = linter._option_dicts["messages-config-path"]["default"]
     messages_config = _load_messages_config(messages_config_path, messages_config_default_path)
-
+    overwrite_error_messages = linter.config.overwrite_error_messages
     global PYLINT_PATCHED
     if not PYLINT_PATCHED:
-        patch_all(messages_config)  # Monkeypatch pylint (override certain methods)
+        patch_all(
+            messages_config, overwrite_error_messages
+        )  # Monkeypatch pylint (override certain methods)
         PYLINT_PATCHED = True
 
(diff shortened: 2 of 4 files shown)


david-yz-liu (commented):
@vsahni3 good work! As I was reviewing your PR, I was reminded that there's a subtlety with this feature I forgot to mention to you.

Not only does PyTA override default pylint error messages, it also provides instructors the ability to give a custom message file (the `messages-config-path` option). The current behaviour is to first merge the custom messages the user provides with the ones that PythonTA provides, then overwrite pylint.

With the new option I've asked you to add, it should disable the _PythonTA messages_ but *not* the custom ones provided by the user. (If the user just wants the pylint messages, they should not provide a custom message file and also set your new "overwrite" option to `False`.)

So in addition to the inline comments I made, the most substantive change here is to _implement_, _document_, and _test_ this behaviour formed by the interaction of the existing custom message option and the new configuration option you're adding. As a side benefit of this, you'll be contributing to the documentation page for this custom message option!

--

Unrelated, please add a new entry for this change in the `CHANGELOG` under a new "Enhancements" subsection.

tests/test_messages_config/test_messages_config.py
```diff
@@ -2,40 +2,85 @@
...
+    """Test the error messages given a valid messages-config-path with use-pyta-error-messages as True"""
     output = subprocess.run(
         [
             sys.executable,
             "-m",
             "python_ta",
             "--config",
-            "tests/test_messages_config/test.pylintrc",
-            "examples/pylint/e0111_bad_reversed_sequence.py",
+            "tests/test_messages_config/test1.pylintrc",
```
david-yz-liu, line 13: The test cases are all good, but you should pick more meaningful names for each configuration file. You can name them whatever you want, so pick names that suggest what test case they're being used for.

tests/test_messages_config/test_messages_config.py
```diff
@@ -2,40 +2,85 @@
...
+            "tests/test_messages_config/testing_code.py",
         ],
         capture_output=True,
         text=True,
     )
 
-    assert "This custom error message is modified." in output.stdout
+    assert (
+        "This custom error message is modified." in output.stdout
+        and "The first reversed() argument is not a sequence" not in output.stdout
```
david-yz-liu, line 22: In general do not combine boolean expressions in the same assert statement. It's okay here to use two different assert statements; this will give you more information if an assertion fails.

david-yz-liu (commented):
@vsahni3 thanks for updating everything! It all looks good, but I forgot to mention that you should *also* add to the CHANGELOG under "enhancements", since this is a new feature that's public-facing.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Enhancements
 
+- Added new configuration option `use-pyta-error-messages` to let users choose whether PythonTA should overwrite pylint's error messages.
+
+### Enhancements
```
david-yz-liu, line 14: There is a duplicate heading here (possibly the result of a merge conflict). You should combine your new entry with the one below.

david-yz-liu (approved):

PR #969, opened 2023-10-05
Fix `invalid-range-index` bug

### LINKED ISSUE #967: bug in invalid_range_index_checker
**Expected behaviour**: invalid_range_index_checker reports a problem when an index is empty or contains only one element.

**Actual behaviour**: invalid_range_index_checker reports a problem when it finds that the first parameter is less than the second parameter and the third parameter is positive in a range function.
image

I also wrote a test code containing `range(17,3,2)`, and PythonTA reported no problem. This should not happen either.

**Related code**: 
https://github.com/pyta-uoft/pyta/blob/166d7e1764ae6759e9d7149db949c605139680a2/python_ta/checkers/invalid_range_index_checker.py#L48-L49

The $\gt$ sign in line 48 and the $\lt$ sign in line 49 should be reversedtoo detect an empty range.

Read more in ed discussion: https://edstem.org/us/courses/39351/discussion/3556383

ch-iv:
This PR fixes a bug where `InvalidRangeIndexChecker` would incorrectly identify invalid range calls. This PR also implements tests to test the fix. Fixes #967

## Motivation and Context

It solves an issue some students in CSC110 were having with range calls.

image

The range produces more than 1 value, however it is flagged as invalid regardless.

## Your Changes

Implement a correct way of checking range calls.

**Description**:

## Testing

Tested with hypothesis for correctness. Also unit tests provided.

## Checklist


david-yz-liu (commented):
@ch-iv thanks for making this pull request! I left a few quick comments inline.

In addition to those:

1. Please update this branch from `master`; you'll find when merging that I've added a "Bug Fixes" subsection underneath "Unreleased", and you can add your CHANGELOG entry underneath that.
2. If you'd like to be recognized for your work, please add your name to the list of contributors in the README file :)

ch-iv:
@david-yz-liu Thanks for the feedback. I just pushed all the suggested fixes to the branch.

david-yz-liu (approved):
@ch-iv great work, thank you!

PR #970, opened 2023-10-06
Emphasizing Error-Inducing Code Chunks for Text-Based Reporters

umututku03:
## Motivation and Context

PythonTA currently supports various types of reporters, including `HTMLReporter` (which generates an HTML file) and text-based reporters like `PlainReporter` and `ColourReporter`. The HTML-based reporter effectively highlights the sections that cause errors. However, there's an issue with the `PlainReporter`. It generates a `.txt` file without any highlighting, making it difficult to distinguish between the code with errors and other content. Additionally, `ColourReporter` inherits from `PlainReporter`, necessitating adjustments to maintain its existing functionality while addressing this highlighting concern.

## Your Changes

**Description**:

I've overridden the `PlainReporter._add_line` method to incorporate an over-line character (Unicode 203E) beneath any section highlighted as an error. Additionally, I've overridden the `ColourReporter._add_line` method to ensure it inherits  the functionality of `PythonTaReporter._add_line`. This change allows us to maintain the existing behavior of `ColourReporter` while addressing the required enhancements (notice that `PlainReporter `is the parent class of `ColourReporter`).

## Testing

I did a lot of manual testings for this change. Below, I've included several screenshots showcasing the modifications I've implemented, along with corresponding comparisons across various types of reporters.

`ColourReporter`

image

`PlainReporter`

image

`HTMLReporter`

image

More examples 

image

image

image

image

Is it advisable to introduce an abstraction layer for text-based reporters? Considering the SOLID Principles we're currently studying in CSC207, it appears to align with the Dependency Inversion Principle. Would this interpretation accurately apply to our subject?

## Checklist


david-yz-liu (commented):
@umututku03 this is a great start. Some overall comments:

1. Make sure to update the changelog (add this entry as an "enhancement").
2. Right now the `PlainReporter._add_line` method is duplicating a substantial amount of code from the superclass `_add_line` method. The right way to think about this method is that it should _first_ call the superclass method (to build an initial snippet), and then add an additional line of text after that snippet that contains the underlining characters.
3. As a related point, right now you're using `len(snippet)` (and similar other code) to determine the amount of "pre-space" to use. Rather than doing that, you should calculate more precisely exactly how much space to be added without relying on the snippet at all! The implementation of `_add_line_number` gives you a starting point, as does the `slice_.start` value.
4. I suspect if you make the change from (3), it will also fix the issue you're seeing with the `ColorReporter`. While you are correct that dependency injection and an intermediate layer "text reporter" parent class would be possible (with subclasses filling in the "underlining" functionality), in this case I suspect there's a simpler approach involving updating the `PlainReporter._add_line` method.

david-yz-liu:
@umututku03 great, please also do a merge from `upstream master` (you'll see there's a merge conflict in the Changelog, which is normal and a good opportunity for you to practice resolving merge conflicts).

python_ta/reporters/plain_reporter.py
```diff
@@ -81,3 +82,33 @@ def _colour_messages_by_type(self, messages: Dict[str, List[NewMessage]]) -> str
...
+            ) * self._SPACE  # number 7 for prespaces included for adding line number
+            snippet_so_far = self._overline_helper(snippet_so_far, text[slice_], prespace)
+
+        elif linetype == LineType.DOCSTRING:
+            prespace = (7 + len(text) - len(text.lstrip(" "))) * self._SPACE
+            snippet_so_far = self._overline_helper(snippet_so_far, text.lstrip(" "), prespace)
+
+        return snippet_so_far
+
+    def _overline_helper(self, snippet: str, text: str, prespace: str) -> str:
```
david-yz-liu, line 105: Now that you've simplified this function a little, the `snippet` parameter really isn't necessary. You can just return the _new line_ and then add it to `snippet_so_far` in `_add_line`

python_ta/reporters/plain_reporter.py
```diff
@@ -81,3 +82,33 @@ def _colour_messages_by_type(self, messages: Dict[str, List[NewMessage]]) -> str
...
+        """
+        Helper method _add_line. Adds the Unicode203E (the overline character "‾") under any
```
david-yz-liu, line 107: you can call the character `Unicode U+203E`

python_ta/reporters/plain_reporter.py
```diff
@@ -81,3 +82,33 @@ def _colour_messages_by_type(self, messages: Dict[str, List[NewMessage]]) -> str
...
+    def _add_line(self, lineno: int, linetype: LineType, slice_: slice, text: str = "") -> str:
+        """Format given source code line as specified and return as str.
+
+        Called by _build_snippet, relies on _colourify.
+        """
+        snippet_so_far = super()._add_line(lineno, linetype, slice_, text)
+        if linetype == LineType.ERROR:
+            start = slice_.start or 0
+            prespace = (
+                7 + start
```
david-yz-liu, line 95: Overall we should avoid using magic numbers like this. It looks like `PythonTaReporter` already defines one constant (`_PRE_LINE_NUM_SPACES`); please add two more constants (for the number length and the number of spaces after that). Then you can use them in `PythonTaReporter` as well as here!

david-yz-liu (approved):

PR #975, opened 2023-10-29
fix limitation 2 for accumulation table and add tests

vsahni3:
## Motivation and Context

Currently, valid accumulators for Accumulation Tables are limited to variables. Accumulators that are part of an object, such as `self.sum_so_far`, cannot be used. This limitation has been removed so users now have more flexibility when choosing their accumulation variables. 
This pull request fixes [limitation 2](https://www.cs.toronto.edu/~david/pyta/debug/) for AccumulationTable.

## Your Changes

**Description**:
I modified `accumulation_table.py` so that the accumulator string passed in by the user is evaluated using the eval function if it is not found as a variable. If the variable cannot be found then a NameError will be raised.

## Testing

I made sure all the existing AccumulationTable tests worked with the new feature. I also added 3 new tests, 1 with the accumulator as an instance variable, 1 with a class variable, and 1 with an invalid accumulator. Finally, I did some manual testing with the cases described above.

## Checklist


david-yz-liu (commented):
@vsahni3 good work. Make sure to update the Changelog and remove this limitation in the documentation.

Also I merged in #977 so you should pull from master and resolve the merge conflict.

david-yz-liu (approved):

PR #977, opened 2023-11-07
Fixed limitation 4 of loop debugging

merrickliu888:
## Motivation and Context
Currently, the accumulation table cannot track variables that are initialized within the loop. It can only track variables initialized before the loop. See limitation 4 [here](https://www.cs.toronto.edu/~david/pyta/debug/index.html#current-limitations).

## Your Changes
I fixed this limitation by modifying the check made on accumulators in the `_record_iteration` method. Specifically I added an `elif` branch that checked if the `accumulator` was in `frame.f_code.co_varnames` or `frame.f_code.co_names`, which essentially checks if the variable name is in the code. The reason why I do both is because `co_varnames` is empty when `AccumulationTable` is set to track a top level loop. It should be noted that an interesting side of effect of this implementation is that variables declared after the loop can also be "tracked". Their value will just be "N/A" however.

I also moved the string `"N/A"` into a top level constant because it is not used in multiple places.

**Description**:

## Testing

I added three unit tests to the Accumulation Table test suite:
- The first test checks that, after the changes, we are able to track accumulator inside the loop. 
- The second test checks we are able to track accumulators initialized conditionally. 
- The third test checks that a Name Error is raised when the accumulation table is asked to track an uninitialized or invalid loop accumulator.

To check that the documentation updated, I built the new documentation with `Sphinx` and made sure the changes were reflected. I followed a similar process for checking if the `.md` files reflected their changes.

## Checklist


david-yz-liu (commented):
@merrickliu888 great start! Here are some comments.

1. Overall the function `get_variables_initialized_in_loop` is correct (and great use of `inspect` and `astroid` in combination), but isn't the simplest approach. The `inspect` module gives you a way to access *all* local variables within a function, not just the ones that are currently defined in the frame. The hint I'll give you is to start with `frame.f_code` and [go from there](https://docs.python.org/3/library/inspect.html#types-and-members).
2. The constant `'N/A'` is now used in multiple places, so should be pulled out and defined as an actual top-level constant.
3. Update the documentation page to remove the limitation.
4. Update the CHANGELOG to describe the change (it's an "Enhancement").
5. Good tests!

david-yz-liu (approved):

PR #984, opened 2023-11-30
Recursion table

vsahni3:
## Motivation and Context

This pull request contains a new feature for recursive tracing similar to AccumulationTable. It outputs a table of the inputs, return value, and caller inputs in order of the recursive calls. The table also maintains a tree representation of the recursive call which can be used for visualization purposes. 

## Your Changes

**Description**:
I created a new file for the RecursionTable class, one for the Tree class, and various tests to ensure the recursive table and corresponding tree structure are correct.

**Type of change**:

## Testing

I made an extensive test suite for various types of recursive functions and made sure to verify both the outputted table and the tree structure created. I also did manual testing in a playground to ensure the outputs and visualizations were as expected.

## Checklist

### DIFF
--- a/python_ta/debug/recursion_table.py
+++ b/python_ta/debug/recursion_table.py
@@ -4,6 +4,7 @@ for a recursive function.
 """
 from __future__ import annotations
 
+import copy
 import inspect
 import sys
 import types
@@ -22,7 +23,7 @@ def clean_frame_variables(frame: types.FrameType) -> dict[str, Any]:
     """
     raw_variables = frame.f_locals
     parameters = inspect.getargvalues(frame).args
-    cleaned_variables = {param: raw_variables[param] for param in parameters}
+    cleaned_variables = {param: copy.deepcopy(raw_variables[param]) for param in parameters}
     return cleaned_variables
 
 
@@ -115,7 +116,7 @@ class RecursionTable:
             return {}
         # intialize table columns using the first frame
         parameters = inspect.getargvalues(next(iter(self.frames_data))).args
-        recursive_dict = {key: [] for key in parameters + ["called by", "return value"]}
+        recursive_dict = {key: [] for key in parameters + ["return value", "called by"]}
 
         for frame in self.frames_data:
             current_frame_data = self.frames_data[frame]
--- a/python_ta/debug/recursion_table.py
+++ b/python_ta/debug/recursion_table.py
@@ -1,26 +1,27 @@
 """
 Table data structure that prints a nicely formatted table
-for a recursive function
+for a recursive function.
 """
 from __future__ import annotations
 
 import inspect
 import sys
 import types
-from typing import Any, Callable
+from typing import Any, Callable, Optional
 
 import tabulate
 
 from python_ta.util.tree import Tree
 
+DEFAULT_FUNCTION_STRING = "N/A"
+
 
 def clean_frame_variables(frame: types.FrameType) -> dict[str, Any]:
-    """remove the local variables from the frame's locals and keep only the
-    parameters
+    """Remove the local variables from the frame's locals and keep only the
+    parameters.
     """
     raw_variables = frame.f_locals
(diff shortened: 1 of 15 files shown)


python_ta/debug/recursion_table.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+
+    def _trace_recursion(self, frame: types.FrameType, event: str, _arg: Any) -> Callable:
+        """Trace through the recursive exexution and call the corresponding
+        method depending on whether a call or return is detected
+        """
+        # ignore the execution of the __exit__ method
+        if event == "call" and frame.f_code.co_name != "__exit__":
+            self._record_call(frame)
+        if event == "return":
```
david-yz-liu, line 147: prefer `elif` to consecutive `if`s when conditions are mutually exclusive

python_ta/util/tree.py
```diff
@@ -0,0 +1,20 @@
+"""
+Simple Tree class to be used for RecursionTable
+"""
+
+
+class Tree:
```
david-yz-liu, line 6: Add docstrings and type annotations for this class and all of its methods.

python_ta/util/tree.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, value):
+        self.value = value
+        self.children = []
+
+    def add_child(self, child_node):
+        self.children.append(child_node)
+
+    def check_tree_equality(self, tree1):
```
david-yz-liu, line 14: Name this method `__eq__` and then in the tests below use `==` to check whether the trees are equal.

python_ta/debug/recursion_table.py
```diff
@@ -0,0 +1,161 @@
...
+        frames_mapping: a mapping between the frame for a
+            recursive function and their return values and inputs
+        frames_ordered: a list to conserve the order in which the
+            frames were created
+        function_name: name of the function being traced
+        _trees: mapping of the frames to the corresponding tree
+            representing the function call
+    """
+
+    frames_mapping: dict[types.FrameType, list]
```
david-yz-liu, line 43: First, change the name of the attribute to `frame_data`.

More involved change: Instead of the dict values being lists, have them be dictionaries that store the individual pieces of data in the same structure as the eventual returned table, plus an extra field for the computed string representation of the call.

More concretely, the type annotation should be `dict[types.FrameType, dict[str, Any]]`. For the `fact` example each associated value should look like:

```python
{
  'n': ...,  # e.g., 4
  'return value': ...,  # e.g., '24'
  'called by': ...,  # e.g., 'fact(5)'
  'call string': ...  # e.g., 'fact(4)'
}
```

That is, move some of the logic in the `get_recursive_dict` method into the `_record_call` method to provide more structure to the data at the time when the call is initially recorded.

Finally, note that I've asked you to store `'call string'`; if you do so, you should be able to *look up previously-computed call strings* rather than recomputing them for every "parent" call string.

david-yz-liu (commented):
@vsahni3 great work. I've left some comments for you to work on. In addition to those comments, please draft a new section of the documentation for `RecursionTable`, similar to `AccumulationTable`.

python_ta/debug/recursion_table.py
```diff
@@ -0,0 +1,164 @@
...
+        current_frame_data["call string"] = current_func_string
+
+        self.frames_data[frame] = current_frame_data
+        self._insert_to_tree(current_func_string, frame, caller_frame)
+
+    def _record_return(self, frame: types.FrameType, return_value: Any) -> None:
+        """Update the state of the table representation after a function return is detected.
+        Note: the frame must already have been seen as returns are done 'on the way out'.
+        """
+        self.frames_data[frame]["return value"] = return_value
```
david-yz-liu, line 107: You'll probably want a `deepcopy` here as well

david-yz-liu (approved):

Issue #999, opened 2024-01-20
Add PyTA Documentation for W0134 return-in-finally Pylint Checker


MariaMa-GitHub:
- Figure out what it does, perhaps by experimenting and consulting the pylint documentation.
- Add an example file of the error (in the examples folder). Make sure when PythonTA is run on that file, the error is actually reported!
- Add an entry for the error in the checker documentation page (in the docs folder).

PR #1000, opened 2024-01-20
Added documentation and code examples for W0134 return-in-finally checker

MariaMa-GitHub:
## Motivation and Context

W0134 return-in-finally is a new check in Pylint which was added at the end of last semester, so it needed to be added to the PythonTA documentation. This pull request resolves the issue by providing documentation and examples for this checker.

Fixes pyta-uoft/pyta#999.

## Your Changes

**Description**:

- Added documentation for W0134 return-in-finally checker.
- Wrote a code example that raises the W0134 return-in-finally error.
- Provided a correction to the example code that demonstrates how to resolve the W0134 return-in-finally error.

## Testing

This pull request was tested using PythonTA with the code below. 

```python
import python_ta

python_ta.check_all()
```
The PythonTA report correctly showed that the W0134 (return-in-finally) error was raised.

I am not too sure how to classify the W0134 (return-in-finally) error. I chose to put it under "Improper Python usage" since I think that it does "indicate a misuse of variables, control flow, or other Python features in our code" and it is similar to other errors in the same section (such as Return outside function (E0104) and Unreachable (W0101)). However, when I ran PythonTA, it put the W0134 (return-in-finally) error under "Style and Convention", which also makes a bit of sense. Since I am not too sure how exactly this checker should be classified, I would like to ask under which section should it be put and in the future how should I decide if there are conflicts like this happening.

## Checklist

### DIFF
--- a/docs/checkers/index.md
+++ b/docs/checkers/index.md
@@ -22,16 +22,14 @@ This error occurs when a `return` statement is used inside a `finally` block.
 Corrected Version:
 
 ```python
-def error_code():
-
-    error_codes = [0, 1, 2]
+def error_code(error_codes):
 
     try:
-        return error_codes[2]
+        print(error_codes[0])
     except IndexError:
-        ...
+        return error_codes[1]
 
-    return error_codes[0]
+    return error_codes[2]
 ```
 
 (E0601)=
--- a/docs/checkers/index.md
+++ b/docs/checkers/index.md
@@ -9,34 +9,6 @@ email to \[david at cs dot toronto dot edu\].
 These errors generally indicate a misuse of variables, control flow, or other Python features in our
 code.
 
-(W0134)=
-
-### Return in finally
-
-This error occurs when a `return` statement is used inside a `finally` block. Doing so overwrites any previous return
-values therefore should be avoided.
-
-For example, in the code example below, if an `IndexError` occurs we want
-`error_codes[1]` to be returned, however since there is a return statement in the `finally` block `error_codes[2]` will be
-returned instead. Moving that return statement outside the `finally` block would resolve the issue.
-
-```{literalinclude} /../examples/pylint/w0134_return_in_finally.py
-
-```
-
-Corrected Version:
-
-```python
-def error_code(error_codes):
-
-    try:
-        print(error_codes[0])
-    except IndexError:
-        return error_codes[1]
-
-    return error_codes[2]
-```
-
 (E0601)=
 
 ### Used before assignment (E0601)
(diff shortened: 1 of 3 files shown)


sanrodarmawan98 (approved):

sanrodarmawan98 (approved):

sanrodarmawan98 (approved):

docs/checkers/index.md
```diff
@@ -488,7 +513,7 @@ This error occurs when a non-mapping value is used in a place where mapping is e
 
 ### Unnecessary negation (C0117)
 
-This error occurs when a boolean expression contains an nnecessary negation. If we are getting this
+This error occurs when a boolean expression contains an unnecessary negation. If we are getting this
```
david-yz-liu, line 516: This change is not related to this pull request, and so should not be included here.
 
However, this is a good catch, so what you could do it make a second, separate pull request that just fixes this issue!

docs/checkers/index.md
```diff
@@ -9,6 +9,31 @@ email to \[david at cs dot toronto dot edu\].
...
+
+```python
+def error_code():
+
+    error_codes = [0, 1, 2]
+
+    try:
+        return error_codes[2]
+    except IndexError:
+        ...
```
david-yz-liu, line 32: I would modify this example to capture a more typical "intended" behaviour as follows:

- Make `error_codes` a function parameter (in both the correct version and original example)
- Keep the return statement in the `except`
- Remove the return statement in the `try` (perhaps replace it with a `print` or something), and instead make the bottom return statement `return error_codes[2]`, which is what the "typical" behaviour would be.

david-yz-liu (commented):
Good work, @MariaMa-GitHub! I left a few inline comments, and in addition to that have two overall comments:

1. Since this is your first pull request, please add your name to the list of contributors :)
2. In terms of decided where to put the error, it's a bit ad hoc right now. One guiding principle I use is to look for similar errors, or errors that are related to the same language features. In this case, you should look for errors related to exceptions (some recent ones are the "too broad exception" ones) and put this entry close to those. Don't worry about categorizing as "style" vs. "error" within PythonTA though, that's something we can revisit later.

docs/checkers/index.md
MariaMa-GitHub: I just made the modifications and pushed the changes.

docs/checkers/index.md
```diff
@@ -9,6 +9,29 @@ email to \[david at cs dot toronto dot edu\].
 These errors generally indicate a misuse of variables, control flow, or other Python features in our
 code.
 
+(W0134)=
+
+### Return in finally
+
+This error occurs when a `return` statement is used inside a `finally` block.
```
david-yz-liu, line 16: It would be good to go into more detail about _why_ this is a problem (remember that the intended audience is students learning Python). You gave a great explanation in yesterday's meeting, and could write something similar here.

It's useful to use concrete examples to help with this. So, I would recommend that you refer to the example you created, e.g. "In the example below, ..."

docs/checkers/index.md
MariaMa-GitHub: It was clear, I just haven't had the change to do so. Now it is fixed and a separate pull request has been created.

docs/checkers/index.md
```diff
@@ -9,6 +9,34 @@ email to \[david at cs dot toronto dot edu\].
 These errors generally indicate a misuse of variables, control flow, or other Python features in our
 code.
 
+(W0134)=
+
+### Return in finally
```
david-yz-liu, line 14: Oh, I missed this last time but (1) you should add the error code at the send of the heading (to be consistent with other headings), and (2) make sure this is categorized under "Exceptions" to make it go with the other exception-related errors.

david-yz-liu (approved):

PR #1004, opened 2024-01-26
Enhancement: Extended Snapshot - Capturing Variables in '__main__' Stack Frame

umututku03:
## Motivation and Context

The current `snapshot` function gathers variables only from function calls, potentially nested. To enhance its utility, we aim to expand its capabilities, and our goal is to include variables in the top-level scope, specifically those defined in the main block. This expansion caters to a beginner-friendly Python context, providing a more comprehensive snapshot functionality.

## Your Changes

**Description**:

I've updated the snapshot.py module to improve the snapshot function. Now, it captures specific global variables, particularly those defined at the top level within the module where the snapshot function is called. The unit tests have been adjusted accordingly to validate and cover these improvements thoroughly.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactoring (internal change to codebase, without changing functionality)
- [X] Test update (change that modifies or updates tests only)
- [ ] Other (please specify):

## Testing

I've modified the unit tests for the snapshot function and tested the new feature across different levels of nesting. This ensures that the tests now align with the recent changes and thoroughly validate the functionality under various scenarios.

N/A

## Checklist


python_ta/debug/snapshot.py
```diff
@@ -14,6 +33,12 @@ def snapshot():
     local_vars = []
     frame = inspect.currentframe().f_back
 
+    # Handling possible errors
+    if frame is None:
+        raise RuntimeError("Cannot capture snapshot from top-level script.")
+
+    local_vars.append(get_filtered_global_variables(frame=frame))
```
david-yz-liu, line 40: You're on the right track here, but you can simplify the code by merging this part into the while loop. Think about why you insert the check on line 43 in the first place, and what you can do now. (This change will also remove the need for the "Handling possible errors" part above.

python_ta/debug/snapshot.py
```diff
@@ -15,6 +34,9 @@ def snapshot():
     frame = inspect.currentframe().f_back
 
     while frame:
+        if len(local_vars) == 0:  # to ensure we do not add main stack frame multiple times
```
david-yz-liu, line 37: This still isn't what I'm looking for. Look at the condition on line 40 and ask why it's there, and how you can use it to address the current problem.

umututku03:
Hi @david-yz-liu! I believe the necessity of the if condition was to only pinpoint local variables, and do not investigate global context. That's why I called the helper function after that check (and with the same indentation level with the if condition) so that the variables within the global context will be added to the accumulator, after the check is made.

python_ta/debug/snapshot.py
```diff
@@ -15,8 +34,14 @@ def snapshot():
     frame = inspect.currentframe().f_back
 
     while frame:
-        if frame.f_code.co_name != "<module>":
+        # whether the current stack frame corresponds to the global module context or not
+        if frame.f_code.co_name != "<module>":  # skips the global module context
```
david-yz-liu, line 38: Let me put this another way. Suppose this condition evaluates to `False`. Determine what should happen in that case, and use what you come up with to simplify your code.

umututku03: I agree with the proposed changes. Initially, I attempted to investigate global variables within the else branch using the helper. Although this method worked in the Python console and during informal testing, it failed to pass unit tests—similar to a previous issue.

In inspecting the global variables captured by the function within the else branch during unit tests, I got the following output:

`{'__main__': {'JB_DISABLE_BUFFERING': False, 'path': None, 'targets': ['test_accumulation_table.py::test_snapshot_one_level'], 'additional_args': [], 'joined_targets': ['test_accumulation_table.py::test_snapshot_one_level'], 'plugins_to_load': [<module 'teamcity.pytest_plugin' from 'C:\\Users\\semra\\AppData\\Local\\JetBrains\\Toolbox\\apps\\PyCharm-P\\ch-0\\233.13135.95\\plugins\\python\\helpers\\pycharm\\teamcity\\pytest_plugin.py'>], 'args': ['-ra', 'test_accumulation_table.py::test_snapshot_one_level', '--no-header', '--no-summary', '-q']}}`

Also, here's what's inside (using PyCharm debugger) the `<module>' frame when running with pytest, so else branch (i.e. when the condition evaluates to False) seems not valid:

image


This made me reconsider the current check's validity (`if frame.f_code.co_name != "<module>":`) and the need for an alternative. Although it's taking some time, my goal is a thorough understanding, not a hasty, brute-force solution.

david-yz-liu: @umututku03 okay that's good information to provide. Fundamentally, implementation should not be influenced by how the testing works. If the tests are failing with a logically sound implementation, then that points to a need to test code in a different way.

Now for this specific issue: the root of the testing problem is that when tests are run, `__main__` is a pytest module rather than the test file itself. That's why the global variables aren't showing up in the `<module>` stack frame. Unfortunately you can't really change how pytest runs.

So instead, you can adopt a different test strategy:

1. Create a separate Python file that contains your example, with the global variable.
2. Have that file print out the "main" stack frame as JSON.
3. In the test file, write a test case that _executes_ that module (you can use `subprocess` here), and captures the output.
4. Parse that output and then assert that the output is as expected.

Note that this is the strategy we use for some other test cases, e.g. in the `test_check.py` file. It's more cumbersome and we shouldn't use it without a good reason, but there is a good reason here. In this case, it's appropriate to leave a comment in the test docstring explaining why this approach is used.

tests/test_debug/snapshot_main_frame.py
```diff
@@ -0,0 +1,18 @@
...
+import json
+
+from python_ta.debug.snapshot import snapshot
+
+# globally defined variables
+team_lead = "David Liu"
+SDS_projects = ["PyTA", "MarkUs", "Memory Models"]
+team_num = 9
+
+global_vars = snapshot()[0]
```
david-yz-liu, line 16: Keep things simple and don't store this value in a variable (since it will also appear as a global variable), instead just convert and print it directly

tests/test_debug/snapshot_main_frame.py
```diff
@@ -0,0 +1,18 @@
...
+json.dumps(global_vars)
```
david-yz-liu, line 17: This function just returns a string, it doesn't print anything out. You should be doing `print(json.dumps(...))`

tests/test_debug/test_accumulation_table.py
```diff
@@ -481,6 +483,28 @@ def test_snapshot_three_levels() -> None:
...
+    Specifically, it verifies whether global variables (formatted as JSON) from another
+    file within the same module ("snapshot_main_frame.py") are accurately captured.
+    Subprocesses are utilized due to pytest's module configuration, where main is treated
+    as a pytest module rather than the test file itself, causing global variables to be
+    absent in the <module> stack frame. This behavior is inherent to pytest and cannot be modified.
+    """
+    current_directory = os.path.dirname(os.path.abspath(__file__))
+    snapshot_main_frame_path = os.path.join(current_directory, "snapshot_main_frame.py")
+    main_frame = subprocess.run(
+        ["python", snapshot_main_frame_path], capture_output=True, text=True
```
david-yz-liu, line 498: Don't use `"python"`, as this is also a hard-coded name. Instead look up `sys.executable`

tests/test_debug/test_accumulation_table.py
```diff
@@ -481,6 +483,28 @@ def test_snapshot_three_levels() -> None:
...
+    )
+    global_vars = main_frame.stdout
```
david-yz-liu, line 500: Parse this string as json, so that the `assert` below can use `==` on the underlying *dict* values.

david-yz-liu (approved):

PR #1010, opened 2024-02-01
Allowed importing specific names for ForbiddenImportChecker 

michaelZhaoCode:
## Motivation and Context

Currently ForbiddenImportChecker only allows the specification of modules to be allowed. There are cases when the user wants specific names (like functions or variables) from modules to be allowed but not the entire module. This pull request allows the user to allow specific names (e.g., math.floor) from modules to be imported in the existing two options, `allowed-import-modules` and `extra-imports`.

## Your Changes

Changed logic in ForbiddenImportChecker to allow for specific names to be imported in existing functions by specifying the full path, e.g. math.floor.  Also added a static helper function to convert the names from the ImportFrom nodes to be in the same format as the options. Changed the logic in the ImportFrom node handling to show a message only if the namebeing imported is not in the specified options. Changed error messages in PythonTA to specify if specific names were imported illegally.

**Description**:

## Testing

Added two new tests that both passed. One where the name being imported is allowed as specified by the config and produces no messages. One where multiple names being imported are not allowed but in the same module as another allowed name and produces a message containing both of them as a result.

## Checklist

### DIFF
--- a/python_ta/checkers/forbidden_import_checker.py
+++ b/python_ta/checkers/forbidden_import_checker.py
@@ -36,7 +36,7 @@ class ForbiddenImportChecker(BaseChecker):
                 "default": (),
                 "type": "csv",
                 "metavar": "<extra-modules>",
-                "help": "Extra names to be imported.",
+                "help": "Extra allowed names to be imported.",
             },
         ),
         (
@@ -135,7 +135,6 @@ def _get_full_import_names(modname: str, names: list) -> list:
     """Given a module name and a list of names imported from the module, return a list of strings
     in the form {module name}.{function name}.
 
-    modname and names are in the format as provided from the corresponding attributes in the pylint.ImportFrom node
+    modname and names are in the format as provided from the corresponding attributes in the astroid.ImportFrom node
     """
-
     return [modname + "." + name[0] for name in names]
--- a/python_ta/checkers/forbidden_import_checker.py
+++ b/python_ta/checkers/forbidden_import_checker.py
@@ -2,6 +2,7 @@
 """
 import os
 
+from __future__ import  annotations
 from astroid import nodes
 from pylint.checkers import BaseChecker
 from pylint.checkers.utils import only_required_for_messages
@@ -131,7 +132,7 @@ def register(linter: PyLinter) -> None:
     linter.register_checker(ForbiddenImportChecker(linter))
 
 
-def _get_full_import_names(modname: str, names: list) -> list:
+def _get_full_import_names(modname: str, names: list[tuple[str, str]]) -> list:
     """Given a module name and a list of names imported from the module, return a list of strings
     in the form {module name}.{function name}.
 
--- a/python_ta/checkers/forbidden_import_checker.py
+++ b/python_ta/checkers/forbidden_import_checker.py
@@ -15,7 +15,7 @@ class ForbiddenImportChecker(BaseChecker):
     name = "forbidden_import"
     msgs = {
         "E9999": (
-            "You may not import any modules - you imported %s on line %s.",
+            "You may not import %s.",
             "forbidden-import",
             "Used when you use import",
         )
@@ -27,7 +27,7 @@ class ForbiddenImportChecker(BaseChecker):
                 "default": (),
                 "type": "csv",
                 "metavar": "<modules>",
-                "help": "Allowed modules to be imported.",
+                "help": "Allowed names to be imported.",
             },
         ),
         (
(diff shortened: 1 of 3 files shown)


python_ta/checkers/forbidden_import_checker.py
```diff
@@ -48,6 +48,15 @@ class ForbiddenImportChecker(BaseChecker):
                 "help": "Allow local modules to be imported.",
             },
         ),
+        (
+            "allowed-function-imports",
```
david-yz-liu, line 52: Don't include a separate option; instead, both `allowed-import-modules` and `extra-imports` should allow both module names (e.g., `'math'`) and qualified functions (e.g., `math.floor`). Modify the `"help"` description of each of those options to refer to both modules and functions.

python_ta/checkers/forbidden_import_checker.py
```diff
@@ -77,6 +86,10 @@ def visit_importfrom(self, node: nodes.ImportFrom) -> None:
             node.modname not in self.linter.config.allowed_import_modules
             and node.modname not in self.linter.config.extra_imports
             and node.modname not in self.get_allowed_local_files()
+            and not all(
```
david-yz-liu, line 89: The overall logic here is good (though see my above comment about consolidating options), however I would indeed like to see the error message improved and be more specific, and so you can't just use an `all`.

First, the original message was misleading (`"You may not import any modules - you imported %s on line %s."`) because it's not that students aren't allowed to import any modules, just that they are restricted in what they can import.

Instead, please modify the message to just `"You may not import %s"` (removing the line number---since that shows in the context anyway), where `%s` can be replaced by either `"module <x>"` for a forbidden module, or `"<y> from module <x>"` if it's a specific name. Ideally `<y>` would be a comma-separated list of names, in the case that there are multiple imports on the same line that aren't allowed.

python_ta/checkers/forbidden_import_checker.py
```diff
@@ -119,3 +132,13 @@ def get_allowed_local_files(self) -> list:
 def register(linter: PyLinter) -> None:
     """Required method to auto register this checker"""
     linter.register_checker(ForbiddenImportChecker(linter))
+
+
+def _get_full_import_names(modname: str, names: list) -> list:
+    """Given a module name and a list of functions imported from the module, return a list of strings
```
david-yz-liu, line 138: I also realized when reviewing this that you can import all types of variables---not just functions---from modules. So a better description of this task, that should be reflected in all naming/documentation, is "allowed importing specific names for ForbiddenImportChecker". Similarly here say "list of functions" instead of "list of names"

david-yz-liu (commented):
@michaelZhaoCode good work. I left a few inline comments; additionally, because this is a user-facing change, it is appropriate to add an entry in the changelog under `[enhancements]`

python_ta/checkers/forbidden_import_checker.py
```diff
@@ -119,3 +129,13 @@ def get_allowed_local_files(self) -> list:
 def register(linter: PyLinter) -> None:
     """Required method to auto register this checker"""
     linter.register_checker(ForbiddenImportChecker(linter))
+
+
+def _get_full_import_names(modname: str, names: list) -> list:
```
david-yz-liu, line 134: About type annotations: at the top of this file, add the line `from __future__ import annotations`, and then here you'll be able to use the more specific type annotations.

python_ta/checkers/forbidden_import_checker.py
```diff
@@ -36,7 +36,7 @@ class ForbiddenImportChecker(BaseChecker):
                 "default": (),
                 "type": "csv",
                 "metavar": "<extra-modules>",
-                "help": "Extra allowed modules to be imported.",
+                "help": "Extra names to be imported.",
```
david-yz-liu, line 39: say "Extra allowed names to be imported."

python_ta/checkers/forbidden_import_checker.py
```diff
@@ -119,3 +129,13 @@ def get_allowed_local_files(self) -> list:
 def register(linter: PyLinter) -> None:
     """Required method to auto register this checker"""
     linter.register_checker(ForbiddenImportChecker(linter))
+
+
+def _get_full_import_names(modname: str, names: list) -> list:
+    """Given a module name and a list of names imported from the module, return a list of strings
+    in the form {module name}.{function name}.
+
+    modname and names are in the format as provided from the corresponding attributes in the pylint.ImportFrom node
```
david-yz-liu, line 138: It's the `astroid.ImportFrom` node (not `pylint`)

python_ta/checkers/forbidden_import_checker.py
```diff
@@ -119,3 +129,13 @@ def get_allowed_local_files(self) -> list:
...
+    """
+
```
david-yz-liu, line 140: delete this blank line

PR #1011, opened 2024-02-08
New PEP8 custom renderers for error E203 and E226

elescogna:
## Motivation and Context
In the HTLM page report, the lines that produced an error are highlighted in full; however, PythonTA uses custom renderers for some types of errors, including PEP8 errors, which allow to highlight only the problematic portion of a line of code, instead of the whole line. This allows to distinguish when different errors are present in the same line and address them separately.
This pull request serves two main purposes:

- Update the search criterion from error message to error code in the PEP8 custom renderer
- Add a custom renderer for some of the missing PEP8 errors

## Your Changes

**Description**:

- Update the function `render_pep8_errors` in `node_printers.py` to search the different PEP8 errors by error code instead of error message when calling the individual custom renderers
- Add the implementation of the custom renderers `render_pep8_errors_e226` and `render_pep8_errors_e203`, and add the relative function calls to `render_pep8_errors`.

## Testing

Initially, I tested my code by running PythonTA on `e9989_pycodestyle.py` and comparing the html page before and after my changes, one at a time, particularly:
- changes in `render_pep8_errors` in `node_printers.py`: I checked that the output was the same before and after my changes, and specifically that the highlighted portion was the same before and after my changes.
- addition of `render_pep8_errors_e226` and `render_pep8_errors_e203`: before adding this function, when a E226 or a E203 error was present the whole line was highlighted, so I checked that by calling the new function, only the portion of the line which caused the error was highlighted in the html report.

Later, I added unit tests example test files in `examples/custom_checkers/e9989_pycodestyle/` and unit tests in the file `tests/test_custom_checkers/test_pycodestyle_checker` for errors E226 and E203. Particularly I added:
- Test for error E203
- Test for no error E203
- Test for error E226
- Test for no error E226

### Questions
- Can I merge the custom renderers for errors E201, E202 and E203 into the same function `render_pep8_errors_e201_and_e202_and_e203` since they all share the same function body?

## Checklist


tests/test_custom_checkers/test_pycodestyle_checker.py
```diff
@@ -0,0 +1,46 @@
+import os
+
+import pylint.testutils
+from astroid.astroid_manager import MANAGER
+
+from python_ta.checkers.pycodestyle_checker import PycodestyleChecker
+
+DIR_PATH = (
+    os.path.abspath(os.path.join(__file__, "../../.."))
```
david-yz-liu, line 9: Defining this constant here is good. You don't need to use `abspath` (the relative from `__file__` is sufficient).

Also include the `examples/...` path inside the `os.path.join`; in general, all path segments should be combined together using `os.path.join`, not simple `+`. Note that you can write `'../../../examples/...'` in a path string.

tests/test_custom_checkers/test_pycodestyle_checker.py
```diff
@@ -0,0 +1,46 @@
...
+    + "/examples/custom_checkers/e9989_pycodestyle/"
+)
+
+
+class TestPycodestyleChecker(pylint.testutils.CheckerTestCase):
+    CHECKER_CLASS = PycodestyleChecker
+    CONFIG = {"pycodestyle_ignore": ["E121", "E123", "E126", "E24", "E704", "W503", "W504"]}
```
david-yz-liu, line 16: Since this will be combined with other tests for other pycodestyle errors, I think the most conservative thing to do is set `pycodestyle_ignore = []`. If for whatever reason an empty list doesn't work, pick just a single error to ignore.

tests/test_custom_checkers/test_pycodestyle_checker.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def test_error_e203(self) -> None:
+        """Test that PEP8 error E203 (whitespace before ‘,’, ‘;’, or ‘:’) is triggered"""
+        mod = MANAGER.ast_from_file(DIR_PATH + "e203_error.py")
```
david-yz-liu, line 20: Similar to my above comment, don't use `+`, instead use `os.path.join`

david-yz-liu (commented):
@elescogna great work. I left a few inline comments. And yes it's okay to combine the duplicated error functions into a single function in this PR (for duplicates of custom renderers that you're adding in this PR).

tests/test_custom_checkers/test_pycodestyle_checker.py
elescogna: An empty list causes pycodestyle to use the DEFAULT_IGNORE value, so I just left E24 in the list.

david-yz-liu (approved):

PR #1013, opened 2024-02-09
Created new PEP8 custom renders for E222 and E262

MariaMa-GitHub:
## Motivation and Context

In PythonTA reports there is the E9989 (pep8-errors) category, under which all pycodestyle (PEP8) style errors are placed. Since it is a generic error, the problematic code is entirely highlighted (complete lines) in the report. However, PythonTA has custom renderers that can highlight specific parts of the code, which helps users identify where the problem occurred. 

This pull request adds two of these custom renderers: E222 and E262.

## Your Changes

**Description**: I added custom renderers for error E222 and error E262.

## Testing

I tested the custom renders by running PythonTA on a problematic code file:

```python
def func(a: int, b: float) -> None:
    pass


def call_func() -> None:
    a =   1
    func(**{'a': 10, 'b': 15.2})  #  This works
    func(**a)  # Error on this line: non-mapping value 'a' used in a mapping context


import python_ta
python_ta.check_all()
```
The report that is generated after running this code confirms that I have correctly written the custom renderers as I do see the expected code fragments being highlighted.

## Checklist


david-yz-liu:
@MariaMa-GitHub FYI I merged in #1011, so you should do a merge from master; there will be some merge conflicts to resolve, but those should be pretty minor. Also you can modify your code to check for the pycodestyle error code rather than the string now.

PR #1021, opened 2024-03-02
snapshot_to_json step1

umututku03:
## Motivation and Context

Currently, our process for extracting snapshot data does not align seamlessly with the requirements of our `memory-models-rough` library. This discrepancy hinders our ability to leverage snapshot data effectively within memory models, limiting the potential insights and optimizations we can achieve. To bridge this gap, we propose the development and integration of a new function, `snapshot_to_json`, specifically designed to convert snapshot outputs into a JSON format that is directly consumable by the memory-models-rough module. This initiative not only aims to streamline our data processing pipeline but also enhances the synergy between different components of our system, unlocking new possibilities for data analysis and model performance. We have segmented this task into three distinct phases, with this pull request specifically targeting the initial phase: processing stack frames that contain solely primitive data types.

## Your Changes

**Description**:

The code transforms snapshot data into a JSON format, uniquely identifying each primitive value with a global ID. It iterates through snapshot data, assigning IDs to variables and their values, avoiding duplicates by checking against a global ID registry. Each variable's value is stored in a separate entry, while stack frames are represented as objects with metadata indicating their structure. The process culminates in merging these entries into a single JSON document, formatted for readability. We are operating under the assumption that the snapshot data exclusively contains primitive data types.

## Testing

I have incorporated a unit test that evaluates the new function's behavior across various scenarios, including aliasing, to ensure comprehensive coverage.

N/A

## Checklist

### DIFF
--- a/python_ta/debug/snapshot.py
+++ b/python_ta/debug/snapshot.py
@@ -2,13 +2,13 @@
 Use the 'inspect' module to extract local variables from
  multiple stack frames. Useful for dynamic debugging.
 """
+from __future__ import annotations
+
 import inspect
-import json
 from types import FrameType
-from typing import Dict, List
 
 
-def get_filtered_global_variables(frame: FrameType) -> Dict:
+def get_filtered_global_variables(frame: FrameType) -> dict:
     """
     Helper function for retriving global variables
     (i.e. top level variables in "__main__" frame's scope)
@@ -48,7 +48,7 @@ def snapshot():
     return variables
 
 
-def snapshot_to_json(snapshot_data: List[Dict]):
+def snapshot_to_json(snapshot_data: list[dict]) -> list[dict]:
     """
     Convert the snapshot data into a simplified JSON format, where each primitive value
     has its own entry with a matching ID.
@@ -66,10 +66,11 @@ def snapshot_to_json(snapshot_data: List[Dict]):
                 if var_id not in global_ids:
                     global_ids[var_id] = id_counter
                     var_id_diagram = id_counter
+                    id_counter += 1
                 else:
                     var_id_diagram = global_ids[var_id]
                 frame_variables[var_name] = var_id_diagram
-                id_counter += 1
+
                 # Create a separate entry for the variable's value
                 value_entry = {
                     "isClass": False,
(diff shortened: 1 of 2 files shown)


python_ta/debug/snapshot.py
```diff
@@ -3,10 +3,12 @@
  multiple stack frames. Useful for dynamic debugging.
 """
 import inspect
+import json
 from types import FrameType
+from typing import Dict, List
```
david-yz-liu, line 8: You don't need these type annotations. Instead, add the import `from __future__ import annotations` at the top of this file, and you can then use the lowercase versions of `list` and `dict`

python_ta/debug/snapshot.py
```diff
@@ -44,3 +46,50 @@ def snapshot():
         frame = frame.f_back
 
     return variables
+
+
+def snapshot_to_json(snapshot_data: List[Dict]):
```
david-yz-liu, line 51: Add a return type here. 

But also, return a `dict`, not a `str`. You don't need to use `json.dumps` here, it's a bit more useful to just have the `dict` and then let the user decide what to do with it.

david-yz-liu (commented):
@umututku03 good job, I just left a few minor comments.

python_ta/debug/snapshot.py
umututku03: Hello @david-yz-liu, thank you for your feedback. I'm currently considering the rationale behind returning a dictionary in our implementation. Initially, I've been accumulating the objects/frames to be drawn in a list. Transitioning to a dictionary prompts the question: what would be the appropriate key for each corresponding value?

python_ta/debug/snapshot.py
umututku03: Hello @david-yz-liu, Thank you for providing feedback. I implemented these changes in response to certain tests failing on Python version 3.8. Do you think it would be advisable to maintain these modifications? 

image

python_ta/debug/snapshot.py
david-yz-liu: Oh yes, you can return a `list`, that's good. I meant more that you shouldn't return a string.

python_ta/debug/snapshot.py
david-yz-liu: @umututku03 do what I've said to do and the test should not be an issue

david-yz-liu (approved):

Issue #998, opened 2024-01-16
Ignore C9103 For Unit Tests (Function Name Exceeds Max Characters [30])


klokailo:
Common python convention may include unit test functions with really long names (since there are many of them).

This is a suggestion to automatically ignore C9103 for all functions beginning with `test_`.

PR #1022, opened 2024-03-04
Issue 998 - Ignore naming convention violation option

elescogna:
## Motivation and Context
Issue 998: Ignore C9103 for unit tests suggests to ignore C9103 (name length exceeds 30 characters) for unit tests.
This pull request provides more flexibility by adding a configuration option to the `InvalidNameChecker` to ignore any names that match the pattern of a regular expression provided by the user.

 https://github.com/pyta-uoft/pyta/issues/998

## Your Changes

**Description**:
I added two new options to the `InvalidNameChecker`:
- `ignore-names`: takes a list of regular expression, and allows to ignore any name that matches the regex pattern
- `ignore-module-name`: takes a list of regular expressions, and allows to ignore any module name that matches the regex pattern

I implemented the logic for this change by adding a condition in every `visit_<node>` method checking whether the node's name matches any of the regular expressions provided in the relevant option.

## Testing

I added a new configuration as well as four additional unit tests to the `TestInvalidNameCheckerClass` to test the changes I made. 
The configuration I added to the class contains regular expressions that match the names used in the new four tests: `test_ignore_function_name`, `test_ignore_class_name`, `test_ignore_variable_name` and `test_ignore_module_name` all check that otherwise invalid names do not cause an error because they match the patterns provided in `ignore-names` and `ignore-module-names`.

Finally, I made sure that the regex I provided matched _only_ the names used in the new tests, to ensure that all previous tests passed after my changes. This also verifies that the naming convention is in fact ignored _only_ for the names that match the regex provided.

**Comment**:
As of now, to guarantee that the checker will work as expected, it is necessary to provide the regex in a list. When I tried to set the configuration to `options = {"ignore_names": "(test_[0-9a-z_]*)$"}` in the tests the regex library raised an error, but did not do so when I set it from the Python Console. Setting the configurations to `options = {"ignore_names": ["(test_[0-9a-z_]*)$"]}` solves the problem.

**Question**:
Should I simply cast the configurations to a list before passing it to `re.match`, or is it okay that only a list will be accepted?

## Checklist


python_ta/checkers/invalid_name_checker.py
```diff
@@ -291,20 +296,42 @@ class InvalidNameChecker(BaseChecker):
             "Used when the name doesn't conform to standard Python naming conventions.",
         ),
     }
+    options = (
+        (
+            "ignore-names",
+            {
+                "default": (),
+                "type": "csv",
```
david-yz-liu, line 304: Instead use `regexp`; see, for example, https://github.com/pylint-dev/pylint/blob/6a09d298b2313ce143d986502b1d4d2dcd872b18/pylint/checkers/variables.py#L1230-L1237. I think you can set the default to be an empty string. This means the user should specify just a single regular expression rather than a list of regular expressions, but I think that makes sense.

Same for the other option.

python_ta/checkers/invalid_name_checker.py
```diff
@@ -291,20 +296,42 @@ class InvalidNameChecker(BaseChecker):
...
+                "metavar": "<ignore-matching-names>",
+                "help": "Ignore C9103 naming convention violation for names that exactly match the patterns",
```
david-yz-liu, line 306: "patterns" -> "pattern"

python_ta/checkers/invalid_name_checker.py
```diff
@@ -327,6 +355,10 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
         frame = node.frame(future=True)
         assign_type = node.assign_type()
 
+        # Do not check this node if included in the ignore-names option
```
david-yz-liu, line 358: You can move this guard to the very top of the function

david-yz-liu (approved):

python_ta/checkers/invalid_name_checker.py
elescogna: I changed this and made other related changes. I had to include`re.compile()` in the config for the test class because the string is not turned automatically into a `re.Pattern` object as it happens in the checker class directly.

david-yz-liu (approved):
Good job, @elescogna!

PR #1023, opened 2024-03-08
Fixed crashes at PossiblyUndefinedChecker #981

### LINKED ISSUE #981: Check crashes at PossiblyUndefinedChecker
When running a check on the following snipped the program crashes
```python
for i in range(10):
    def foo():
        break
```
giving the following error message:

```
Exception on node <FunctionDef.foo l.3 at 0x25d06000650> in file 'C:\Users\[redacted]\PycharmProjects\pyta2\generated_code.py'
  File "C:\Users\[redacted]\PycharmProjects\pyta2\venv\Lib\site-packages\pylint\utils\ast_walker.py", line 91, in walk
    callback(astroid)
  File "C:\Users\[redacted]\PycharmProjects\pyta2\python_ta\checkers\possibly_undefined_checker.py", line 44, in visit_functiondef
    self._analyze(node)
  File "C:\Users\[redacted]\PycharmProjects\pyta2\python_ta\checkers\possibly_undefined_checker.py", line 58, in _analyze
    cfg.start = node.cfg_block
                ^^^^^^^^^^^^^^
AttributeError: 'FunctionDef' object has no attribute 'cfg_block'
```
An error is first raised at `transforms.patch_ast_transforms`, but it is caught by a try\catch block. After that the checks like `possibly_undefined_checker` keep running despite the failed transform. I think this is the reason why `possibly_undefined_checker` raises an error causing the program to crash. 

I think the error can be fixed by excluding some checks from running if the transform fails.

MariaMa-GitHub:
## Motivation and Context

Issue #981 mentions that the program crashes when executing the following code:

```python
for i in range(10):
    def foo():
        break
```

This pull request fixes that crash.

Closes #981.

## Your Changes

**Description**: A KeyError occurred when executing the code snippet, so I added an error-checking condition to check whether the key exists in the dictionary.

- [X] Bug fix (non-breaking change which fixes an issue)

## Testing

I tested the code by running the same code snippet as above and verifying that indeed no error messages are appearing.

## Checklist


python_ta/cfg/visitor.py
```diff
@@ -312,11 +323,13 @@ def _visit_jump(
...
                 break
+
         else:
-            raise SyntaxError(
+            logging.warning(
                 f"'{type(node).__name__}' outside"
                 f' {"function" if isinstance(node, nodes.Return) else "loop"}'
             )
+
         unreachable_block = self._current_cfg.create_block()
```
david-yz-liu, line 333: @MariaMa-GitHub okay I looked into this a bit more. You've made good progress, and the issue with the "black" box not being correct linked is due to the following reason. With the "valid" jumps inside the loop, the logic has two steps:

1. link `old_curr` with the correct "exit block"
2. add `node` to `old_curr` (so that it appears inside the block). This needs to happen _after_ the linking because the current implementation of `link` restricts linking when the last statement inside the block is a jump node. (I suspect this is actually not the best design, but it's what we have right now.)

So to mimic this behaviour for the "invalid" jump cases, you need to (1) link the current block to the new `unreachable` block created at the bottom of this file, and then (2) call `add_statement` directly, mimicking the "valid case" logic. `visit_generic` should also work, but I think it's best to be consistent with the existing code.

david-yz-liu (commented):
@MariaMa-GitHub great work! I've left a few comments. Please also update the changelog; you can add two separate entries, one that documents fixing the bug described in the original issue, and one "enhancement" that describes changing the behaviour of the cfg generation when a jump statement is syntactically invalid.

PR #1024, opened 2024-03-10
allowed annotated constants to be permitted as top level code

### LINKED ISSUE #1012: Adding type annotation to a constant causes a E9992 (forbidden-top-level-code)
The following snippet causes a E9992 (forbidden-top-level-code)
```python
COUNT: int = 10

if __name__ == "__main__":
    python_ta.check_all(config={
        'output-format': 'python_ta.reporters.PlainReporter'
    })
```
The expected behaviour is that the constant type annotation show not be flagged as forbidden top level code.

michaelZhaoCode:
## Motivation and Context

This pull request aims to resolve #1012 

Although uncommon, Python allows for type annotation to variable assignments. However, performing annotated constant variable assignment is in a different node type than regular constant assignment and causes the checker `top_level_code_checker` to incorrectly parse the statement. This causes it to consider the statement as forbidden top level code since it was uncounted for and incorrectly adds an error message. This pull request fixes the issue by parsing both annotated and unannotated constant variable assignment statements. 

Additionally, changing the attributes of classes in top level code was not picked up by the checker. This corrects the behaviour.
## Your Changes

Changed the `_is_allowed_assignment` function in `top_level_code_checker` to parse nodes of type `AnnAssign` in addition to nodes of type `Assign` to consider if they are constant assignments.

Added a check to see if names was non-empty when checking assignment statements to prevent Attribute assignments to be passed.

Fix was noted in the changelog.

**Description**:

## Testing

Tested manually that annotated constant assignment causes no message to be raised.
Added test case in test suite for this case and verified test passed in addition to preexisting tests for this checker.
Additional test was added to check that attribute assignment is caught correctly

Is this separation of cases acceptable or is there a more efficient method to implement the change?

## Checklist

### DIFF
--- a/python_ta/checkers/top_level_code_checker.py
+++ b/python_ta/checkers/top_level_code_checker.py
@@ -65,7 +65,7 @@ def _is_allowed_assignment(statement: nodes.NodeNG) -> bool:
             node.name for node in statement.target.nodes_of_class(nodes.AssignName, nodes.Name)
         )
 
-    return all(
+    return names and all(
         re.match(UpperCaseStyle.CONST_NAME_RGX, name)
         or re.match(DEFAULT_PATTERNS["typealias"], name)
         for name in names
--- a/python_ta/checkers/top_level_code_checker.py
+++ b/python_ta/checkers/top_level_code_checker.py
@@ -53,12 +53,15 @@ def _is_allowed_assignment(statement: nodes.NodeNG) -> bool:
     """
     Return whether or not <statement> is a constant assignment or type alias assignment.
     """
-    if not isinstance(statement, nodes.Assign):
+    if not isinstance(statement, nodes.Assign) and not isinstance(statement, nodes.AnnAssign):
         return False
 
     names = []
-    for target in statement.targets:
-        names.extend(node.name for node in target.nodes_of_class(nodes.AssignName, nodes.Name))
+    if isinstance(statement, nodes.Assign):
+        for target in statement.targets:
+            names.extend(node.name for node in target.nodes_of_class(nodes.AssignName, nodes.Name))
+    else:
+        names.append(statement.target.name)
 
     return all(
         re.match(UpperCaseStyle.CONST_NAME_RGX, name)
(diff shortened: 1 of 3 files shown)


python_ta/checkers/top_level_code_checker.py
```diff
@@ -53,12 +53,15 @@ def _is_allowed_assignment(statement: nodes.NodeNG) -> bool:
...
         return False
 
     names = []
-    for target in statement.targets:
-        names.extend(node.name for node in target.nodes_of_class(nodes.AssignName, nodes.Name))
+    if isinstance(statement, nodes.Assign):
+        for target in statement.targets:
+            names.extend(node.name for node in target.nodes_of_class(nodes.AssignName, nodes.Name))
+    else:
+        names.append(statement.target.name)
```
david-yz-liu, line 64: While you have correctly identified `statement.target` as being singular (as opposed to the `.targets` above), you should be careful not to assume a particular class. Check out the astroid documentation for `nodes.AnnAssign`: it shows different possible types for the `target` attribute.

So, I still recommend using `nodes_of_class` here.

tests/test_custom_checkers/test_top_level_code_checker.py
```diff
@@ -86,6 +86,15 @@ def test_no_message_constant_assignment(self):
         with self.assertNoMessages():
             self.checker.visit_module(mod)
 
+    def test_no_message_annotated_constant_assignment(self):
```
david-yz-liu, line 89: This is a good test to add. Since this is a "no message" case, I suggest adding a second test which is a "reports message" case when the variable isn't in `ALL_CAPS` naming style.

david-yz-liu (commented):
@michaelZhaoCode great work. In addition to the inline comments, you can use GitHub keywords in the PR description to indicate that this PR fixes the issue. Then when the PR is merged in, the issue will automatically be closed. See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

david-yz-liu (approved):

PR #1025, opened 2024-03-14
Raise AssertionError when precondition contains precondition violations

### LINKED ISSUE #1003: check_contracts fails silently when preconditions contain precondition violations
The following code runs without error, but feels like it should reasonably be expected to raise some kind of `AssertionError` instead. The `AssertionError` raised when evaluating the precondition for `my_condition` is being flagged as indicating that the precondition for `my_function` "could not be evaluated". It maybe isn't entirely clear what the behaviour _should_ be in this case, but the current behaviour feels likely undesirable in terms of communicating that there _really was_ a precondition violation detected during execution of the code — it is just that it was _while_ checking the original precondition.

```python
from python_ta.contracts import check_contracts

from python_ta import contracts

contracts.DEBUG_CONTRACTS = True

@check_contracts
def my_condition(x: int):
    """
    Preconditions:
    - x > 0
    """
    return x > 0


@check_contracts
def my_function(y: int):
    """

    Preconditions:
    - my_condition(y)
    """
    return y


if __name__ == '__main__':
    print(my_function(-1))
```

The output with debugging contracts turned on is:

```
[DEBUG] Checking type of parameter y in call to my_function
[DEBUG] Adding assertion to my_function: my_condition(y)
[DEBUG] Checking precondition for my_function: my_condition(y)
[DEBUG] Checking type of parameter x in call to my_condition
[DEBUG] Adding assertion to my_condition: x > 0
[DEBUG] Checking precondition for my_condition: x > 0
[DEBUG] Warning: could not evaluate precondition: my_condition(y)
-1
```

If one instead runs `print(my_condition(-1))`, the output is:

```
[DEBUG] Checking type of parameter x in call to my_condition
[DEBUG] Adding assertion to my_condition: x > 0
[DEBUG] Checking precondition for my_condition: x > 0
  File "....", line 28, in <module>
    print(my_condition(-1))
          ^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/python_ta/contracts/__init__.py", line 96, in _enable_function_contracts
AssertionError: my_condition precondition "x > 0" was violated for arguments {x: -1}
```

Something more like this is what one would likely expect when executing the original `print(my_function(-1))`

elescogna:
## Motivation and Context
When the preconditions for a function contain a call to another function whose preconditions are violated, `check_contracts` fails silently, without raising an `AssertionError`, as it happens when the violated precondition is checked direclty instead. This pull request addresses this behavior, so that nested precondition violations are correctly signaled through an `AssertionError`. In addition, this pull request also solves a related issues with Representation Invariants, allowing for the use or user defined functions (possibly with preconditions) within representation invariants.

 closes #1003 

## Your Changes

I modified part of the exception handling in `python_ta.contracts.__init__.py`. The functions `_check_assertions` and `_check_invariants` now raise an `AssertionError` directly when an `AssertionError` is found in function preconditions or class invariants, instead of simply adding a message to the debugging log.
This allows for preconditions and representation invariants violations to be signaled as soon as they occur, and to display the error message relative to the inner precondition violation.

**Description**:

## Testing

I have created the test file `nested_preconditions_example.py` with example code that failed silently before the change. Then, I added unit test to `test_contracts.py` and `test_class_contracts.py` which use the code in `nested_preconditions_example.py` to test that an `AssertionError` is correctly raised in the appropriate scenarios.

## Checklist

- [x]  I have reviewed the test coverage changes reported on Coveralls. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Fixed bug where `_` was marked as a built-in when running PythonTA after running doctest
 - Fixed issue where annotated constant variable assignment was not considered as permissible top level code and triggered error E9992
 - Fixed issue where top level class attribute assignment was considered as permissible top level code
-- Fixed issue where `check_contracts` fails silently when function preconditions contain precondition violations
+- Fixed issue where `check_contracts` fails silently when function preconditions contain precondition violations, and when a representation invariant contains a call to a top-level function (not built-in or imported library).
 
 ## [2.7.0] - 2024-12-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Fixed bug where `_` was marked as a built-in when running PythonTA after running doctest
 - Fixed issue where annotated constant variable assignment was not considered as permissible top level code and triggered error E9992
 - Fixed issue where top level class attribute assignment was considered as permissible top level code
+- Fixed issue where `check_contracts` fails silently when function preconditions contain precondition violations
 
 ## [2.7.0] - 2024-12-14
 
--- a/tests/test_contracts/test_class_contracts.py
+++ b/tests/test_contracts/test_class_contracts.py
@@ -424,5 +424,40 @@ def test_no_premature_check_from_deep_helper_in_init() -> None:
     assert dark_widget.secondary_color == "mahogany"
 
 
+def test_invariant_with_function_defined_in_module() -> None:
+    """Test that a representation invariant violation is detected when the invariant
+    contains a call to a function (top-level) defined by the user
+    This test is based on the code found at ./test_nested_preconditions_example.py
+    """
+    import test_nested_preconditions_example as example
+
+    with pytest.raises(AssertionError) as exception_info:
+        example.Student("Bob", 0000000000, 3.5, 18)
+
+    assert (
+        str(exception_info.value) == '"Student" representation invariant '
+        '"validate_student_number(self.student_number)" '
+        "was violated for instance attributes "
+        "{name: " + "'Bob'" + ", student_number: 0, gpa: 3.5, age: 18}"
+    )
+
+
+def test_invariant_with_class_method() -> None:
+    """Test that a representation invariant violation is detected when the invariant
+    contains a call to a function (class method) defined by the user.
+    This test is based on the code found at ./test_nested_preconditions_example.py
+    """
+    import test_nested_preconditions_example as example
+
+    with pytest.raises(AssertionError) as exception_info:
+        example.Student("Bob", 1001001000, 3.5, -1)
+
+    assert (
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Fixed bug where `_` was marked as a built-in when running PythonTA after running doctest
 - Fixed issue where annotated constant variable assignment was not considered as permissible top level code and triggered error E9992
 - Fixed issue where top level class attribute assignment was considered as permissible top level code
+- Fixed issue where `check_contracts` fails silently when function preconditions contain precondition violations
```
david-yz-liu, line 26: update this to also talk about representation invariants

tests/test_contracts/test_class_contracts.py
```diff
@@ -424,5 +424,40 @@ def test_no_premature_check_from_deep_helper_in_init() -> None:
     assert dark_widget.secondary_color == "mahogany"
 
 
+def test_invariant_with_function_defined_in_module() -> None:
+    """Test that a representation invariant violation is detected when the invariant
+    contains a call to a function (top-level) defined by the user
+    This test is based on the code found at ./test_nested_preconditions_example.py
+    """
+    import test_nested_preconditions_example as example
```
david-yz-liu, line 432: Move this import statement to the top level of this file. Also, just to be safe don't name that module starting with `test_`, we don't want pytest to scan the file.

tests/test_contracts/test_contracts.py
```diff
@@ -702,3 +702,50 @@ class Person:
...
     assert my_person.age == "John"
+
+
+def test_nested_preconditions_contract_checking() -> None:
+    """
+    Test that an AssertionError is correctly raised when a precondition violation occurs while
+    checking the precondition for another function.
+    This test is based on the code found at ./test_nested_preconditions_example.py
+    """
+    from test_nested_preconditions_example import my_function
```
david-yz-liu, line 713: see above comment

tests/test_contracts/test_nested_preconditions_example.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@check_contracts
+class Student:
+    """
+    A representation of a student
+
+    Representation Invariants:
+     - validate_student_number(self.student_number)
+     - self.condition2(self.gpa)
```
david-yz-liu, line 43: Try running this type of example with `DEBUG_CONTRACTS` turned on. You'll see that this essentially triggers an infinite recursion situation (somewhat indirectly, through `eval`) because each time a method is called, the representation invariants are checked again, or something like that.

Anyway, this is a current limitation of PythonTA so I don't think you need to worry about fixing it, for now you should just avoid calling methods as representation invariants, and instead use the other top-level functions in this file.

elescogna: Ah, this makes a lot of sense! I have just noticed that I was having issues creating valid instances of the class, as no AssertionErrors were raised, and that representation invariants kept being checked cyclically, though I was not sure why.

david-yz-liu (approved):

PR #1031, opened 2024-04-01
Fixed infinite recursion in representation invariants with method calls

elescogna:
## Motivation and Context

When a representation invariant of a class invokes one of its methods, the representation invariants for that class keep being checked recursively, leading to infinite recursion. This pull request fixes this issue, allowing to use class methods in representation invariants for more flexibility.

## Your Changes

I temporarily added an attribute `__pyta_currently_checking` to the instance whose representation invariants are being checked. Such attribute is added at the beginning of the check and deleted at the end, and the representation invariants are only checked when this attribute exists. Finally, I modified `new_setattr` to skip the representation invariant check exclusively for the `__pyta_currently_checking` attribute.

**Description**:

## Testing

I added four tests to the `test_class_contract.py` file testing for the correct behavior with valid and invalid attribute values both when initializing a class instance and when setting an attribute to a new value.

**Comments**:
I decided to delete the `__pyta_currently_checking` attribute instead of simply setting it to `False` because it should not be accessible to the user after the RI check has concluded.

**Questions**:
- For now, these changes don't seem to allow to call methods of a different class in the representation invariants. When this happens, the check fails silently, and the class instance is created successfully.
- When a class method is called with an argument that is not defined, the precondition check also fails silently.

What should be the expected behavior in both these cases?

## Checklist

### DIFF
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -164,11 +164,6 @@ def add_class_invariants(klass: type) -> None:
 
         Check representation invariants for this class when not within an instance method of the class.
         """
-        if name == "currently_checking":
-            # This is a special attribute, so representation invariants should not be checked
-            super(klass, self).__setattr__(name, value)
-            return
-
         if not ENABLE_CONTRACT_CHECKING:
             super(klass, self).__setattr__(name, value)
             return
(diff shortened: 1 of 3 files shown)


python_ta/contracts/__init__.py
```diff
@@ -399,6 +404,12 @@ def _check_class_type_annotations(klass: type, instance: Any) -> None:
 
 def _check_invariants(instance, klass: type, global_scope: dict) -> None:
     """Check that the representation invariants for the instance are satisfied."""
+    if hasattr(instance, "currently_checking"):
+        # If already checking invariants for this instance, skip to avoid infinite recursion
+        return
+
+    instance.currently_checking = True
```
david-yz-liu, line 411: Rather than using dot notation, use the `super(...).__setattr__(...)` call here. This skips the call to `new_setattr` above, so you won't need to change the implementation of that method.

python_ta/contracts/__init__.py
```diff
@@ -414,6 +425,10 @@ def _check_invariants(instance, klass: type, global_scope: dict) -> None:
             _debug(f"Warning: could not evaluate representation invariant: {invariant}")
         else:
             if not check:
+                # Remove the "currenlty_checking" attribute before printing the error message to the user
```
david-yz-liu, line 428: It's a bit cumbersome to duplicate the "delattr" code. You are trying to achieve the behaviour of "regardless of what happens inside this loop, make sure `delattr` is called".

The cleanest way to do this is by using an outer `try` block that has a `finally` clause (you can look up what `finally` does if you haven't used it before).

david-yz-liu (commented):
@elescogna good work. In addition to the inline comments, a few overall comments:

- Name the attribute `__pyta_currently_checking` (the double underscore marks it as more private, and the name `pyta` makes it clear where this attribute is from
- I'm not sure what you mean by "RIs from another class aren't being called properly" (paraphrasing), but you can push an example of that to show a failing test case
- If there's any other error in the RIs (e.g. an undefined variable used as an argument), it's the expected behaviour to print a debug message warning about this, but not to raise an error.

david-yz-liu (approved):

PR #1033, opened 2024-04-04
Added function preconditions to initial function CFG edge

michaelZhaoCode:
## Motivation and Context

The CFG visitor creates a new CFG for any function blocks encountered. Assuming we are checking preconditions, these blocks would only be ran if the preconditions are satisfied. Thus, the initial edge in these function CFGs now stores the preconditions of the function, as an `and` node in the condition attribute of the edge. 

## Your Changes

Created helper function to extract preconditions, combine them into an `and` expression, and parse into AST node
Used helper function to set CFGEdge condition in the initial function CFG edge to store the condition node
Allowed `parse_arguments` from contracts to take in AST function definition nodes with no docstring
Created tests to verify conditions were set properly

**Description**:

## Testing

Implemented tests to ensure condition was set properly.

Should this be added to the changelog file?

## Checklist

### DIFF
--- a/python_ta/cfg/visitor.py
+++ b/python_ta/cfg/visitor.py
@@ -1,6 +1,8 @@
 from typing import Any, Dict, List, Optional, Tuple, Union
 
-from astroid import nodes
+from astroid import nodes, parse
+
+from python_ta.contracts import parse_assertions
 
 from .graph import CFGBlock, ControlFlowGraph
 
@@ -124,7 +126,11 @@ class CFGVisitor:
         self._current_cfg.start.add_statement(func.args)
         func.cfg_block = self._current_cfg.start
 
-        self._current_block = self._current_cfg.create_block(self._current_cfg.start)
+        preconditions_node = _get_preconditions_node(func)
+
+        self._current_block = self._current_cfg.create_block(
+            self._current_cfg.start, edge_condition=preconditions_node
+        )
 
         for child in func.body:
             child.accept(self)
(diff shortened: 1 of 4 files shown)


python_ta/cfg/visitor.py
```diff
@@ -435,3 +441,16 @@ def _get_raise_exc(node: nodes.Raise) -> str:
...
+
+def _get_preconditions_node(func: nodes.FunctionDef) -> Optional[nodes.BoolOp]:
+    """A helper method that takes in a function definition node, retrieves its preconditions, and then parses them
+    into a AST node representing all the preconditions combined in an and statement. Returns None if there are no
+    preconditions."""
+    assertions = parse_assertions(func)
+    if not assertions:
+        return None
+    precondition_string = " and ".join(assertions)
+    condition, *_ = parse(precondition_string).nodes_of_class(nodes.Expr)
```
david-yz-liu, line 454: Two comments here:

1. You should be able to use `extract_node` to obtain the minimal expression node, rather than using `nodes_of_class`
2. More importantly, this will fail if users specify non-Python preconditions (e.g. if there's a typo, or an English precondition). You should add test cases for this.

    In terms of expected behaviour, the easier approach is to return `None` in this case. The more complex approach is to filter out the non-Python preconditions and only join together the Python preconditions. I'll leave it to you to decide which approach to take, based on how much time you'd like to commit to this issue. 😃

python_ta/cfg/visitor.py
```diff
@@ -435,3 +441,16 @@ def _get_raise_exc(node: nodes.Raise) -> str:
...
+    x = 5
```
david-yz-liu, line 455: this line should be removed

python_ta/cfg/visitor.py
```diff
@@ -435,3 +442,28 @@ def _get_raise_exc(node: nodes.Raise) -> str:
...
+    """A helper method that takes in a function definition node, retrieves its preconditions, and then parses them
+    into a AST node representing all the preconditions combined in an and statement. Returns None if there are no
+    preconditions."""
+    assertions = parse_assertions(func)
+    valid_assertions = [assertion for assertion in assertions if _is_python_precondition(assertion)]
+    # for now, return None if at least one non-valid assertion
+    # remove the length comparison and improve valid precondition checker to change this behaviour
+    if not valid_assertions or len(valid_assertions) != len(assertions):
+        return None
+    precondition_string = " and ".join(valid_assertions)
```
david-yz-liu, line 457: One other thing, you should really wrap each assertion in parentheses, to avoid any issues regarding `and` and [operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence). It's the same issue as C macros that you learned about in CSC209.

david-yz-liu (commented):
@michaelZhaoCode good work. I left one comment; also, I think this change warrants an entry under the Changelog "enhancements".

david-yz-liu (approved):

PR #1039, opened 2024-05-21
Separated tests for __main__.py as requested

quiz3:
## Proposed Changes

_(Describe your changes here. Also describe the motivation for your changes: what problem do they solve, or how do they improve the application or codebase? If this pull request fixes an open issue, [use a keyword to link this pull request to the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).)_

...

<details>

</details>

## Type of Change

## Checklist

- [ ] I have performed a self-review of my changes.
- [ ] I have added tests for my changes, if applicable.
- [ ] I have updated the project documentation, if applicable.
- [ ] I have updated the project Changelog (this is required for all changes).

## Questions and Comments

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,0 +1,77 @@
+"""Run from the `pyta` root directory to use the local `python_ta` rather than
+installed `python_ta` package.
+"""
+
+import subprocess
+import sys
+from os import environ
+from unittest.mock import Mock
+
+import python_ta
+
+
+def test_check_no_errors_zero() -> None:
+    """Test that python_ta exits with status code 0 when it does not detect errors."""
+    output = subprocess.run(
+        [
+            sys.executable,
+            "-m",
+            "python_ta",
+            "--config",
+            "tests/test.pylintrc",
+            "tests/fixtures/no_errors.py",
+        ]
+    )
+
+    assert output.returncode == 0
+
+
+def test_check_errors_nonzero() -> None:
+    """Test that python_ta exits with non-zero status code when it detects errors."""
+    output = subprocess.run(
+        [
+            sys.executable,
+            "-m",
+            "python_ta",
+            "--config",
+            "tests/test.pylintrc",
+            "examples/nodes/name.py",
+        ]
+    )
+
+    assert output.returncode != 0
+
+
+def test_check_exit_zero() -> None:
+    """Test that python_ta --exit-zero always exits with status code 0,
+    even when given a file with errors.
+    """
+    output = subprocess.run(
+        [
+            sys.executable,
+            "-m",
+            "python_ta",
+            "--exit-zero",
+            "--config",
+            "tests/test.pylintrc",
+            "examples/nodes/name.py",
+        ],
+        env={**environ, "PYTHONIOENCODING": "utf-8"},
+    )
+
+    assert output.returncode == 0
+
+
+def test_check_version() -> None:
+    """Test that python_ta --version outputs python_ta.__version__ to stdout."""
+    stdout = subprocess.run(
+        [
+            sys.executable,
+            "-m",
+            "python_ta",
+            "--version",
+        ],
+        capture_output=True,
+        text=True,
+    ).stdout
+    assert stdout.rstrip("\n") == python_ta.__version__
--- a/tests/test_check.py
+++ b/tests/test_check.py
@@ -2,9 +2,7 @@
 installed `python_ta` package.
 """
 
-import subprocess
-import sys
-from os import environ, path, remove
+from os import path, remove
 from unittest.mock import Mock
 
 import python_ta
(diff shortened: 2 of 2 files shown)


david-yz-liu:
@quiz3 great questions. You can run `pytest` with test coverage enabled locally with the command `pytest --cov python_ta --cov-config=.coveragerc tests`. For more information, check out https://pytest-cov.readthedocs.io/en/latest/readme.html.

quiz3:
I made changes to the `README.md` because I found that some details were out-of-date. Let me know if these changes should be put into a separate PR.

quiz3:
I think the CI tests are a little janky. 
I say this because I had multiple failures related to the accumulation table, which I did not touch at all in this PR. 
Then, I deleted a comment I had previously added to force the tests to run again. 
They passed. 
Something is clearly amiss in this situation. 

I have created an issue attempting to describe the accumulation table error, but I believe part of the issue may actually relate to the _tests_ of the accumulation table rather than the source code itself.

tests/test_main.py
```diff
@@ -0,0 +1,124 @@
...
+            "python_ta",
+            "--generate-config",
+            "--config",
+            "tests/test.pylintrc",
+        ],
+        capture_output=True,
+        text=True,
+    ).stdout.rstrip("\n")
+
+    actual_config = subprocess.run(
```
david-yz-liu, line 97: You don't need subprocess here. Instead, you can just read the contents of that file (i.e., use the file `read` method). To avoid the problem of relative paths, talk to @Raine-Yang-UofT about what I suggested to him and use the same approach here.

quiz3: I was slightly concerned about the relative paths but I am actually somewhat more concerned about the fact that the paths are hard-coded. 

Is there a way to set up a global constant defining, for instance, the path to the default config file?

david-yz-liu: Though to be clear, such a constant should only include the relative path from the source code root to the config file. So you'll still need to avoid the problem of relative paths by following what I mentioned above.

quiz3: I wrote @Raine-Yang-UofT but he has not yet gotten back to me. Did you put that information in a public comment on his PR as well?

quiz3: Did you mean I should introduce a constant and store it under `python_ta/config` or that the constant should store the path `python_ta/config`. 

I initially interpreted your comment as the former, but I have found that this interpretation doesn't seem practical given what is currently stored in that location. Also, the `.pylintrc` config file is right in that exact directory, so storing such a constant there, right next to the file it points to, seems a bit redundant.

quiz3: > Though to be clear, such a constant should only include the relative path from the source code root to the config file. So you'll still need to avoid the problem of relative paths by following what I mentioned above.

I think this is the only thing I am now missing. I am not sure what the approach you would recommend is for avoiding the relative path issue. What I would imagine is it involves some way of finding the source code root from anywhere in the project directory structure, such that a relative path from the source root to anywhere is still useful, since we have now found the path to the source root. I hope that makes sense...

quiz3: > > Though to be clear, such a constant should only include the relative path from the source code root to the config file. So you'll still need to avoid the problem of relative paths by following what I mentioned above.
> 
> I think this is the only thing I am now missing. I am not sure what the approach you would recommend is for avoiding the relative path issue. What I would imagine is it involves some way of finding the source code root from anywhere in the project directory structure, such that a relative path from the source root to anywhere is still useful, since we have now found the path to the source root. I hope that makes sense...

Ok, he got back to me. I believe I understand your reasoning for asking for that but I think in some cases a more generalized approach may be more beneficial. I will submit what I mean shortly in a new commit...

python_ta/__main__.py
```diff
@@ -1,11 +1,20 @@
...
 
 import click
 
 from python_ta import __version__, check_all, check_errors
 
+# Gives absolute path to source root directory (pyta)
+SOURCE_ROOT = path.normpath(path.join(path.dirname(__file__), ".."))
+
+# Defines relative path to the default config directory (python_ta/config)
+CONFIG_LOCATION = path.join("python_ta", "config")
```
david-yz-liu, line 13: Please just add one variable `DEFAULT_CONFIG_LOCATION` that stores the value `config/.pylintrc`. I.e., it should be relative to this file. I asked you to add this constant to the `python_ta/config` subfolder: I don't think you understood what I meant, but the best place to add it is the `__init__.py` file. That way you can import it from `python_ta.config`.

Remove the other variables, they aren't necessary.

python_ta/__main__.py
```diff
@@ -55,8 +64,7 @@ def main(
 
     # `config` is None if `-c` flag is not set
     if generate_config:
-        pylintrc_location = os.path.join(os.path.dirname(__file__), "config/.pylintrc")
```
david-yz-liu, line 58: Keep this line but use the variable I mentioned above.

tests/test_main.py
```diff
@@ -0,0 +1,141 @@
...
+import python_ta
+
+# Find the source root directory
+source = path.dirname(__file__)
+while path.basename(source) != "pyta":
+    source = path.dirname(source)
+
+# Gives absolute path to source root directory (pyta)
+SOURCE_ROOT = source
+# SOURCE_ROOT = path.normpath(path.join(path.dirname(__file__), ".."))  # NOTE: Professor may prefer this approach
```
david-yz-liu, line 18: This is indeed the approach I want you to use. You can also import the correct relative path from `python_ta.config` (see my above comment).

In general, you should not have top-level code other than defining constants (so while loops aren't good). You also shouldn't assume the root-level folder is called `pyta`, it could be called something else.

python_ta/__main__.py
quiz3: Would you want me to change lines like these in the `python_ta/config/__init__.py` file?

david-yz-liu: No, don't change these lines. You're correct that they're looking in a `config` directory, but the file names are not the same in the two branches. Not in scope for this PR.

tests/test_main.py
```diff
@@ -0,0 +1,125 @@
+"""Run from the `pyta` root directory to use the local `python_ta` rather than
+installed `python_ta` package.
+"""
+
+import subprocess
+from os import environ, path
+from sys import executable as sys_executable
```
david-yz-liu, line 7: @quiz3 Please change this back. Don't make unnecessary changes to get tests to re-run. You can message me on Slack and I'll re-run them.

david-yz-liu (approved):

david-yz-liu:
Great job, @quiz3 🎉

PR #1055, opened 2024-06-16
feat: Implement checkers for inconsistent and missing return statements

Raine-Yang-UofT:
## Proposed Changes

Replaced the pylint checker R1710 inconsistent-return-statements with a custom checker inconsistent-or-missing-returns. The new checker uses a CFG-based algorithm and provides custom error rendering for missing return statements. Added custom error messages R9710 inconsistent-returns and R9711 missing-returns for better error reporting.

...

<details>

</details>

## Type of Change

## Checklist

## Questions and Comments


python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        if (
+            node.returns is None
+            or isinstance(node.returns, nodes.Const)
+            and node.returns.value is None
+        ):
+            return
+
+        # get the end of CFG
+        cfg = ControlFlowGraph()
```
david-yz-liu, line 56: You shouldn't need to create a new CFG; there is a PythoNTA node visitor that creates a CFG before any of the checkers run. Try looking at some of the existing custom checkers that are CFG-based to see how they access nodes.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,82 @@
...
+        cfg.start = node.cfg_block
+        # The last element of cfg.get_blocks_postorder() does not guarantee to be the end block.
+        # However, based on the initialization of CFG, end block must have id == 1
+        end = [block for block in cfg.get_blocks_postorder() if block.id == 1][0]
```
david-yz-liu, line 60: The `ControlFlowGraph` class should have an `end` attribute that specifies the last block. Use that, and if there are any logical errors (in the graph structure itself) please investigate those further.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,82 @@
...
+        """Visit a function definition"""
+        self._check_return_statements(node)
+
+    def _check_return_statements(self, node) -> None:
+        """
+        Construct a CFG from the function. Check for inconsistent returns if there are
+        multiple return statements, and missing return statements if there are none.
+        """
+        if (
+            node.returns is None
```
david-yz-liu, line 49: This is a bit subtle. I believe `node.returns is None` is true when there's _no_ type annotation at all? In this case I wonder what pylint's R1710 does.

I suspect that in this case, the check could still execute, but then there's a special case when there's no return statement, or only bare `return`s, in which case no error is reported. That is, in this case the check is conditioned on whether at least one return statement has an expression, in which case all of them should.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,82 @@
...
+        cfg.start = node.cfg_block
+        # The last element of cfg.get_blocks_postorder() does not guarantee to be the end block.
+        # However, based on the initialization of CFG, end block must have id == 1
+        end = [block for block in cfg.get_blocks_postorder() if block.id == 1][0]
+        end_blocks = [edge.source for edge in end.predecessors]
+
+        # gather all return statements
+        for block in end_blocks:
+            has_return = False  # whether a return statement exists for this branch
+            for statement in block.statements:
```
david-yz-liu, line 66: I wonder if a loop here is necessary. I can't recall the CFG implementation, but in principle the return statement should always be the last statement in a block.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,82 @@
...
+            Note: one common source of this error is if you're using if statements without an explicit else branch.
+            In this case, you should consider revising your code to either add an else branch,
+            or, if you are confident that the if and elif conditions cover all possible cases,
+            "you can convert the final "elif " into an " else ".",""",
+            "inconsistent-returns",
+            "Used to replace R1710 message inconsistent-return-statements",
+        ),
+        "R9711": (
+            "Missing return statement in function",
+            "missing-return-statements",
```
david-yz-liu, line 31: make the name singular (i.e., `missing-return-statement`)

python_ta/reporters/node_printers.py
```diff
@@ -614,6 +614,36 @@ def render_pep8_errors_e306(msg, _node, source_lines=None):
...
 
 
+def render_missing_return_statements(msg, _node, source_lines=None):
+    """
+    Render a missing return statements message
+    """
+    line = msg.line
+    end = _node.tolineno
+
+    # render function header and context
```
david-yz-liu, line 624: While this was not a bad suggestion from the group, it's quite different in style than other PythonTA error rendering. Instead of doing this, I recommend passing the *function name* in the `args` to the message, and then displaying it in the text of the message. (As a comment, this might actually be good to do for other checks as well, but that's another matter.)

python_ta/reporters/node_printers.py
```diff
@@ -614,6 +614,36 @@ def render_pep8_errors_e306(msg, _node, source_lines=None):
...
+
+    # insert the message
+    yield (
+        None,
+        slice(indentation, None),
+        LineType.ERROR,
+        insertion_text,
+    )
+
+    yield from render_context(end + 1, end + 2, source_lines)
```
david-yz-liu, line 644: Similar to my above comment, our convention is to render *2* lines before and after the "error line" as the context

python_ta/reporters/node_printers.py
```diff
@@ -614,6 +614,36 @@ def render_pep8_errors_e306(msg, _node, source_lines=None):
...
+    if line - 1 > _node.fromlineno:
+        yield from render_context(_node.fromlineno, _node.fromlineno + 1, source_lines)
+        if line - 1 > _node.fromlineno + 1:
+            yield (None, slice(None, None), LineType.CONTEXT, '"""MORE CODE OMITTED"""')
+    yield from render_context(line - 1, line + 1, source_lines)
+
+    # calculate indentation for the insertion point
+    body = source_lines[end - 1]
+    indentation = len(body) - len(body.lstrip())
+    insertion_text = body[:indentation] + '"""MISSING RETURN STATEMENT"""'
```
david-yz-liu, line 634: Change this to a comment rather than a docstring, and use the more direct instruction `INSERT RETURN STATEMENT HERE`

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,82 @@
...
+            has_return = False  # whether a return statement exists for this branch
+            for statement in block.statements:
+                if isinstance(statement, nodes.Return):
+                    has_return = True
+                    if statement.value is None:
+                        # check for inconsistent returns
+                        self.add_message("inconsistent-returns", node=statement)
+
+            # check for missing return statement
+            if not has_return:
```
david-yz-liu, line 74: This is more of a comment for investigation, but I wonder if it is only possible for `has_return` to be `False` for at most one block in `end_blocks`? Or perhaps not (e.g. when an if statement with multiple branches is the last statement of a function body).

david-yz-liu (commented):
@Raine-Yang-UofT great work! I left one round of review comments which are mostly high-level. After you work through those I'll take a more detailed look.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
Raine-Yang-UofT: Hi Professor, I checked that a function definition node only contains a reference to the start of the function (cfg_block attribute), so I do need to invoke get_blocks_postoreder() method in ControlFlowGraph to access other blocks of the CFG. Some other CFG-based checkers, such as redundant_assignment_checker, use the same approach.

However, if we make a slight change in CFGVisitor and append each module definition and function definition with a reference to its corresponding CFG graph, we can directly access the whole CFG inside the checker. I was wondering if you'd like to consider this change, and then we can probably also refactor the existing checkers as well.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
Raine-Yang-UofT: Yes, `node.returns is None` does mean that there is no return type annotation. Previously I thought if having no annotation is equivalent to having return type as `None`. I guess what you mean here is that when there is no type annotation, we only check for inconsistent returns but not missing returns?

david-yz-liu: @Raine-Yang-UofT right, "no annotation" should be interpreted as "unspecified", not "return type `None`".

Your interpretation is correct: when no type annotation is provided, check for inconsistent returns but not missing returns. However, note that if *all* return statements are just `return`, this should not be considered "inconsistent returns".

python_ta/checkers/inconsistent_or_missing_returns_checker.py
david-yz-liu: Great research, @Raine-Yang-UofT. Please modify the `CFGVisitor` class as you suggested so that the `ControlFlowGraph` can be accessed. However, only use the new attribute in this checker; we'll leave the work of refactoring the other checkers as a separate PR.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,82 @@
...
+    def _check_return_statements(self, node) -> None:
+        """
+        Construct a CFG from the function. Check for inconsistent returns if there are
+        multiple return statements, and missing return statements if there are none.
+        """
+        if isinstance(node.returns, nodes.Const) and node.returns.value is None:
+            return
+
+        has_return_annotation = node.returns is not None
+        any_branches_have_return = False
```
david-yz-liu, line 52: "any branches" isn't grammatically correct. Rename to just `has_return`

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,82 @@
...
+        end_blocks = [edge.source for edge in end.predecessors]
+
+        # gather the return statement of each code block
+        return_statements = {}
+        for block in end_blocks:
+            return_statements[block] = None
+            statement = block.statements[-1]
+            if isinstance(statement, nodes.Return):
+                return_statements[block] = statement
+                if return_statements[block].value is not None:
```
david-yz-liu, line 65: After reviewing this, I realized that I also would like you to check for `return None` (count this as a "non-expression return statement" for the purposes of setting this flag.

That is, if all return statements are of the form `return None`, I still want the `missing-return-statement` error to be triggered. You should update the tests, documentation, and error message accordingly.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,82 @@
...
+                    any_branches_have_return = True
+
+        # check for inconsistent or missing returns
+        if has_return_annotation or any_branches_have_return:
+            for block in return_statements:
+                statement = return_statements[block]
+                if statement is None:
+                    # for rendering purpose, the line is set to the last line of the function branch where return statement is missing
+                    self.add_message(
+                        "missing-return-statement", node=node, line=block.statements[-1].fromlineno
```
david-yz-liu, line 75: You probably should use `endlineno` instead (test this manually with multi-line statements).

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,82 @@
...
+                    )
+                elif statement.value is None:
```
david-yz-liu, line 77: To be clear, this check is good and I _don't_ want you to also check for `return None` in this condition.

python_ta/reporters/node_printers.py
```diff
@@ -614,6 +614,31 @@ def render_pep8_errors_e306(msg, _node, source_lines=None):
     yield from render_context(msg.line, msg.line + 2, source_lines)
 
 
+def render_missing_return_statement(msg, _node, source_lines=None):
+    """
+    Render a missing return statements message
+    """
+    # add function name to reported message
+    msg.msg += f" `{_node.name}`"
```
david-yz-liu, line 622: see my above comment about the error message

python_ta/reporters/node_printers.py
```diff
@@ -614,6 +614,31 @@ def render_pep8_errors_e306(msg, _node, source_lines=None):
     yield from render_context(msg.line, msg.line + 2, source_lines)
 
 
+def render_missing_return_statement(msg, _node, source_lines=None):
```
david-yz-liu, line 617: Note: we use `node` instead of `_node` if the parameter is actually being used in the function body.

python_ta/reporters/node_printers.py
```diff
@@ -614,6 +614,31 @@ def render_pep8_errors_e306(msg, _node, source_lines=None):
...
+    """
+    # add function name to reported message
+    msg.msg += f" `{_node.name}`"
+
+    yield from render_context(msg.line - 1, msg.line + 1, source_lines)
+
+    # calculate indentation for the insertion point
+    body = source_lines[msg.line - 1]
+    indentation = len(body) - len(body.lstrip())
+    insertion_text = body[:indentation] + "# INSERT RETURN STATEMENT HERE"
```
david-yz-liu, line 629: I'm thinking about this a bit more and I think we need to give a bit more flexibility in the suggested fix. E.g. "INSERT RETURN STATEMENT HERE (OR BELOW)" in case students want to write a final return statement at the end of the function body. I think it should be possible to detect whether we're at the end of the function body or not, and if so add the extra text to the message

python_ta/checkers/inconsistent_or_missing_returns_checker.py
Raine-Yang-UofT: It turns out this problem is actually more complicated, and the current implementation of rendering indeed has some issues. The problem is that either `fromlineno` or `tolineneo` only indicates the line of the statement, not the end of statements nested inside. One approach I can think of is to recursively search for the last node of each nested block.

image

image

python_ta/checkers/inconsistent_or_missing_returns_checker.py
Raine-Yang-UofT: Sorry that I didn't fully understand your comment here. To be clear, `statement.value is None` is the case for `return`, `return None` would have a value `Const(None)`.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
david-yz-liu: @Raine-Yang-UofT okay I don't think you understood this comment (or the one on line 82 below), so I'll rephase.

On this line (*Line 61*) the condition should check whether the statement is not of the form `return` _or_ `return None`. In other words, set `has_return_value = True` only when encountering a statement of the form `return <expr>`, where `<expr>` is not `Const(None)`.

On *Line 82*, you can keep the condition the way that it is (this is what I meant by "this check is good"). That's because for that condition I only want you to check for `return`, but not `return None`.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,108 @@
...
+            if isinstance(statement, nodes.Return):
+                return_statements[block] = statement
+                if return_statements[block].value is not None:
+                    has_return_value = True
+
+        # check for inconsistent or missing returns
+        if has_return_annotation or has_return_value:
+            for block in return_statements:
+                statement = return_statements[block]
+                end_lines = self._search_for_end_line(block, set())
```
david-yz-liu, line 68: There was a miscommunication here because I misspelled the attribute I wanted you to use.

PythonTA sets a custom attribute called `end_lineno` which should be correct for compound statements. Use that instead of `tolineno` (or `fromlineno`) to get the correct ending line for a statement without needing to do a graph traversal.

Raine-Yang-UofT: Hi Professor, in fact using `end_lineno` attribute wouldn't work. The reason is that during the CFG construction, statements like `if` and loops are not included in `statements` in each block, as shown in the part of graph here.

image

I will try to think of more efficient alternatives of graph traverse. One solution I can current think of is to turn on the `separate-condition-blocks` option of CFGVisitor, but I'm not sure whether it's safe to modify such a global parameter that could affect every CFG-based checker.

david-yz-liu: @Raine-Yang-UofT okay, in this case I do not fully understand what the problem is. Please do the following:

1. Return to the previous approach you were using.
2. Write a test case that shows the incorrect behaviour.

After you push that I'll take a closer look.

Raine-Yang-UofT: @david-yz-liu Hi Professor, the problem is more complicated than I though, as my implementation also works incorrectly at some cases like try/except statements. Before I proceed I want to first make sure where you want the comment `# INSERT RETURN STATEMENT` to appear.

For example, for this function, the outer `for` loop does not guarantee to have a return value
```python
def nested_statements(x):
    result = 0
    for i in range(x):
        result += i
        for j in range(0, 5):
            if j > 3:
                result += 1
            else:
                return result
```
Do we want the comment to be added below the `for` loop (which is how the code behaves if we pass `end_lineno` as the line number)

image

Or, as I'm current trying to implement, adding the comment below the end of loop body

image

david-yz-liu: The expected behaviour is the latter. It's fine if you're a bit stuck on how to proceed. The reason I asked you to do the things in my last message is that this is the best way to provide enough information for me to help you.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,90 @@
...
+                    """
+                    For rendering purpose:
+                    line: the line where the error occurs, used to calculate indentation
+                    end_line: the line to insert the error message
+                    """
+                    self.add_message(
+                        "missing-return-statement",
+                        node=node,
+                        line=block.statements[-1].tolineno,
+                        end_lineno=block.statements[-1].end_lineno,
```
david-yz-liu, line 82: Okay great, the `end_lineno` is good. The issue with the failing tests has to do with how while and for loops are handled. In both cases, when a loop is the last statement before the end of the function, the final block executed is the *while loop condition* or *for loop target*, both of which are the first line of the loop (rather than the last line).

You can detect these two as special cases before calling `self.add_message`. Specifically, check whether `block.statements[-1]` is a while loop condition/for loop target, and if so pass in its *parent*'s `end_lineno` instead (and I think you'll want the parent's `lineno` or `from_lineno` so that you get the right level of indentation signalling that the return should be _outside_ of the loop body).

CHANGELOG.md
```diff
@@ -32,6 +32,10 @@ For more information on these checkers, please see the
 [Pylint release notes](http://pylint.pycqa.org/en/latest/whatsnew/index.html). Note that the above
 list only contains the Pylint checkers enabled by default in PythonTA.
 
+Custom checkers:
+
+- `inconsistent-or-missing-returns-checker`: Provide clearer error messages when reporting missing return statements. This replaces pylint's [R1710](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/inconsistent-return-statements.html) check.
```
david-yz-liu, line 37: List both of the error names here: that is, replace `inconsistent-or-missing-returns-checker` with `inconsistent-returns` and `missing-return-statement`

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,102 @@
...
+                    statement.value is None
+                    or isinstance(statement.value, nodes.Const)
+                    and statement.value.value is None
+                ):
+                    has_return_value = True
+
+        # check for inconsistent or missing returns
+        if has_return_annotation or has_return_value:
+            for block in return_statements:
+                statement = return_statements[block]
```
david-yz-liu, line 71: Use `for block, statement in return_statements.items()` (more idiomatic in Python)

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,102 @@
...
+                if statement is None:
+                    """
```
david-yz-liu, line 73: Do not use docstrings for multiline comments. Instead, use a block of inline comments.

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,102 @@
...
+                    line: the line where the error occurs, used to calculate indentation
+                    end_line: the line to insert the error message
+
+                    For `while` and `for` loops, line and end_line need to set to those of the parent node
+                    to make sure the message is rendered at the end of the loop
+                    """
+                    last_statement = block.statements[-1]
+                    line = last_statement.lineno
+                    end_line = last_statement.end_lineno
+                    if isinstance(last_statement.parent, nodes.While) or isinstance(
```
david-yz-liu, line 84: you can simplify this by passing a tuple as the second argument to `isinstance`)

python_ta/reporters/node_printers.py
```diff
@@ -614,6 +614,36 @@ def render_pep8_errors_e306(msg, _node, source_lines=None):
...
+
+    # calculate indentation for the insertion point
+    body = source_lines[msg.end_line - 1]
+    indentation = len(source_lines[msg.line - 1]) - len(source_lines[msg.line - 1].lstrip())
+
+    # determine whether reaching the end of function
+    function_indentation = len(source_lines[node.lineno - 1]) - len(
+        source_lines[node.lineno - 1].lstrip()
+    )
+    if msg.end_line == node.end_lineno and indentation == function_indentation + 4:
```
david-yz-liu, line 631: This is quite good, but you shouldn't assume that indentation is always 4 spaces. A more robust approach would be to compare this indentation against the indentation of the first statement in the function body

python_ta/checkers/inconsistent_or_missing_returns_checker.py
```diff
@@ -0,0 +1,102 @@
...
+        # get the blocks connected to the end of cfg
+        end = node.cfg.end
+        end_blocks = [edge.source for edge in end.predecessors]
+
+        # gather the return statement of each code block
+        return_statements = {}
+        for block in end_blocks:
+            return_statements[block] = None
+            statement = block.statements[-1]
+            if isinstance(statement, nodes.Return):
```
david-yz-liu, line 58: I realized after looking at this again that this if condition will ignore `raise` statements. However, we should consider (uncaught) raise statements as valid "endings" of function control flow, and _not_ report a "missing return" error in that case.

Please verify whether this current implementation handles raise statement correctly (i.e., following the description of what I said in the previous sentence), and some test cases, and then if necessary fix this implementation to handle raises.

david-yz-liu (approved):

PR #1056, opened 2024-06-21
augment cfg with z3 vars

quiz3:
## Proposed Changes

Augment `ControlFlowGraph` with new protected `_z3_vars` attribute. 

TODO:
- add test cases

<details>
There is an extra empty list getting checked for some reason ... turns out due to an unnecessary accept statement.
IMG_5988

</details>

## Type of Change

| 🔧 _Internal_ (change that _only_ affects developers or continuous integration)         |     X     |

## Checklist

## Questions and Comments

Would this be considered refactoring or perhaps a new feature?

Should it also be a dependency update if it ends up requiring additions/modifications to the project dependencies? Would this designation be in conjunction with another, or are the categories of PRs mutually exclusive?


python_ta/cfg/visitor.py
```diff
@@ -55,7 +56,7 @@ def __getattr__(self, attr: str):
         if attr.startswith("visit_"):
             return self.visit_generic
         else:
-            raise AttributeError(f"'CFGVisitor' object has not attribute '{attr}'")
+            raise AttributeError(f"'CFGVisitor' object has no attribute '{attr}'")
```
quiz3, line 59: Fixes minor typo in error message. Is this an appropriate change to include in this PR?

python_ta/cfg/graph.py
```diff
@@ -2,8 +2,11 @@
 
 from typing import Any, Generator, List, Optional, Set, Tuple, Union
 
+import z3
 from astroid import Break, Continue, NodeNG, Raise, Return
 
+from ..transforms import ExprWrapper
```
quiz3, line 8: `ExprWrapper` is unused. I tried using it but since function definitions are not expressions, I had trouble figuring out how to apply what `ExprWrapper` can do without contorting and modifying its original functionality.

python_ta/cfg/visitor.py
david-yz-liu: It's a good catch, but yeah that should be in a separate PR.

python_ta/cfg/graph.py
david-yz-liu: You can change the type for the `node` parameter in the initializer to `astroid.NodeNG` (matching the actual attribute annotation).

Then in the code below, you can use the `types` attribute of the instance of `ExprWrapper` instead of just a local variable `types`, and similarly use the `ExprWrapper.reduce` method and call it on the parameter names. You might need to add an extra case for `AssignName` nodes, I'm not sure.

python_ta/cfg/graph.py
```diff
@@ -2,8 +2,11 @@
 
 from typing import Any, Generator, List, Optional, Set, Tuple, Union
 
+import z3
```
david-yz-liu, line 5: Because `z3` is an _optional_ dependency, we should be careful with how this is being imported. One strategy is to use a try-except:

```python
try:
    import z3
except ImportError:
    z3 = None
```

And then check whether `z3 is None` below.

david-yz-liu (commented):
@quiz3 good work. I responded to your comments, and for the two questions you asked in the PR description:

1. Somewhat borderline. Mark the attribute as protected (with a leading underscore) and put this under internal change, since we don't want others to rely on it yet.
2. In general, it is possible for a dependency update to be combined with other changes within the same pull request. But I don't think this applies here since no new dependencies should be added.

python_ta/cfg/graph.py
quiz3: Assuming `z3 is None`, what is to be done? 

Is most of the file's body meant to be nested under an if-statement verifying that `z3 is not None`?

quiz3: > Because `z3` is an _optional_ dependency, we should be careful with how this is being imported. One strategy is to use a try-except:
> 
> ```python
> try:
>     import z3
> except ImportError:
>     z3 = None
> ```
> 
> And then check whether `z3 is None` below.

I think I am overcomplicating something because this piece of advice led me down a path to many different points of failure. I don't believe it is related to why the tests are failing, as I have isolated the imports and nothing changed. Please advise as to whether I have taken the right approach.

python_ta/cfg/visitor.py
```diff
@@ -128,6 +129,20 @@ def visit_functiondef(self, func: nodes.FunctionDef) -> None:
         self._current_cfg.start.add_statement(func.args)
         func.cfg_block = self._current_cfg.start
 
+        if func.name != "__main__":
```
david-yz-liu, line 132: In response to your question about the `z3` import: overall you're on the right track, though the code can be simplified. The simplest way to think about it is that *this specific condition should have an `and z3 is not None` attached*. So that if `z3` fails to import, the code inside this if branch (which is responsible for populating the `_z3_vars` attribute) won't execute, but the rest of the existing code of this visitor class will.

However, you are correct in pointing out the issues that the imports span multiple files (e.g., this file doesn't import `z3` directly, just `ExprWrapper`). There are a few different ways to handle this elegantly. One option is to use the try-except strategy but set `ExprWrapper = None` if there's an `ImportError` raised, which could be a "nested error" because `ExprWrapper.py` imports `z3`.

Another options, which is my preference though slightly more involved, is to move this logic into a `ControlFlowGraph` method. Noting that the key AST node at play here is the `Arguments` node, one can imagine a method called `ControlFlowGraph.add_arguments` that takes an `Arguments` node and (1) adds it to the starting block (currently done in line 129 above) as well as extracts `z3` variables from those arguments using `ExprWrapper`. This would be the same overall logic as you have currently here, just moved into a single file.

david-yz-liu: Oh and as a separate comment, I'm not actually sure why the `__main__` check is necessary. This method is `visit_functiondef`, so should only be called on `FunctionDef` nodes, and not the root `Module` node (or the `if __name__ == '__main__'` block).

python_ta/cfg/graph.py
```diff
@@ -1,6 +1,11 @@
 from __future__ import annotations
 
-from typing import Any, Generator, List, Optional, Set, Tuple, Union
+from typing import Any, Dict, Generator, List, Optional, Set
+
+try:
+    from z3 import ExprRef
+except ImportError:
+    ExprRef = None
```
david-yz-liu, line 8: Great, you can set this to `Any` instead

quiz3: Do you mean that I should change the type annotation for `_z3_vars` to `_z3_vars: Dict[str, Any]` instead of `_z3_vars: Dict[str, ExprRef]`?

python_ta/cfg/visitor.py
```diff
@@ -128,6 +129,20 @@ def visit_functiondef(self, func: nodes.FunctionDef) -> None:
...
         func.cfg_block = self._current_cfg.start
 
+        if z3 is not None:
+            # Parse types
+            expr = ExprWrapper(func)
+            z3_vars_list = expr.parse_function_def(func)
+            # Confirm that type list is given
+            if isinstance(z3_vars_list, list):
+                self._current_cfg._z3_vars.update(
+                    {z3_var.decl().name(): z3_var for z3_var in z3_vars_list}
```
david-yz-liu, line 139: The dictionary's keys should use the `func.args` argument names, not the names of the variables. In principle these could be different, and it's good to stick with the source of truth where possible.

That said, your current implementation of `parse_function_def` skips parameters if there's no type annotation, or if the type annotation can't be inferred, so the returned list could be of a different length than `func.args`. You can fix this by instead having the function return a dictionary mapping the function argument names to `z3` variables, and then simply passing that return value directly to `self._current_cfg._z3_vars.update`.

python_ta/cfg/visitor.py
```diff
@@ -128,6 +129,20 @@ def visit_functiondef(self, func: nodes.FunctionDef) -> None:
         self._current_cfg.start.add_statement(func.args)
         func.cfg_block = self._current_cfg.start
 
+        if z3 is not None:
+            # Parse types
+            expr = ExprWrapper(func)
+            z3_vars_list = expr.parse_function_def(func)
+            # Confirm that type list is given
+            if isinstance(z3_vars_list, list):
```
david-yz-liu, line 137: the return type of the `parse_function_def` method is a list, so this check is redundant

david-yz-liu (commented):
@quiz3 great work, I think you're close to wrapping up the implementation, except I noted you did not move the code into the `ControlFlowGraph` class (my preferred approach from my previous comment). You should do so.

You should also add some test cases; I'd recommend an approach of creating control flow graphs for different code snippets and then accessing the `_z3_vars` attribute to assert that the expect variables have been added.

python_ta/cfg/graph.py
quiz3: Oh, I didn't realize Any could be a value and not just a type annotation... thanks!

python_ta/transforms/ExprWrapper.py
```diff
@@ -1,9 +1,13 @@
-from typing import Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 import astroid
-import z3
 from astroid import nodes
 
+try:
```
quiz3, line 6: I did this because if `z3` were set to `Any` or `None` in some circumstances, referencing `z3.ExprRef` or any of `z3`'s other attributes would raise an error.

quiz3:
I have added a screenshot to the PR message showing a key point that may be at the root of the failures. There is an extra empty list item getting checked that should not be there.

python_ta/cfg/visitor.py
```diff
@@ -126,8 +126,9 @@ def visit_functiondef(self, func: nodes.FunctionDef) -> None:
             )
         )
 
-        self._current_cfg.start.add_statement(func.args)
-        func.cfg_block = self._current_cfg.start
+        # Current CFG block is self._current_cfg.start while initially creating the function cfg
+        func.args.accept(self)
```
david-yz-liu, line 130: This isn't necessary, and I suspect is now causing a problem with the tests. The `Arguments` node is being added in the `add_arguments` method

david-yz-liu: Another note is that under `patches/transforms.py`, you can see that any exceptions when using `CFGVisitor` are being caught but unreported. You could try removing the try-except and just keeping the `ast.accept` call and seeing if you see more useful errors when you run the tests

python_ta/transforms/ExprWrapper.py
```diff
@@ -139,24 +160,44 @@ def apply_bool_op(self, op: str, values: Union[z3.ExprRef, List[z3.ExprRef]]) ->
...
+    def parse_arguments(self, node: astroid.Arguments) -> Dict[str, ExprRef]:
+        """Convert an astroid Arguments node's parameters to z3 variables."""
+        z3_vars = {}  # initialize mapping of z3 variables
+
+        annotations = node.annotations
+        arguments = node.args
+        for ann, arg in zip(annotations, arguments):
+            if ann is None:
+                continue
+            inferred = ann.inferred()
```
david-yz-liu, line 192: I looked into the failing tests a bit more. Instead of using `.inferred`, use the `pylint.util.safe_infer` function, which returns `None` or a single inferred value (whereas `.inferred` returns a list).

python_ta/cfg/visitor.py
```diff
@@ -126,9 +126,9 @@ def visit_functiondef(self, func: nodes.FunctionDef) -> None:
             )
         )
 
-        self._current_cfg.start.add_statement(func.args)
-        func.cfg_block = self._current_cfg.start
-        func.cfg = self._current_cfg
+        # Current CFG block is self._current_cfg.start while initially creating the function cfg
+        self._current_cfg.add_arguments(func.args)
+        # func.args.accept(self)
```
david-yz-liu, line 131: you can delete this line of code

python_ta/transforms/ExprWrapper.py
david-yz-liu: Overall now that we've settled on the import strategy in the `graph.py` file, you should undo the changes you mentioned in this comment. More specifically:

1. This file (`ExprWrapper.py`) should *assume* that `z3` can always be imported correctly.
2. In `graph.py`, make the importing of `ExprWrapper` to be conditioned on whether the import of `z3` was successful. This would look like

```python
try:
    from z3 import ExprRef
    from ..transforms import ExprWrapper
except ImportError:
    ExprRef = Any
    ExprWrapper = Any
```

python_ta/transforms/ExprWrapper.py
```diff
@@ -52,21 +66,29 @@ def reduce(self, node: astroid.NodeNG = None) -> z3.ExprRef:
             node = node.value
         elif isinstance(node, nodes.Name):
             node = self.apply_name(node.name)
+        elif isinstance(node, nodes.Assign):
+            node = self.reduce(node.targets[0])
+        elif isinstance(node, nodes.AssignName):
+            node = self.apply_name(node.name)
+        # elif isinstance(node, nodes.Arguments):
```
david-yz-liu, line 73: I see the challenge here; it's fine to just remove these cases.

python_ta/transforms/ExprWrapper.py
```diff
@@ -139,24 +161,43 @@ def apply_bool_op(self, op: str, values: Union[z3.ExprRef, List[z3.ExprRef]]) ->
...
         right = self.reduce(right)
 
         return self.apply_bin_op(left, op, right)
 
-    def parse_bool_op(self, node: astroid.BoolOp) -> z3.ExprRef:
+    def parse_bool_op(self, node: astroid.BoolOp) -> ExprRef:
         """Convert an astroid BoolOp node to a z3 expression."""
         op, values = node.op, node.values
         values = [self.reduce(x) for x in values]
-
```
david-yz-liu, line 161: revert this change

python_ta/transforms/ExprWrapper.py
```diff
@@ -139,24 +161,43 @@ def apply_bool_op(self, op: str, values: Union[z3.ExprRef, List[z3.ExprRef]]) ->
...
+        z3_vars = {}  # initialize mapping of z3 variables
+
+        annotations = node.annotations
+        arguments = node.args
+        for ann, arg in zip(annotations, arguments):
+            if ann is None:
+                continue
+
+            inferred = safe_infer(ann)
+            if inferred is None or inferred is astroid.Uninferable:
```
david-yz-liu, line 195: A more specific check should be the type of `inferred`: is it a `Name` node (or something else)? The next statement accesses a `name` attribute, which makes an assumption about the type of inferred node.

python_ta/transforms/ExprWrapper.py
```diff
@@ -139,24 +161,43 @@ def apply_bool_op(self, op: str, values: Union[z3.ExprRef, List[z3.ExprRef]]) ->
...
+                continue
+
+            self.types[arg.name] = inferred.name
+
+            if arg.name in self.types and self.types[arg.name] in ["int", "float", "bool"]:
```
david-yz-liu, line 200: Use a set literal instead of a list literal when just checking membership

david-yz-liu (commented):
@quiz3 good work, now that all of the tests are passing I did another review.

Please makes these changes and do another self-review of your code, and then re-request a review when you're done.

quiz3:
> @quiz3 good work, now that all of the tests are passing I did another review.
> 
> Please makes these changes and do another self-review of your code, and then re-request a review when you're done.

I have made the changes and performed a self-review. 

I am working on writing tests but am struggling to find the lines that require test coverage. For some reason, the strategies I remember using before no longer show which lines are uncovered and the coveralls reports will not load.

I will continue to look into this (perhaps a browser issue or a temporary server-side issue)...

python_ta/transforms/ExprWrapper.py
```diff
@@ -1,8 +1,9 @@
 from typing import Dict, List, Optional, Set, Tuple, Union
 
 import astroid
-import z3
```
david-yz-liu, line 4: From my previous comment: you should revert the `z3`-related import changes in this file. They aren't necessary, as we (continue to) assume that `z3` is imported correctly in this file.

Look carefully at the diff of this file. There shouldn't be any changes where `z3.` is removed from a type name, etc. Make sure to verify this after making these changes to check your work.

python_ta/transforms/ExprWrapper.py
```diff
@@ -26,13 +27,21 @@ class ExprWrapper:
     node: astroid.NodeNG
     types: Dict[str, str]
 
-    def __init__(self, expr: nodes.Expr, types=None):
-        self.node = expr.value
-        if types is None:
```
david-yz-liu, line 31: You should not change the default value of `types`. In general, this was not a good change to make since it was unrelated to your PR.

In this particular case, see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.

python_ta/transforms/ExprWrapper.py
```diff
@@ -26,13 +27,21 @@ class ExprWrapper:
...
 
-    def reduce(self, node: astroid.NodeNG = None) -> z3.ExprRef:
+        if isinstance(node, astroid.Expr):
+            self.node = node.value  # take node attribute to be the value of the expression
+        elif isinstance(node, astroid.Assign):
+            self.node = node.value  # take node attribute as expression (right side) of assignment
+        elif isinstance(node, astroid.Arguments):
+            self.node = node  # take node attribute to be the function declaration node itself
+        else:
+            raise Z3ParseException(
```
david-yz-liu, line 40: raise the more general `ValueError` type; this method isn't doing any "z3 parsing".

python_ta/transforms/ExprWrapper.py
```diff
@@ -149,30 +158,48 @@ def apply_bool_op(self, op: str, values: Union[z3.ExprRef, List[z3.ExprRef]]) ->
...
-    def parse_container_op(
-        self, node: Union[nodes.List, nodes.Set, nodes.Tuple]
-    ) -> List[z3.ExprRef]:
+    def parse_container_op(self, node: Union[nodes.List, nodes.Set, nodes.Tuple]) -> List[ExprRef]:
         """Convert an astroid List, Set, Tuple node to a list of z3 expressions."""
         return [self.reduce(element) for element in node.elts]
+
+    def parse_arguments(self, node: astroid.Arguments) -> Dict[str, ExprRef]:
+        """Convert an astroid Arguments node's parameters to z3 variables."""
+        z3_vars = {}  # initialize mapping of z3 variables
```
david-yz-liu, line 188: this comment is unnecessary

tests/test_transforms/test_expr_wrapper.py
```diff
@@ -0,0 +1,27 @@
...
+
+from astroid import parse
+
+from python_ta.transforms import ExprWrapper, Z3ParseException
+
+
+def test_expr_wrapper_invalid_node() -> None:
+    exception_caught = False
+
+    try:
```
david-yz-liu, line 11: look up `pytest.raises` and use that here

tests/test_transforms/test_expr_wrapper.py
```diff
@@ -0,0 +1,27 @@
...
+    except Z3ParseException:
+        exception_caught = True
+
+    assert exception_caught
+
+
+def test_expr_wrapper_assignment() -> None:
+    assignment = "n = 2 + 3"
+
+    mod = parse(assignment)
```
david-yz-liu, line 22: use `astroid.extract_node` for simpler AST processing

tests/test_z3_visitor.py
```diff
@@ -149,6 +159,25 @@ def not_in_empty_tuple(x: int):
 expected_list = [arithmetic_expected, boolean_expected, container_expected]
 
 
+def test_z3_vars():
```
david-yz-liu, line 162: This test should have a much more specific name and description

tests/test_z3_visitor.py
```diff
@@ -149,6 +159,25 @@ def not_in_empty_tuple(x: int):
...
+    Test the z3 variables
+    """
+    mod = astroid.parse(z3_vars_example)
+    node = mod.body[0]
+
+    cfg = ControlFlowGraph()
+    cfg.add_arguments(node.args)
+
+    # Note that this assert implicitly includes the assertion that 'a' not in cfg._z3_vars
+    assert all(
```
david-yz-liu, line 173: This test can be improved by making the assertions more specific.

First, you should have a separate assertion for each parameter, and that assertion should also assert the *type* of z3 variable (so not just `z3.ExprRef`).

Second, assert that the size of the dictionary is 3. In general size is a good thing to check for to ensure no extraneous data was added.

tests/test_z3_visitor.py
david-yz-liu: Also, move this test to after all of the other tests (which belong together in a block; see changes in #1061.

david-yz-liu (approved):

PR #1061, opened 2024-07-02
feat: add support for container literals and in operator in ExprWrapper

Raine-Yang-UofT:
## Proposed Changes

Extend `ExprWrapper` to support `List/Set/Tuple` literals and `in/not in` operators.
...

<details>

</details>

## Type of Change

| 🔧 _Internal_ (change that _only_ affects developers or continuous integration)         |     X     |

## Checklist

## Questions and Comments

### DIFF
--- a/tests/test_z3_visitor.py
+++ b/tests/test_z3_visitor.py
@@ -56,14 +56,15 @@ def test_bool_constraints():
         assert solver.check() == z3.sat
 
 
-def test_list_constraints():
+def test_container_constraints():
     z3v = Z3Visitor()
     code = """
     def f(x: int):
         '''
         Preconditions:
             - x in [1, 2, 3]
-            - x not in [4, 5, 6, 7, 8]
+            - x not in {4, 5, 6, 7, 8}
+            - x in (1, 2)
         '''
         pass
     """
@@ -72,7 +73,11 @@ def test_list_constraints():
     # Declare variables
     x = z3.Int("x")
     # Construct expected
-    expected = [z3.Or(x == 1, x == 2, x == 3), z3.And(x != 4, x != 5, x != 6, x != 7, x != 8)]
+    expected = [
+        z3.Or(x == 1, x == 2, x == 3),
+        z3.And(x != 4, x != 5, x != 6, x != 7, x != 8),
+        z3.Or(x == 1, x == 2),
+    ]
     actual = function_def.z3_constraints
     assert actual != []
     for e, a in zip(expected, actual):
--- a/tests/test_z3_visitor.py
+++ b/tests/test_z3_visitor.py
@@ -83,21 +83,33 @@ container_list = [
         pass
     """,
     """
-    def in_empty_container(x: int):
+    def in_empty_list(x: int):
         '''
         Preconditions:
             - x in []
-            - x in set()
+        '''
+        pass
+    """,
+    """
+    def in_empty_tuple(x: int):
+        '''
+        Preconditions:
             - x in ()
         '''
         pass
     """,
     """
-    def not_in_empty_container(x: int):
+    def not_in_empty_list(x: int):
         '''
         Preconditions:
             - x not in []
-            - x not in set()
+        '''
+        pass
+    """,
+    """
+    def not_in_empty_tuple(x: int):
+        '''
+        Preconditions:
             - x not in ()
         '''
         pass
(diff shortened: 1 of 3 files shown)


tests/test_z3_visitor.py
```diff
@@ -54,3 +54,33 @@ def f(x: bool, y: bool, z: bool, a):
         solver = z3.Solver()
         solver.add(e == a)
         assert solver.check() == z3.sat
+
+
+def test_container_constraints():
```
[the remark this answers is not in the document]
david-yz-liu: This is better, but can still be improved by taking advantage of pytest's test parametrization. Check out some of the past PRs from your teammates if you want to see some examples of this in action.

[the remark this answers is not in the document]
Raine-Yang-UofT: Hi Professor, sorry that I didn't fully understanding how you want me to use pytest's test parameterization. Do we want to only parameterize the test for `list/set/tuple` or all the tests in the module. Currently I'm using test parameterization inside `test_container_constraints` function, and I will try to incorporate more code into the inner `check_constraint` function.

One difficulty for parameterizing all test cases is that we cannot duplicate the z3 variables we define. For the tests we need to define z3 variables such as `x = z3.Int("x")` for the expected z3 constraints. If we want to move the list of code and expected constraints outside the functions these variables will become global, so we cannot reuse a variable for a different type once its type is defined. This is also the reason why I'm keeping the parameterized test within the outer `test_container_constraints` to limit the scope of variables.

tests/test_z3_visitor.py
```diff
@@ -41,16 +30,127 @@ def f(x: bool, y: bool, z: bool, a):
...
-        solver.add(e == a)
-        assert solver.check() == z3.sat
+    return function_def.z3_constraints
+
+
+@pytest.mark.parametrize("code, expected", zip(code_list, expected_list))
+def test_constraint(code, expected):
+    for function, function_expected in zip(code, expected):
+        function_actual = _get_constraints_from_code(function)
+        assert len(function_actual) == len(function_expected)
```
Raine-Yang-UofT, line 152: Although checking the length of expected and actual expressions to be equal has been our existing approach, and is used to ensure `zip` function works correctly, I'm not sure this is the best practice to perform the test. In principle, we do not need the number of expressions to be equal to that asseted, or the expressions appear in the exact same order, as long as the two list of expressions are equivalent. Our approach seems to be testing against our specifc implementation instead of the module's feature.

david-yz-liu: While the relevant module functionality is to convert the preconditions into Z3 expressions, it is appropriate for the module to do this conversion as a one-to-one mapping (i.e., without further simplification) until we decide to change this.

More to the point, however, this reveals a limitation on our current testing approach. The changes you made were to `ExprWrapper`, which currently is only tested indirectly through the tests in this file for `Z3Visitor`. Certainly I would not accept any simplification in this file until there were unit tests for `ExprWrapper` itself. (I'm not saying to add those in this PR, but rather to explain why _even if I approve simplification of the precondition expressions_ generally, I would not approve changing the tests in this file without additional tests added.)

tests/test_z3_visitor.py
```diff
@@ -41,16 +30,127 @@ def f(x: bool, y: bool, z: bool, a):
...
+# expected container expressions
+x = z3.Int("x")
+container_expected = [
+    [z3.Or(x == 1, x == 2, x == 3)],
+    [z3.Or(x == 1, x == 2, x == 3)],
+    [z3.Or(x == 1, x == 2, x == 3)],
+    [z3.And(x != 1, x != 2, x != 3)],
+    [z3.And(x != 1, x != 2, x != 3)],
+    [z3.And(x != 1, x != 2, x != 3)],
+    [z3.BoolVal(False), z3.BoolVal(False)],
```
david-yz-liu, line 129: First, each of these should be in their own function (as you've done for all of the above test cases in `container_expected`). Same with the "not in empty" test below.

Second (more as a comment), I'm surprised that there are three precondition expressions but only two Z3 expressions here. Can you explain why that is?

Raine-Yang-UofT: Hi Professor, sorry that I wan't quite clear what do you mean by the first comment. I didn't see how the test cases for empty containers differ from the ones above.

About the second problem, sorry that I wasn't being careful enough. The precondition for empty set turns out to be ignored since PythonTA couldn't infer the return type of `set()` function and skipped creating z3 expression for this precondition. However, I cannot fix this problem with modifications in `ExprWrapper` as the problem lies in the `parse_assertions` function in `contracts/__init__.py`.  Currently we may not be able to create the test cases with empty sets.

david-yz-liu: @Raine-Yang-UofT sure, what I meant was that I'd like to see each precondition as a separate test case, i.e., splitting up the `x in []` and `x in ()` preconditions into separate test cases (and function). This aligns with the other test cases you've added, which each have one precondition per function.

Good point about the usage of `set()`. It's fine to not include test cases for empty sets in this PR.

Raine-Yang-UofT: @david-yz-liu Thanks, I have updated the tests. In addition, perhaps I can submit a issue concerning the problem with empty sets.

david-yz-liu (approved):

PR #1075, opened 2024-08-08
Snapshot save feature

lana-w:
## Proposed Changes

Implemented an additional (optional) save parameter to the snapshot function. If save is True, snapshot calls on the memory-viz CLI, using the memory_viz_args and memory_viz_version arguments to dictate what is passed in to memory-viz and what version to use. Returns an SVG of the snapshot, either to a specified path, or to stdout by default. 

## Type of Change

| 🚨 _Breaking change_ (fix or feature that would cause existing functionality to change) |X|

## Checklist

## Questions and Comments

August 7: I'm leaving this as a draft pull request while I iron out some of the details of this implementation, and as I am waiting to finalize tests until after this is finalized. All four combinations of stdin, file path input, stdout, and path output are functional with the latest version with this implementation.

Some Questions:
- MemoryViz Versions : having the memory_viz_version parameter is handy, but it changes details of the implementation a little bit. Since older versions don't support passing in the json via stdin, to allow the user to specify an older version means that we would need snapshot to alternately write the json to a file, and then pass the path to that file to the CLI. I am wondering if you would like me to implement that, or if it is better to not allow the specification of a memory-viz version at all, since documentation for older versions is also not accessible.
- What should snapshot do in the case of the user not specifying an output path in the memory_viz_args? the memviz CLI provides the svg contents as stdout which I've captured in results and printed, but I'm not sure if printing is the best way to get that svg content to the user.

August 12 Update:
I have updated the code with the requested changes. Two notes in particular:
- For the stdout and stderr for subprocess.run, I have them set to sys.stdout and sys.stderr as this ensures they are outputted by default into the terminal. I can set them to subprocess.PIPE, but I wasn't sure how they would then be accessible to the user, as they didn't seem to appear in the terminal, even when the output was piped into a file. Let me know if it should be subprocess.PIPE regardless and I'll make that change ASAP!
- I was only able to remove the usage of shell=True by specifying the executable as the path for npx, which I get using shutil.which. Documentation doesn't seem to discourage the use of executable as it does shell, but let me know if there are any issues with this as well.

Note that I haven't pushed any test updates as I want to ensure my implementation is logical first!

August 17 Update:
I have a full testing suite up, with two considerations:
- The precommit hooks keep changing the svg files that I am using to do snapshot testing and thus causing the tests to fail since the expected output no longer matches the produced output, any suggested workarounds here? 
- Testing the creation of the svg file using snapshot's save then means that when the test is run, a file is created in the repository, which seems not ideal. I'm wondering it it makes sense to have the test manually delete the file upon completion? I would use pytest's tmp_path fixture, but considering that snapshot is now being called in an external file, I'm unsure of how to then get the file path of the svg created so that it can be read and compared with the expected output.

A note on the above, with the inclusion of the test that checks that no file is created when snapshot's save is False, this causes an issue at the moment since I have both testing on the same file path. So since the test that checks the file is created doesn't delete it at the moment, then the next time the tests are run, test_snapshot_no_save_file will fail. 

August 19 Update:
Have now made it so the test which creates a file does so with the pytest temp feature. Have also pushed the snapshot svgs that are used to confirm output after excluding them from the precommit hooks. 

I believe the only issue now is with my updates to the test.yml files, as the CI tests are no longer running on this PR.


python_ta/debug/snapshot.py
```diff
@@ -29,11 +34,17 @@ def get_filtered_global_variables(frame: FrameType) -> dict:
     return {"__main__": true_global_vars}
 
 
-def snapshot():
+def snapshot(
+    save: Optional[bool] = False,
```
david-yz-liu, line 38: You're misunderstanding the `Optional` type annotation; it doesn't mean "optional parameter", it means "could be `None`". So you should use `Optional` for memory_viz_args, but not the other two.

python_ta/debug/snapshot.py
```diff
@@ -29,11 +34,17 @@ def get_filtered_global_variables(frame: FrameType) -> dict:
...
+    memory_viz_args: Optional[list[str]] = None,
+    memory_viz_version: Optional[str] = "latest",
+):
     """Capture a snapshot of local variables from the current and outer stack frames
     where the 'snapshot' function is called. Returns a list of dictionaries,
     each mapping function names to their respective local variables.
     Excludes the global module context.
+
+    When save is True, a MemoryViz-created svg is produced
```
david-yz-liu, line 47: This is good, but make sure to document the other two arguments as well.

python_ta/debug/snapshot.py
```diff
@@ -47,7 +58,27 @@ def snapshot():
...
+            input=str(json_compatible_vars),
+            shell=True,
+            check=True,
+            capture_output=True,
+            text=True,
+        )
+
+        print(result.stdout)
+
+        return variables
```
david-yz-liu, line 78: This can be outdented so that you don't need the `else` branch

python_ta/debug/snapshot.py
```diff
@@ -47,7 +58,27 @@ def snapshot():
 
         frame = frame.f_back
 
-    return variables
+    if save:
+        json_compatible_vars = snapshot_to_json(variables)
+        command = ["npx", "memory-viz@" + memory_viz_version]
```
david-yz-liu, line 63: As a pre-processing step, if `memory_viz_version` is not `"latest"`, parse the version information (https://packaging.pypa.io/en/latest/version.html#packaging.version.Version). If less than 0.3.1, print a warning message (use the `logging` module, see other parts of PythonTA) saying that PythonTA only supports MemoryViz versions 0.3.1 and later.

python_ta/debug/snapshot.py
```diff
@@ -47,7 +58,27 @@ def snapshot():
...
+        if memory_viz_args:
+            command.extend(memory_viz_args)
+
+        result = subprocess.run(
+            command,
+            input=str(json_compatible_vars),
+            shell=True,
```
david-yz-liu, line 70: we should avoid using `shell=True`

python_ta/debug/snapshot.py
```diff
@@ -47,7 +58,27 @@ def snapshot():
...
-    return variables
+    if save:
+        json_compatible_vars = snapshot_to_json(variables)
+        command = ["npx", "memory-viz@" + memory_viz_version]
+        if memory_viz_args:
+            command.extend(memory_viz_args)
+
+        result = subprocess.run(
+            command,
+            input=str(json_compatible_vars),
```
david-yz-liu, line 69: Use `json.dumps` instead of `str` here

python_ta/debug/snapshot.py
```diff
@@ -133,11 +164,11 @@ def process_value(val: Any) -> int:
...
-                "name": frame_name,
+                "name": str(frame_name),
                 "id": None,
                 "value": frame_variables,
             }
             json_data.append(json_object_frame)
 
     json_data.extend(value_entries)
-    return json_data
+    return json.dumps(json_data)
```
david-yz-liu, line 174: This change is not appropriate; we should be returning the raw data as a `list[dict]`, not a string

python_ta/debug/snapshot.py
```diff
@@ -84,7 +115,7 @@ def process_value(val: Any) -> int:
             if isinstance(val, (list, set, tuple)):
                 element_ids = [process_value(element) for element in val]
                 value_entry = {
-                    "type": type(val).__name__,
+                    "type": str(type(val).__name__),
```
david-yz-liu, line 118: I'm not sure why these changes are necessary; in what cases is `__name__` not already a string?

python_ta/debug/snapshot.py
```diff
@@ -133,11 +164,11 @@ def process_value(val: Any) -> int:
 
             json_object_frame = {
                 "type": ".frame",
-                "name": frame_name,
+                "name": str(frame_name),
```
david-yz-liu, line 167: See above comment; `frame_name` should always be a string

david-yz-liu (commented):
@lana-w good work, I left some comments and addressed your questions in them as well.

david-yz-liu (commented):
@lana-w this is great work. Using `shutil` to get the executable path is good, as is the use of `sys.stdout`/`sys.stderr`.

Please go ahead and finish up this work, adding new test cases, etc.

Also, one other bit of error checking: if `npx` cannot be found, we should display another error message to the user, something like "Could not find npx on your system. Have you installed Node.js?"

david-yz-liu:
@lana-w thanks for the update. I can provide some quick answers to your questions:

1. In general, you can configure pre-commit to *ignore specific files for checks* using the `exclude` key. See https://pre-commit.com/#regular-expressions, for example.
2. Indeed, using a temporary file fixture is best. You can pass in an arbitrary path to memory-viz using the `--output` command-line argument.
3. For testing that _no_ file is created, you can just check that no svg files exist in the current working directory. Of course this isn't perfect since that doesn't check that no file is created _anywhere_, but it's better than nothing.

.github/workflows/test.yml
```diff
@@ -13,13 +13,19 @@ jobs:
...
       - uses: actions/checkout@v4.1.7
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5.1.1
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Set up Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+        run: npm ci
```
david-yz-liu, line 28: @lana-w this needs to be written as a separate step: see https://github.com/david-yz-liu/memory-viz/blob/master/.github/workflows/ci.yml#L21-L22

david-yz-liu (approved):
@lana-w great work! I made one small documentation change myself and now will merge this in. 🎉

PR #1084, opened 2024-09-21
Check cfg edge constraints

Raine-Yang-UofT:
## Proposed Changes

Add `is_feasible` attribute to `CFGEdge` and implemented update to edge feasibility based on the `z3_constraints` attribute.
Add handling of `AugAssign` node in creating edge Z3 constraints.

Existing issue:
The current implementation cannot correctly handle variable reassignment after a conditional statement inside a loop, as shown in the test case `test_feasible_reassignment_in_while` in `test_edge_feasibility`

...

<details>

</details>

## Type of Change

| 🔧 _Internal_ (change that _only_ affects developers or continuous integration)         |    X      |

## Checklist

## Questions and Comments


python_ta/cfg/graph.py
```diff
@@ -274,11 +277,32 @@ def update_edge_z3_constraints(self) -> None:
 
                 # traverse into target node
                 for node in edge.target.statements:
-                    if isinstance(node, Assign):
```
david-yz-liu, line 277: This change should be extracted into a separate pull request. It is separate from checking edge feasibility; it is extending the existing capability of the Z3 constraint updating, which is a good change to make but is separate.

In your new PR, you should make sure to add an entry to the changelog describing this, as well as add new test cases for just this change.

tests/test_cfg/test_edge_feasibility.py
```diff
@@ -0,0 +1,302 @@
...
+    assert all(edge.is_feasible == expected for edge, expected in zip(paths[1], expected_if_path2))
+    assert all(
+        edge.is_feasible == expected for edge, expected in zip(paths[2], expected_other_path1)
+    )
+    assert all(
+        edge.is_feasible == expected for edge, expected in zip(paths[3], expected_other_path2)
+    )
+
+
+# TODO: determine and implement the expected behavior of this test case
```
david-yz-liu, line 266: It's okay to not include this test, but please don't include the commented out code in this PR. Instead, save it and include it in a future PR.

python_ta/cfg/graph.py
```diff
@@ -3,12 +3,14 @@
 from typing import Any, Dict, Generator, List, Optional, Set
 
 try:
+    import z3
```
david-yz-liu, line 6: We should use a consistent import style here. Either import `Solver` on the line below, or only import `z3` and then modify the usages of `Z3_OP_UNINTERPRETED` to `z3.Z3_OP_UNINTERPRETED`, etc. consistently throughout.

python_ta/cfg/graph.py
```diff
@@ -274,11 +277,32 @@ def update_edge_z3_constraints(self) -> None:
...
+        """
+        if not z3_dependency_available:
+            return
+
+        for path_id, path in enumerate(self.get_paths()):
+            solver = z3.Solver()
+            for edge in path:
+                solver.add(edge.z3_constraints[path_id])
+                if solver.check() == z3.sat:
+                    edge.is_feasible = True
```
david-yz-liu, line 305: Overall, I'd like the default value of `edge.is_feasible` to be `True` (set in the initializer), and then set to `False` when the Z3 solver gives "unsatisfiable" for every path that uses the edge.

This is the more conservative default value, in the sense that later on, we'll modify checks to ignore edges that are infeasible, so we want to avoid at all costs the error of marking an edge as infeasible when it's actually feasible.

Your current approach makes the updating a bit awkward because you're iterating across all paths, and then calling `solver.check` for each path's edge. You should be able to do it the other way, iterating across all *edges* and then calling `solver.check` for each edge's set of path constraints. ("if all constraints are unsatisfiable, `edge.is_feasible = False`")

python_ta/cfg/graph.py
```diff
@@ -16,12 +24,15 @@
...
     Z3ParseException = Any
+    Solver = Any
+    unsat = Any
     z3_dependency_available = False
 
 from astroid import (
     Arguments,
     Assign,
     AssignName,
+    AugAssign,
```
david-yz-liu, line 35: This is no longer used

david-yz-liu (commented):
@Raine-Yang-UofT okay great. I left one minor comment, but also note in my original instructions I asked you to do 3 things. You've done the first two now, and should add the third.

python_ta/cfg/cfg_generator.py
```diff
@@ -13,6 +13,8 @@
 from astroid import nodes
 from astroid.builder import AstroidBuilder
 
+from python_ta.transforms.z3_visitor import Z3Visitor
```
david-yz-liu, line 16: As with my previous comments, here you should use a relative import (`.transforms.z3_visitor`) and group this with the imports below.

But also similar to `graph.py`, we need to check for whether Z3 has been successfully imported or not, and if not then make sure the previous functionality works correctly. You can test this by uninstalling Z3 from your Python environment and then calling `generate_cfg`.

Raine-Yang-UofT: Hi Professor, I tested that when Z3 is uninstalled, `z3_visitor` would raise an error as we did not check for Z3 dependency in its imports. I suggest we should modify `z3_visitor` such that if Z3 dependency is not available, invoking the visitor would do nothing. This may be more convenient than checking for Z3 dependency whenever we want to use `Z3Visitor`. Should I make this change in a separate PR or include it in this PR.

david-yz-liu: Hi @Raine-Yang-UofT, please use the approach I described in my original comment. I don't agree with the approach of `Z3Visitor` "doing nothing" if `z3` isn't installed, I'd much rather it not be invoked at all.

Raine-Yang-UofT: @david-yz-liu No problem, I have added a check for z3 dependency in generating CFG. Now the CFG generation works correctly even with Z3 uninstalled. However, as the import of z3 are not checked in `Z3Visitor` and `Z3Parser`, we need to fix these two modules for PythonTA to function correctly without Z3. Let me know if I should propose this change in a new PR.

python_ta/cfg/cfg_generator.py
```diff
@@ -13,6 +13,7 @@
 from astroid import nodes
 from astroid.builder import AstroidBuilder
 
+from ..transforms.z3_visitor import Z3Visitor
```
david-yz-liu, line 16: @Raine-Yang-UofT I'm not sure how you tried to test what I asked for, but `transforms.z3_visitor` imports `z3`, and so will raise an `ImportError` if `z3` isn't installed.

david-yz-liu (approved):

PR #1086, opened 2024-09-23
Implement Snapshot Tracer

leowrites:
## Proposed Changes

- Add a context manager `python_ta.debug.SnapshotManager`

## Type of Change

| ✨ _New feature_ (non-breaking change that adds functionality)                          |     x     |
| 🐛 _Bug fix_ (non-breaking change that fixes an issue)                                  |         |
| ♻️ _Refactoring_ (internal change to codebase, without changing functionality)          |         |
| 🚦 _Test update_ (change that _only_ adds or modifies tests)                            |       x   |
| 📚 _Documentation update_ (change that _only_ updates documentation)                    |      x    |

## Checklist

## Questions and Comments


docs/debug/index.md
```diff
@@ -1,8 +1,13 @@
-# Loop and Recursion Tracing
+# Debugging with Tracing
 
-This page describes an additional PythonTA feature: print-based debugging for loops and recursion.
-This feature makes tracing easier by printing the state of each loop iteration or recursive function call in a nicely-formatted table using the [tabulate] library.
-This functionality is found in the `python_ta.debug` submodule.
+This page describes two additional PythonTA features:
+
```
leowrites, line 4: Remove this space

docs/debug/index.md
```diff
@@ -1,8 +1,13 @@
...
+1. Print-based debugging for loops and recursion
+   1. This feature makes tracing easier by printing the state of each loop iteration or recursive function call in a nicely-formatted table using the [tabulate] library.
+2. Memory debugging
+   1. This feature makes it easier to visualize the Python memory model by leveraging the [MemoryViz](https://github.com/david-yz-liu/memory-viz) library.
```
leowrites, line 8: Merge these into one point for each future

docs/debug/index.md
```diff
@@ -250,3 +255,40 @@ The `RecursionTable` is a new PythonTA feature and currently has the following k
...
+
+The following section will focus on tracing the Python memory model. This feature uses the `python_ta.debug.SnapshotManager` as a context manager.
+
+### Example usage
+
+```python
+# demo.py
+from python_ta.debug import SnapshotManager
+
+def func_multi_line(output_path=None) -> None:
```
leowrites, line 269: should rename output_path to output_directory to be more precise

python_ta/debug/snapshot_manager.py
```diff
@@ -0,0 +1,94 @@
...
+import sys
+import types
+from typing import Any, Iterable, Optional
+
+from python_ta.debug.snapshot import snapshot
+
+
+class SnapshotManager:
+    """
+    Class used to manage the snapshots taken during the execution of a program.
```
leowrites, line 16: Should mention this context manager takes a snapshot on every line of function it's a part of

python_ta/debug/snapshot_manager.py
```diff
@@ -0,0 +1,94 @@
...
+            memory_viz_args: The arguments to pass to the memory visualizer.
+            memory_viz_version: The version of the memory visualizer to use.
+            include: A collection of function names, either as strings or regular expressions, whose variables will be captured.
+            output_directory: The path to save the snapshots, defaulting to the current directory.
+                **Note**: Use this argument instead of the `--output` flag in `memory_viz_args` to specify the output directory.
+        """
+        if sys.version_info < (3, 10, 0):
+            logging.warning("You need Python 3.10 or later to use SnapshotManager.")
+        if memory_viz_args is None:
+            memory_viz_args = ["--roughjs-config", "seed=12345"]
```
leowrites, line 50: Probably shouldn't modify arguments

python_ta/debug/snapshot_manager.py
```diff
@@ -0,0 +1,94 @@
...
+        self.memory_viz_version = memory_viz_version
+        self.memory_viz_args = memory_viz_args
+        self._snapshot_counts = 0
+        self.include = include
+
+        if any("--output" in arg for arg in memory_viz_args):
+            raise ValueError(
+                "Use the output_directory argument to specify a different output path."
```
leowrites, line 58: "output_directory parameter"

python_ta/debug/snapshot_manager.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+
+        self.output_directory = output_directory if output_directory else "."
+
+    def _trace_func(self, frame: types.FrameType, event: str, _arg: Any) -> None:
+        """Trace function to take snapshots at each line of code."""
```
leowrites, line 64: Rephrase this to "Take snapshot of the variables in the function stack"

tests/test_debug/test_snapshot_manager.py
```diff
@@ -0,0 +1,151 @@
...
+from python_ta.debug import SnapshotManager
+
+SNAPSHOT_DIR = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)), "snapshot_manager_testing_snapshots"
+)
+TEST_RESULTS_DIR = "/tmp/test_results"
+
+
+def func_one_line(output_path=None) -> None:
+    with SnapshotManager(output_directory=output_path, include=("func_one_line",)):
```
leowrites, line 19: replace `output_path` with `output_directory` for clarity

python_ta/debug/snapshot_manager.py
```diff
@@ -0,0 +1,95 @@
...
+import sys
+import types
+from typing import Any, Iterable, Optional
+
+from python_ta.debug.snapshot import snapshot
+
+
+class SnapshotManager:
+    """
+    A class used for snapshot-based debugging to analyze memory usage at each line in the calling function.
```
david-yz-liu, line 16: "analyze memory usage" has a different connotation about efficiency, like the phrase "memory consumption". I would change this phrase to "visualize program memory".

python_ta/debug/snapshot_manager.py
```diff
@@ -0,0 +1,95 @@
...
+import os
+import re
+import sys
+import types
+from typing import Any, Iterable, Optional
+
+from python_ta.debug.snapshot import snapshot
+
+
+class SnapshotManager:
```
david-yz-liu, line 14: I'm not a big fan of the name "manager" here (naming indeed being a hard problem). Let's go with `SnapshotTracer`. Note that you should rename various files in addition to places in the code.

python_ta/debug/snapshot_manager.py
```diff
@@ -0,0 +1,95 @@
...
+
+import inspect
+import logging
+import os
+import re
+import sys
+import types
+from typing import Any, Iterable, Optional
+
+from python_ta.debug.snapshot import snapshot
```
david-yz-liu, line 11: Use a relative import here, `from .snapshot import snapshot`.

python_ta/debug/snapshot_manager.py
```diff
@@ -0,0 +1,95 @@
...
+        self.memory_viz_version = memory_viz_version
+
+        if memory_viz_args:
+            if any("--output" in arg for arg in memory_viz_args):
+                raise ValueError(
+                    "Use the output_directory parameter to specify a different output path."
+                )
+            self.memory_viz_args = memory_viz_args
+        else:
+            self.memory_viz_args = ["--roughjs-config", "seed=12345"]
```
david-yz-liu, line 58: This isn't a great default value. You can just use an empty list.

python_ta/debug/snapshot_manager.py
```diff
@@ -0,0 +1,95 @@
...
+
+    memory_viz_args: Optional[list[str]]
+    memory_viz_version: Optional[str]
+    _snapshot_counts: int
+    include: Optional[Iterable[str | re.Pattern]]
+    output_directory: Optional[str]
+
+    def __init__(
+        self,
+        memory_viz_args: Optional[list[str]] = None,
```
david-yz-liu, line 33: In terms of the API design, please modify use the following:

```
def __init__(
    self,
    output_directory: Optional[str] = None,
    **kwargs
)
```

The documentation should specify that all keyword arguments are passed to the `snapshot` function, and direct users to view the documentation of that function. That will make this class a bit more robust to changes in the `snapshot` API, since we won't need to change it in two places.

Similarly I don't think you need separate instance attributes for each `snapshot` argument, just one `dict` attribute.

Please do keep the warning about the the `--output` flag though, that's good.

tests/test_debug/test_snapshot_manager.py
```diff
@@ -0,0 +1,151 @@
...
+    snapshot.snapshot_dir = SNAPSHOT_DIR
+    file_name = "snapshot-0.svg"
+    if os.path.exists(file_name):
+        os.remove(file_name)
+    yield
+    os.remove(file_name)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="requires Python 3.10 or higher")
+def test_func_with_non_output_flags(snapshot, setup_curr_dir_testing):
```
david-yz-liu, line 104: Add docstrings to all tests describing exactly what is being tested. Don't rely on the test function names themselves, as they are relatively short.

tests/test_debug/test_snapshot_manager.py
```diff
@@ -0,0 +1,151 @@
...
+
+from python_ta.debug import SnapshotManager
+
+SNAPSHOT_DIR = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)), "snapshot_manager_testing_snapshots"
+)
+TEST_RESULTS_DIR = "/tmp/test_results"
+
+
+def func_one_line(output_directory: str = None) -> None:
```
david-yz-liu, line 18: This file is pretty lengthy with multiple types of definitions (functions that are the "inputs" to the test cases, fixtures, helpers, the actual test cases). You can improve the organization by doing the following:

1. Organize the functions so that they are organized into sections (no mixing of types) and give section a comment as a section header.
2. Write the test cases at the top, then the function inputs, then the helpers, then the fixtures.

tests/test_debug/test_snapshot_manager.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def func_while(output_directory: str = None) -> None:
+    with SnapshotManager(output_directory=output_directory, include=("func_while",)):
+        num = 0
+        while num < 3:
+            num += 1
+
+
+def func_with_args():
```
david-yz-liu, line 60: This is missing a return type. Check this throughout.

tests/test_debug/test_snapshot_manager.py
```diff
@@ -0,0 +1,151 @@
...
+
+import pytest
+from pytest_snapshot.plugin import Snapshot
+
+from python_ta.debug import SnapshotManager
+
+SNAPSHOT_DIR = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)), "snapshot_manager_testing_snapshots"
+)
+TEST_RESULTS_DIR = "/tmp/test_results"
```
david-yz-liu, line 15: pytest provides a fixture for creating a temporary directory, use that instead of defining your own here

tests/test_debug/test_snapshot_manager.py
```diff
@@ -0,0 +1,151 @@
...
+
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="requires Python 3.10 or higher")
+def test_func_with_non_output_flags(snapshot, setup_curr_dir_testing):
+    func_with_args()
+
+    with open("snapshot-0.svg") as actual_file:
+        snapshot.assert_match_dir({"snapshot-0.svg": actual_file.read()}, func_with_args.__name__)
+
+
+functions_to_test = [
```
david-yz-liu, line 111: Don't provide a separate definition, put this inside the `parameterize` call. (This file has a lot of indirect definitions already, so we don't need to add more.)

Also, you don't need to generate `func.__name__` in this list, just pass in the function itself.

tests/test_debug/test_snapshot_manager.py
```diff
@@ -0,0 +1,151 @@
...
+    func_with_names,
+)
+def test_snapshot_manger_with_functions(test_func, snapshot, test_directory):
+    test_func(test_directory)
+
+    assert_output_files_match(test_directory, snapshot, test_func.__name__)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="requires Python 3.10 or higher")
+def test_outputs_to_default_directory_with_no_memory_viz_args(snapshot, setup_curr_dir_testing):
```
david-yz-liu, line 134: Similar to my above comment, the function name alone muddies what is actually being tested. The only thing that this function should test is what happens when the `output_directory` argument is not provided.

tests/test_debug/test_snapshot_manager.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def func_with_args():
+    with SnapshotManager(
+        include=("func_with_args",), memory_viz_args=["--roughjs-config", "seed=12345"]
+    ):
+        flag = "not --output"
+
+
+def assert_output_files_match(output_path: str, snapshot: Snapshot, function_name: str):
```
david-yz-liu, line 67: More generally, add docstrings for all of the functions you defined.

david-yz-liu (commented):
@leowrites good work. In this first review, I focused on the API of `SnapshotManager` as well as some high-level comments to improve the test file. Make sure to also update the documentation accordingly, which I'll review more closely after you've addressed these comments.

tests/test_debug/test_snapshot_manager.py
leowrites: I defined a directory here because the output directory is captured in the snapshots, so we want the directories to be consistent across each test run. 

I looked into using [`tmp_path` and `tmp_path_factory`](https://docs.pytest.org/en/stable/how-to/tmp_path.html#how-to-use-temporary-directories-and-files-in-tests) but I can't seem to get the same base output directory for each run (an example directory I got is `/private/var/folders/rt/x1pynv8j1fb1hkbhs6y9v62m0000gn/T/pytest-of-leoliu/pytest-0`). The only way I found to ensure the base directory is the same is to set the `--basetemp`. Should I rely on this flag to ensure the directories a re consistent?

david-yz-liu: @leowrites that's a great question. This by itself suggests a limitation with the tests, because (of course) you mean to only test the part inside the `with` block, and not this `output_directory` parameter.

I think you should be able to change your function structure by nesting the "function input" definition within the test case itself, making the `output_directory` variable non-local to the function. Something like:

```python
def test_func_one_line(tmp_dir)
    output_directory = ... # tmp_dir's path
    
    def func_one_line():
        with SnapshotTracer(...):
            num = 123

    func_one_line()

    assert_output_files_match(..., ...)
```

This loses the nice use of `pytest.parameterize` that you have, but I think overall the code complexity stays almost the same because you move the function input definitions into the test case code.

leowrites: @david-yz-liu By nesting the functions, the `output_directory` is still captured in `f_locals` inside `func_one_line`. I think this is because `f_locals` also includes free variables in the parent scope when they are referenced. One solution I found is using a global variable so that the variable won't show up in `f_locals`, but it feels unconventional and could break if we were to run the tests concurrently

david-yz-liu: @leowrites wow, yeah okay. I'm not sure how I feel about free variables appearing the stack frame, but I get why it's happening.

All that said, it's not good to use an absolute path (can't assume specific permissions) and it's almost certainly going to look different on Windows.

Why don't we take a pause on this and introduce a different approach. One `snapshot` feature that's on our roadmap is to allow the user to _exclude_ certain variables (again, as an iterable string/regex). It's a different kind of filtering, I suppose, now at the individual frame level. If you implement that, you'll be able to use it here to exclude `output_directory` from the memory model diagram

leowrites: @david-yz-liu This approach worked and I was able to remove the hard coded test directory! 

One thing I noticed while testing is when there are no variables in any frames, `snapshot` calls the memory-viz CLI with an empty JSON `[]`. The memory-viz CLI then errors `Error: Reduce of empty array with no initial value` and the program crashes. 

I added [two checks](https://github.com/pyta-uoft/pyta/pull/1086/files#diff-c6c9e24243fc4ffd8248c423e1c47db81d2340a6d8733bc9cf627054b6169bcdR87-R95) in `snapshot` so that we don't invoke the CLI when the JSON is empty. What do you think about this?

david-yz-liu: @leowrites glad to hear this approach worked.

What you're describing is a bug (or perhaps multiple bugs) with MemoryViz itself. I don't want you to add checks in the `snapshot` code here (it should be fine if MemoryViz receives no objects/stack frames). If you have any test cases that are causing this error, you can mark them pending for now until you fix the MemoryViz bug (and I make a new release).

david-yz-liu (commented):
@leowrites okay great work. I only have one comment left, which is to pin the `memory_viz_version` argument as well (to make the tests robust against future MemoryViz library updates).

david-yz-liu (approved):

PR #1098, opened 2024-10-18
Implement checking for redundant parallel assignment and assignment for multiple targets

Raine-Yang-UofT:
Implement checking for redundant parallel assignment statements and assignment to multiple targets in `redundant_assignment_checker`

Include the name of redundant variable in `redundant-assignment` message.

## Type of Change

| 🐛 _Bug fix_ (non-breaking change that fixes an issue)                                  |    X      |

## Checklist

## Questions and Comments


tests/test_custom_checkers/test_redundant_assignment_checker.py
```diff
@@ -244,3 +244,35 @@ def test_augassign_redundant(self):
...
+        assign_node, *_ = mod.nodes_of_class(nodes.Assign)
+
+        self.checker.visit_module(mod)
+        with self.assertAddsMessages(
+            pylint.testutils.MessageTest(msg_id="redundant-assignment", node=assign_node),
+            ignore_position=True,
+        ):
+            self.checker.visit_assign(assign_node)
+
+    def test_parallel_assign_one_variableredundant(self):
```
david-yz-liu, line 264: add an underscore between "variable" and "redundant"

tests/test_custom_checkers/test_redundant_assignment_checker.py
```diff
@@ -244,3 +244,35 @@ def test_augassign_redundant(self):
...
+        x, y = 0, 0
+        y = 10
+        """
+        mod = astroid.parse(src)
+        mod.accept(CFGVisitor())
+        assign_node, *_ = mod.nodes_of_class(nodes.Assign)
+
+        self.checker.visit_module(mod)
+        with self.assertAddsMessages(
+            pylint.testutils.MessageTest(msg_id="redundant-assignment", node=assign_node),
```
david-yz-liu, line 275: More importantly, however, this test reveals an issue with handling parallel assignment in this way. The error message (calling the assignment statement redundant) is misleading, since only the assignment to `y` is redundant, but not the assignment to `x`. This is a bit more challenging to fix, it might be a special case where only the `y` is highlighted (i.e., passing in the `AssignName` as the `node` attribute to the message, but only in the special case of parallel assignment where not every target is redundant.

Raine-Yang-UofT: Hi Professor, I'd like to know how you expect the one reassigned variable in parallel assignment to be represented differently. Currently the rendering highlights the whole assignment statement, which could be misleading in this case. We could adjust the rendering to highlight just the` y` variable, but this might look inconsistent compared to how we handle other cases where the whole assignment is highlighted.

Another approach would be to leave the rendering as it is, but update the error message to specify which variables are redundant. For example, we could say, "The assignment of variable `y` is redundant". Please let me know which approach you would prefer, or if you have any other suggestions.

david-yz-liu: @Raine-Yang-UofT Yes, these are great points. I'm fine with changing the error message to specify the variable(s) whose assignment is redundant, but please make sure this change applies to regular assignment statements as well. That is, you can make this modification to _all_ uses of this message, not just for parallel assignments.

python_ta/checkers/redundant_assignment_checker.py
```diff
@@ -34,33 +34,47 @@ class RedundantAssignmentChecker(BaseChecker):
...
+            " You can remove it from the program.",
             "redundant-assignment",
             "This assignment statement is redundant;" " You can remove it from the program.",
         )
     }
 
     def __init__(self, linter=None) -> None:
         super().__init__(linter=linter)
-        self._redundant_assignment: set[nodes.Assign] = set()
+        self._redundant_assignment: dict[str, nodes.Assign] = {}
```
david-yz-liu, line 46: As you noted, your presentation last week, the data structure here is awkward because there are multiple times where you are looking up a node from the values of the dictionary.

A better option for this attribute is something like `dict[nodes.Assign, list[str]]`, where for each assignment node the `list[str]` is a list of the target names that are redundant (and therefore displayed in the error message).

Also, I suspect the correct key type is `nodes.Assign | nodes.AugAssign | node.AnnAssign` (not sure if they share a specific common superclass).

python_ta/checkers/redundant_assignment_checker.py
```diff
@@ -34,33 +34,47 @@ class RedundantAssignmentChecker(BaseChecker):
     name = "redundant_assignment"
     msgs = {
         "E9959": (
-            "This assignment statement is redundant;" " You can remove it from the program.",
+            "This assignment statement for variable(s) `%s` is redundant;"
```
david-yz-liu, line 37: Reword to: "Assigning to variable(s) %s here is redundant, because they are not used before getting reassigned later. You can remove the assignment(s) without changing the behaviour of this code."

Please also update the "help" message text on Line 40 below.

python_ta/checkers/redundant_assignment_checker.py
```diff
@@ -34,33 +34,47 @@ class RedundantAssignmentChecker(BaseChecker):
...
+                    variable_list.extend(
+                        n.name for n in target.elts if n.name in self._redundant_assignment.keys()
+                    )
+                elif (
+                    isinstance(target, nodes.AssignName)
+                    and target.name in self._redundant_assignment.keys()
+                ):
+                    variable_list.append(target.name)
+
+            self.add_message("redundant-assignment", node=node, args=",".join(variable_list))
```
david-yz-liu, line 65: use `", " so that there's a space after the commas.

david-yz-liu (approved):

PR #1156, opened 2025-02-23
Integrate Watchdog for Live Code Re-Checking

CulmoneY:
## Proposed Changes

_(Describe your changes here. Also describe the motivation for your changes: what problem do they solve, or how do they improve the application or codebase? If this pull request fixes an open issue, [use a keyword to link this pull request to the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).)_
- Integrated **Watchdog** to enable **automatic re-checking** of Python files when changes are detected.  
- Refactored `_check` into **helper functions**.  
- Ensured that **watch mode only applies to plain text and color text reporters**, leaving HTMLReporter for a future update.  
...

<details>

</details>

## Type of Change

| ♻️ _Refactoring_ (internal change to codebase, without changing functionality)          |.   X  |

## Checklist

## Questions and Comments


python_ta/__init__.py
```diff
@@ -531,3 +581,61 @@ def doc(msg_id: str) -> None:
...
+
+def _watch_files(
+    file_paths: set,
+    level: str,
+    local_config: Union[dict[str, Any], str],
+    load_default_config: bool,
+    autoformat: Optional[bool],
+    linter: PyLinter,
+    current_reporter: BaseReporter | MultiReporter,
+):
```
david-yz-liu, line 594: include the return type

python_ta/__init__.py
```diff
@@ -531,3 +581,61 @@ def doc(msg_id: str) -> None:
...
+    """Watch a list of files for modifications and trigger a callback when changes occur."""
+
+    class FileChangeHandler(FileSystemEventHandler):
```
david-yz-liu, line 597: Don't make this an "internal" class. It can be defined separately from the `_watch_files` function.

python_ta/__init__.py
```diff
@@ -531,3 +581,61 @@ def doc(msg_id: str) -> None:
...
+        """Internal class to handle file modifications."""
+
+        def __init__(self, files_to_watch):
```
david-yz-liu, line 600: include type annotations for all methods (this and the one below)

python_ta/__init__.py
```diff
@@ -127,6 +129,114 @@ def check_all(
...
+    is_any_file_checked: bool,
+    current_reporter: BaseReporter | MultiReporter,
+    level: str,
+    f_paths: list,
+) -> tuple[bool, BaseReporter | MultiReporter, PyLinter]:
+    """Check the file that called this function."""
+    logging.basicConfig(format="[%(levelname)s] %(message)s", level=logging.INFO)
+    allowed_pylint = linter.config.allow_pylint_comments
+    if not _verify_pre_check(file_py, allowed_pylint):
+        return is_any_file_checked, current_reporter, linter
```
david-yz-liu, line 172: Move this block out of the function. Its presence makes the rest of the code more complex.

python_ta/__init__.py
```diff
@@ -127,6 +129,114 @@ def check_all(
...
+    if not PYLINT_PATCHED:
+        patch_all(
+            messages_config, linter.config.z3
+        )  # Monkeypatch pylint (override certain methods)
+        PYLINT_PATCHED = True
+    return linter, current_reporter
+
+
+def _check_file(
+    linter: PyLinter,
```
david-yz-liu, line 158: Overall it is unnecessary to pass the current `linter` (just pass in the specific values you need); similarly, it is unnecessary to return a `current_reporter` (instead just return the linter; you can access the current reporter from the `reporter` attribute of the `linter`). This will also help simplify the internal class you have down below.

python_ta/__init__.py
```diff
@@ -531,3 +581,61 @@ def doc(msg_id: str) -> None:
...
+        """Internal class to handle file modifications."""
+
+        def __init__(self, files_to_watch):
+            self.files_to_watch = set(files_to_watch)
+            self.linter = linter
+            self.current_reporter = current_reporter
+
+        def on_modified(self, event):
+            """Trigger the callback when a watched file is modified."""
+            if event.src_path in self.files_to_watch:
```
david-yz-liu, line 607: This is a good place for a guard: `if event.src_path not in self.files_to_watch` so you can do an early return and then avoid indenting the rest of the code

python_ta/__init__.py
```diff
@@ -150,101 +260,41 @@ def _check(
...
-                # Load config file in user location. Construct new linter each
-                # time, so config options don't bleed to unintended files.
-                # Reuse the same reporter each time to accumulate the results across different files.
-                linter = reset_linter(
-                    config=local_config,
-                    file_linted=file_py,
-                    load_default_config=load_default_config,
+                linted_files.add(file_py)
+                is_any_file_checked, current_reporter, linter = _check_file(
+                    linter,
```
david-yz-liu, line 273: In general, for function calls with many arguments, provide parameter names (e.g. `linter=linter`) to make it clear what each argument refers to.

tests/fixtures/sample_dir/watch/watch_enabled_configuration.py
```diff
@@ -0,0 +1,18 @@
...
+    """blank"""
+    count: int = "ten"
+    return count
+
+
+if __name__ == "__main__":
+    python_ta.check_all(config={
+        "output-format": "python_ta.reporters.PlainReporter",
+        "watch": True,
+    })
```
david-yz-liu, line 18: Configure your IDE to ensure every file ends with a blank line

python_ta/__init__.py
```diff
@@ -531,3 +581,61 @@ def doc(msg_id: str) -> None:
...
+
+        def __init__(self, files_to_watch):
+            self.files_to_watch = set(files_to_watch)
+            self.linter = linter
+            self.current_reporter = current_reporter
+
+        def on_modified(self, event):
+            """Trigger the callback when a watched file is modified."""
+            if event.src_path in self.files_to_watch:
+                print(f"File modified: {event.src_path}, re-running checks...")
```
david-yz-liu, line 608: Use the logging module

python_ta/__init__.py
```diff
@@ -531,3 +581,61 @@ def doc(msg_id: str) -> None:
     msg_url = HELP_URL + "#" + msg_id.lower()
     print("Opening {} in a browser.".format(msg_url))
     webbrowser.open(msg_url)
+
+
+def _watch_files(
```
david-yz-liu, line 586: This function should print (via logging) a message saying "PythonTA is monitoring your files for changes and will re-check after every save"

david-yz-liu (commented):
@CulmoneY nice work!

python_ta/__init__.py
CulmoneY: My apologies, this was an issue with the reset_watch_fixture() function. I will modify this function to include a blank line.

david-yz-liu (commented):
Hi @CulmoneY, very nice work! I left a few small inline comments, but also here are two higher-level comments:

1. I think you may have missed this Slack message about testing (and test coverage): "For this case I recommend adding some separate tests that mock the watchdog functions to assert that they are called correctly. I also encourage you to do some research into how watchdog recommends testing (or how others recommend testing with watchdog)"
2. There's a slight bug right now when a user passes `output=<filename>` to `python_ta.check_all`. I don't think this should be hard to fix, but do make sure to test this part carefully. The desired behaviour is to *append* (not replace) reports to the file. This would mimic redirecting the stdout to a file (for the plain/color/json reporters).

CulmoneY:
> Hi @CulmoneY, very nice work! I left a few small inline comments, but also here are two higher-level comments:
> 
> 
> 
> 1. I think you may have missed this Slack message about testing (and test coverage): "For this case I recommend adding some separate tests that mock the watchdog functions to assert that they are called correctly. I also encourage you to do some research into how watchdog recommends testing (or how others recommend testing with watchdog)"
> 
> 2. There's a slight bug right now when a user passes `output=<filename>` to `python_ta.check_all`. I don't think this should be hard to fix, but do make sure to test this part carefully. The desired behaviour is to *append* (not replace) reports to the file. This would mimic redirecting the stdout to a file (for the plain/color/json reporters).

Hey David, no problem! I'll start fixing these issues.

tests/test_init_logging.py
```diff
@@ -27,7 +27,7 @@ def test_check_log(caplog) -> None:
         assert expected_messages[i] in caplog.records[i].msg
 
 
-@patch("python_ta._get_valid_files_to_check", side_effect=Exception("Testing"))
+@patch("python_ta.get_valid_files_to_check", side_effect=Exception("Testing"))
```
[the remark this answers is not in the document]
CulmoneY: Hey David, As we discussed on monday. I will leave this comment unresolved.

CulmoneY:
Hey David,

Apologies for the number of commits. I was dealing with a flaky test that consistently passed locally but kept failing on GitHub [clearly]. I’ve resolved the issue, but it required adding an extra logging.debug statement to the plain_reporter for better visibility during debugging.

Thanks for your patience!

python_ta/check/watch.py
```diff
@@ -0,0 +1,104 @@
...
+            f_paths=[],
+        )
+        self.current_reporter = self.linter.reporter
+        self.current_reporter.print_messages(self.level)
+        self.linter.generate_reports()
+        upload_linter_results(self.linter, self.current_reporter, self.f_paths, self.local_config)
+
+    def on_close(self) -> None:
+        """Closes the current reporter's output stream"""
+        self.current_reporter.should_close_out = True
```
david-yz-liu, line 65: Avoid setting this attribute here. It should be set when the reporter is first created.

tests/test_check.py
```diff
@@ -217,6 +224,172 @@ def test_check_no_reporter_output(prevent_webbrowser_and_httpserver) -> None:
...
+
+    reset_watch_fixture(str(output_file))
+    process = subprocess.Popen(
+        [sys.executable, "-u", script_path],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    try:
+        wait_for_log_message(
```
david-yz-liu, line 278: Waiting for changes from another process can be tricky. Please do the following:

1. First, in the base reporter's `on_close` method, call `self.out.flush()` to ensure that the full report is flushed to the reporter's output stream. (I think this is safe for all of the reporters, including the HTML reporter, but you should check this.)
2. In this test case, the logic should be:
    - start the subprocess
    - wait for the first log message
    - modify the file
    - wait until the contents of the file contains two instances of "PyTA Report for: ...". Do so in a (generous) timeout, and if the timeout is reached then the test should fail.
    - do the normal cleanup (in the finally block; you don't need an extra `os.kill`)

CulmoneY: Thanks for the guidance. Flushing seems to have resolved the issue I was running into. I’ll go ahead and update the testing logic now.

python_ta/check/watch.py
CulmoneY: Hey David,

I had a small concern about this change. The on_close() method is called each time a report is generated. Since we don’t want the file IO to close while the watch property is still active, I’ve been keeping should_close_output set to False whenever watch=True. Then, once the watch session ends, I close the IO during the cleanup.

The issue is that there’s no reliable way to tell if the current report is the last one, thus I’m unsure how to set the should_close_output attribute correctly during report generation. Would you be able to guide me on a better way to handle this?

tests/test_check.py
CulmoneY: Hey David,

I’ve implemented and tested the methodology you described. The only issue I ran into was that sometimes the call to modify_watch_fixture() wasn’t detected. It looks like the log output was being printed just before the observer fully started, causing a slight timing mismatch.

I’ve seen this happen before and handled it the same way — I now rerun modify_watch_fixture() in a loop until it’s detected, or until a 10-second timeout is reached, at which point the test fails.

python_ta/check/watch.py
david-yz-liu: I see, thanks for clarifying your intent. That approach is okay, but this functionality shouldn't be a method of this class; instead, you can move these two lines directly into the except block down below.

Looking at this more closely, I also note that you do not need to store `current_reporter` as an instance attribute or pass it in as an argument here or in `watch_files`, as the `linter` object holds a reference to the reporter to use.

python_ta/reporters/plain_reporter.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import logging
```
david-yz-liu, line 3: Delete this import (it is no longer being used)

python_ta/reporters/core.py
```diff
@@ -271,6 +271,7 @@ def on_close(self, stats, previous_stats):
 
         Close the reporter's output stream if should_close_out is True.
         """
+        self.out.flush()
```
david-yz-liu, line 274: Okay, you did what I asked you to do (thanks!) but I realized that a better place for this flushing is at the end of `print_messages`/`display_messages` for each reporter, so that the full report is flushed every time the check is run. Because each reporter type has different behaviour, I think it'll be easiest to just add this line to each reporter individually rather than in this parent class.

david-yz-liu (commented):
Thanks @CulmoneY! I just left two more comments. I'm hopeful this will be the last round of review 😄

david-yz-liu (approved):

PR #1162, opened 2025-03-15
Relax Representation Invariant Checking Logic for Recursive Classes

ckatsube:
## Motivation & Proposal

The Representation Invariant checker timing is currently too strict. In many data structures, it's common for instances to mutate other instances of the same type (i.e. graphs/trees or recursive data structures). The current logic enforces RI checks in two places: 1) when an attribute is set EXCEPT for instance methods 2) after the instance method is called. The problem is that Path 1 is executed inside instance methods that set the attributes of different instances of the same type (motivating example below).

The proposed change is to additionally exempt instances of the same type as `self` during the Path 1 check and push checking these RIs until after the instance method has returned, alongside the Path 2 check.

<details>
<summary>More practical motivating example with RecursiveList</summary>

Note: this specific example can technically be avoided by having a better constructor for RecursiveList but the bug still stands regardless.

```python
from typing import Any
from python_ta.contracts import check_contracts


@check_contracts
class RecursiveList:
    """
    Representation Invariants:
    - (self._value is None and self._rest is None) or \
        (self._value is not None and self._rest is not None)
    """
    _value: Any | None
    _rest: RecursiveList | None

    def __init__(self, items: list) -> None:
        if items == []:
            self._value = None
            self._rest = None
        else:
            self._value = items[0]
            self._rest = RecursiveList(items[1:])

    def insert(self, value: Any):
        """Insert value at the current position (push current value to next position)"""
        new_rest = RecursiveList([])
        # new_rest._value and ._rest are originally None by this construction logic.
        new_rest._value = self._value  # RI violation occurs here despite RI being satisfied in the next line.
        new_rest._rest = self._rest
        self._value = value
        self._rest = new_rest


if __name__ == '__main__':
    node_a = RecursiveList([10, 20, 30])
    node_a.insert(5)

```

</details>

## Changes

* Modified the Path 1 check to also exempt RI checking if the type of the parent frame's `self` matches the type of `self`
* Accumulated all of the instances that were exempt in the Path 1 check to later check at Path 2
* Modified the Path 2 logic to also check RIs of the accumulated instance attributes

* Added unit tests to `test_class_contracts` for both the attribute mutation and function arg mutation cases.

## Type of Change

| 🚨 _Breaking change_ (fix or feature that would cause existing functionality to change) |     |
| 🐛 _Bug fix_ (non-breaking change that fixes an issue)                                  |        X |

## Checklist

- [X] I have performed a self-review of my changes.
- [X] I have added tests for my changes, if applicable.
- [x] ~~I have updated the project documentation, if applicable.~~
- [X] I have updated the project Changelog (this is required for all changes).
- [x] ~~If this is my first contribution, I have added myself to the list of contributors.~~

## Questions and Comments

I made a minor logical nuance where subclasses should be able to temporarily violate parent class instances' Representation Invariants in its instance methods (`Child.modify(parent)`). If you think that Child instances should not be allowed to temporarily violate Parent RIs and this check should be strictly equal class types (only Child==Child and Parent == Parent but no Child modifies Parent), let me know! It would effectively just be changing `not isinstance(caller_self, type(self))` to `not type(caller_self) == type(self)`

### DIFF
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -209,9 +209,23 @@ def add_class_invariants(klass: type) -> None:
             original_attr_value_exists = True
             original_attr_value = super(klass, self).__getattribute__(name)
         super(klass, self).__setattr__(name, value)
+
+        if isinstance(self, type(value)):
+            # Add all attributes whose type matches self to __sibling__attributes__ for later
+            # checking contract violations
+            if "__sibling_attributes__" not in self.__dict__:
+                setattr(self, "__sibling_attributes__", {})
+            self.__dict__["__sibling_attributes__"][name] = value
+
         frame_locals = inspect.currentframe().f_back.f_locals
-        if self is not frame_locals.get("self"):
+        if not isinstance(self, type(frame_locals.get("self"))):
             # Only validating if the attribute is not being set in a instance/class method
+            # AND self is an instance of frame_self's type -- aka frame_self is equal to
+            # self, frame_self is another instance of the same class as self,
+            # or frame_self is an instance of self's parent class.
+            # We only additionally check the RI violations of direct instance attributes and
+            # function arguments hence may miss checking RI violations on 2+ chained attribute
+            # mutations or global variable mutations (see :func:`_get_same_typed_instances`).
             if klass_mod is not None:
                 try:
                     _check_invariants(self, klass, klass_mod.__dict__)
@@ -421,12 +435,37 @@ def _instance_method_wrapper(wrapped: Callable, klass: type) -> Callable:
             klass_mod = _get_module(klass)
             if klass_mod is not None and ENABLE_CONTRACT_CHECKING:
                 _check_invariants(instance, klass, klass_mod.__dict__)
+
+                # Also check if other instances in a reasonable scope (immediate instance
+                # attributes and function arguments) had RIs violated after function return.
+                # This does not check mutations in globals--as the set may potentially be large
+                # and likely has mostly unaffiliated values--or the mutation of the attributes of
+                # attributes/args (arg1.attribute.attribute) which may be infinitely recursive.
+                mutable_instances = _get_same_typed_instances(instance, args, kwargs)
+                for mutable_instance in mutable_instances:
+                    instances_klass = type(mutable_instance)
+                    instance_klass_module_dict = _get_module(instances_klass).__dict__
+                    _check_invariants(mutable_instance, instances_klass, instance_klass_module_dict)
         except PyTAContractError as e:
             raise AssertionError(str(e)) from None
         else:
             return r
 
     return wrapper(wrapped)
+
+
+def _get_same_typed_instances(instance: Any, args: list, kwargs: dict) -> list:
+    """Return the instances in instance's attributes, args, and kwargs, whose type
+    matches the type of instance hence may have been mutated during a method call.
+    """
+    attributes = set(instance.__dict__.get('__sibling_attributes__', {}).values())
+    for arg in args:
+        if isinstance(instance, type(arg)) and arg not in attributes:
+            attributes.add(arg)
+    for _name, arg in kwargs.items():
+        if isinstance(instance, type(arg)) and arg not in attributes:
+            attributes.add(arg)
+    return list(attributes)
 
 
 def _instance_init_in_callstack(instance: Any) -> bool:
--- a/python_ta/contracts/__init__.py
+++ b/python_ta/contracts/__init__.py
@@ -213,9 +213,7 @@ def add_class_invariants(klass: type) -> None:
         caller_self = frame_locals.get("self")
         if not isinstance(caller_self, type(self)):
             # Only validating if the attribute is not being set in a instance/class method
-            # AND self is an instance of caller_self's type -- aka caller_self is equal to
-            # self, caller_self is another instance of the same class as self,
-            # or self is an instance of caller_self's parent class.
+            # AND caller_self is an instance of self's type
             if klass_mod is not None:
                 try:
                     _check_invariants(self, klass, klass_mod.__dict__)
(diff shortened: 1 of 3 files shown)


ckatsube:
Converted to draft because the failing test `test_call_super_creates_temp_invalid` where the subclass Teacher's RI is throwing an error after calling a subclass method. Need to figure out why subclass RIs are being evaluated after a parent method is ran (when it shouldn't be).

ckatsube:
Not a current collaborator so can't directly request a reviewer.

@david-yz-liu 
I'd like to request a review.

PR size: small

python_ta/contracts/__init__.py
```diff
@@ -221,6 +225,12 @@ def new_setattr(self: klass, name: str, value: Any) -> None:
                     else:
                         super(klass, self).__delattr__(name)
                     raise AssertionError(str(e)) from None
+        else:
+            caller_klass = type(caller_self)
+            if "__mutated_instances__" in caller_klass.__dict__:
```
david-yz-liu, line 230: Use `hasattr` and `getattr` instead of accessing `__dict__`

python_ta/contracts/__init__.py
```diff
@@ -221,6 +225,12 @@ def new_setattr(self: klass, name: str, value: Any) -> None:
...
+                mutable_instances = caller_klass.__dict__["__mutated_instances__"]
```
david-yz-liu, line 231: The variable should probably be called `mutated_instances`

python_ta/contracts/__init__.py
```diff
@@ -221,6 +225,12 @@ def new_setattr(self: klass, name: str, value: Any) -> None:
                     else:
                         super(klass, self).__delattr__(name)
                     raise AssertionError(str(e)) from None
+        else:
```
david-yz-liu, line 228: I think this `else` is a bit too broad: if `caller_self is self` it doesn't need to be added to `__mutated_instances__`, as the RIs should be checked on `self` at the end of the calling method.

Also, it's worth adding a comment to explain what's going on (parallel to the comment in the if branch).

python_ta/contracts/__init__.py
```diff
@@ -210,8 +210,12 @@ def new_setattr(self: klass, name: str, value: Any) -> None:
             original_attr_value = super(klass, self).__getattribute__(name)
         super(klass, self).__setattr__(name, value)
         frame_locals = inspect.currentframe().f_back.f_locals
-        if self is not frame_locals.get("self"):
+        caller_self = frame_locals.get("self")
+        if not isinstance(caller_self, type(self)):
             # Only validating if the attribute is not being set in a instance/class method
+            # AND self is an instance of caller_self's type -- aka caller_self is equal to
```
david-yz-liu, line 216: The check for `isinstance(caller_self, type(self))` is good (it's fine for an instance of a child class to violate an RI on another instance of the a parent class).

In the comment, the wording should be reversed (caller_self is an instance of self's type); the part including and after the `--` is not necessary and can be removed.

python_ta/contracts/__init__.py
```diff
@@ -421,10 +440,18 @@ def wrapper(wrapped, instance, args, kwargs):
...
+                # across the function call.
+                mutated_instances = getattr(instance_klass, "__mutated_instances__", [])
+                for mutable_instance in mutated_instances:
+                    _check_invariants(mutable_instance, klass, klass_mod.__dict__)
         except PyTAContractError as e:
             raise AssertionError(str(e)) from None
         else:
             return r
+        finally:
+            setattr(instance_klass, "__mutated_instances__", mutated_instances_to_restore)
```
david-yz-liu, line 454: The restoring is generally good; however, we should have different cases for whether `__mutated_instances__` was an attribute or not originally. If it wasn't, we should use `delattr` to remove the instance attribute altogether, so it remains a temporary attribute.

tests/test_contracts/test_class_contracts.py
```diff
@@ -147,6 +168,38 @@ def test_change_age_invalid_in_method(person) -> None:
...
+    Expects normal behavior.
+    """
+    age = person.decrease_and_increase_others_age(person_2, 10)
+    assert age == 10
+
+
+def test_violate_ri_in_other_instance(person, person_2) -> None:
+    """
+    Call a method that changes age of another instance of the same class to something invalid
+    Excepts the RI to be violated hence an AssertionError to be thrown.
```
david-yz-liu, line 184: "Excepts" -> "Expects"

"thrown" -> "raised"

Also, please add a period to the end of the sentence on the previous line.

david-yz-liu (commented):
Thank you, @ckatsube! I left some comments. Also it would be good to add a test case to illustrate the interaction with inheritance.

python_ta/contracts/__init__.py
ckatsube: Sounds good! Thanks for the keen eye

python_ta/contracts/__init__.py
ckatsube: Sounds good!

python_ta/contracts/__init__.py
ckatsube: Thanks for the consistency check 👍

python_ta/contracts/__init__.py
ckatsube: Thanks for following up on the question!

python_ta/contracts/__init__.py
ckatsube: Sounds like a plan--added a branch for deleting the temporary attribute

tests/test_contracts/test_class_contracts.py
ckatsube: Done!

ckatsube:
Resolved the review comments and added the requested tests.

Additionally modified the logic around invariant checking the mutated instances to use the klass/klass_mod of the mutated instance to prevent child class RIs from being enforced on mutated parent class instances (this was caught by the new tests).

ckatsube:
Since the checks passed, this is ready for re-review!

david-yz-liu (approved):
Thanks, @ckatsube!

PR #1179, opened 2025-05-22
Addition of optional argument `on_verify_fail` to `check_all` and `check_error`

Karl-Michaud:
## Proposed Changes

As suggested by the PR title, I added the optional  argument `on_verify_fail` to `check_all` and `check_error` functions (and by extension the helper `_check`) in  `python_ta/check/__init__.py`.

The optional argument `on_verify_fail` can have:
- Default value "log" which keeps the default behaviour when file can't be checked (behaviour before this PR)
- "raise" which immediately raises a `ValueError` and stops execution when file can't be checked
- Else, raise `ValueError` which stops execution since `on_verify_fail` has invalid value

Moreover, since the task didn't specify what type of error should be raised, I decided that the easiest course of action would be to simply to raise a `ValueError`.

```python
if not verify_pre_check(file_py, linter.config.allow_pylint_comments):
    # File cannot be checked
    if on_verify_fail == "raise":
        # Flag to raise error has been chosen
        raise ValueError(f"File cannot be checked: {file_py}")
    # else statement not necessary, but reinforces that "log" keeps default behaviour
    elif on_verify_fail == "log":
        continue
    # Invalid flag
    else:
        raise ValueError(f"on_verify_fail cannot take value: {on_verify_fail}")
```

## Type of Change

| ✨ _New feature_ (non-breaking change that adds functionality)                          |     X    |

## Checklist

- [x] ~~I have updated the project documentation, if applicable.~~

## Questions and Comments

I've updated some docstrings. Should I have also updated the HTML documentation?

Also, for the new test case, I was only able to find one file which triggered `verify_pre_check` to return False. It would be great to create more "bad" files in the future to be more thorough with the testing. 

Finally, it seems that `verify_pre_check` didn't return False for files that should have triggered it. Perhaps, It is an issue to be investigated in the future.

Any tips/criticism is highly appreciated and welcome. I'm always looking to improve!

### DIFF
--- a/python_ta/__init__.py
+++ b/python_ta/__init__.py
@@ -168,10 +168,8 @@ def _check(
                 if not verify_pre_check(
                     file_py, linter.config.allow_pylint_comments, on_verify_fail=on_verify_fail
                 ):
-                    # The only Way to reach this is if verify_pre_check returns False, and `on_verify_fail="log"`.
-                    # Hence, this preserves the default behaviour of the function.
+                    # The only way to reach this is if verify_pre_check returns False, and `on_verify_fail="log"`.
                     continue
-
                 is_any_file_checked, linter = check_file(
                     file_py=file_py,
                     local_config=local_config,
--- a/python_ta/check/helpers.py
+++ b/python_ta/check/helpers.py
@@ -417,14 +417,14 @@ def verify_pre_check(
             + "indentation error at line {}.".format(e.lineno)
         )
         if on_verify_fail == "raise":
-            raise e
+            raise
         return False
     except tokenize.TokenError as e:
         logging.error(
             "python_ta could not check your code due to a " + "syntax error in your file."
         )
         if on_verify_fail == "raise":
-            raise e
+            raise
         return False
     except UnicodeDecodeError as e:
         logging.error(
@@ -437,6 +437,6 @@ def verify_pre_check(
                 if "�" in line:
                     logging.error(f"  Line {i + 1}: {line}")
         if on_verify_fail == "raise":
-            raise e
+            raise
         return False
     return True


python_ta/__init__.py
```diff
@@ -63,6 +63,7 @@ def check_errors(
     output: Optional[Union[str, IO]] = None,
     load_default_config: bool = True,
     autoformat: Optional[bool] = False,
+    on_verify_fail: str = "log",
```
david-yz-liu, line 66: A more appropriate type annotation is `Literal`(https://docs.python.org/3/library/typing.html#typing.Literal)

python_ta/__init__.py
```diff
@@ -106,6 +109,10 @@ def check_all(
             If False, the default PythonTA configuration is not used.
         autoformat:
             If True, autoformat all modules using the black formatting tool before analyzing code.
+        on_verify_fail:
+            determines how to handle files that cannot be checked. If set to "log" (default), an error
```
david-yz-liu, line 113: Capitalize "Determines"

tests/test_check.py
```diff
@@ -256,6 +265,50 @@ def test_check_no_reporter_output(
...
 
+@pytest.mark.parametrize("input_file", INPUTS["test_check_error_raise"])
+def test_check_error_raise(input_file: str | list[str]) -> None:
+    """Test that setting on_verify_fail='raise' causes check_all to raise an error for invalid input files."""
+    with pytest.raises(ValueError, match="File cannot be checked"):
+        python_ta.check_all(
+            input_file,
+            config={
+                "output-format": "pyta-plain",
+                "pyta-error-permission": "no",
```
david-yz-liu, line 276: It's not necessary to add `pyta-error-permission`/`pyta-file-permission`, as the default values are `"no"`. I think other tests containing them are a holdover from previous behaviour.

python_ta/__init__.py
```diff
@@ -155,7 +166,17 @@ def _check(
             for file_py in get_file_paths(locations):
                 linted_files.add(file_py)
                 if not verify_pre_check(file_py, linter.config.allow_pylint_comments):
-                    continue
+                    # File cannot be checked
+                    if on_verify_fail == "raise":
+                        # Flag to raise error has been chosen
+                        raise ValueError(f"File cannot be checked: {file_py}")
```
david-yz-liu, line 172: Overall, this approach loses information that `verify_pre_check` already has access to. Look at the implementation of that function: it's raising errors, then catching them to return `False`. You can configure that function to either return `False` or re-raise the error using a _bare raise_.

Karl-Michaud: Thanks for the input. I opted to modify `verify_pre_check` by adding the optional argument. Through that, I simply checked if an error should be raised (or not).

Karl-Michaud:
Added new changes. I opted to modify verify_pre_check by adding the optional argument. Through that, I simply checked if an error should be raised (or not).

This preserves all raised error data produced in `verify_pre_check`. Hence, in `_check`, if an error is raised, the function immediately jumps to the `except` branch and terminates when `on_verify_fail="raise"`. On the other hand, if  `on_verify_fail="log"` (default), this will preserve the default behaviour since False is returned.

Furthermore, since I modified `verify_pre_check`, I added some tests for better coverage in `tests/test_init_logging.py`.

Finally, due to the new use of `Literal["log", "raise"]`, I removed a useless test.

PR #1188, opened 2025-06-09
Refactoring of the codebase to delay the import of z3 module

Karl-Michaud:
## Proposed Changes

> SEE LATEST COMMENT FOR SUMMARY OF NEW CHANGES

As suggested by the PR title, I refactored `python_ta/checkers/condition_logic_checker.py` and `python_ta/cfg/graph.py` to delay importing the `z3` module.

Specifically, here are the changes made:

1) Removed top-level `z3` imports in both files and moved them inside function. Hence, now it is only executed when the z3 option is enabled in `condition_logic_checker.py`. This is not the case in `graph.py` since the custom config is not accessible (to the best of my knowledge).

2) Wrapped type-only `z3` imports in `TYPE_CHECKING` guards to ensure static type checking as instructed. Thank you for the nudge :)

3) Removed the `z3_dependency_available` variable (no longer necessary). Import issues are now directly handled inside the relevant function bodies by returning early if an error is caught.

Finally, I added new tests in `test_edge_feasibility.py` and `test_impossible_condition_checker.py` ensuring proper coverage of the newly added lines. I used the `unittest.mock` patch decorator to simulate ImportError scenarios.

## Type of Change

| ✨ _New feature_ (non-breaking change that adds functionality)                          |     x     |
| 🐛 _Bug fix_ (non-breaking change that fixes an issue)                                  |    x      |
| ♻️ _Refactoring_ (internal change to codebase, without changing functionality)          |     x    |
| 🚦 _Test update_ (change that _only_ adds or modifies tests)                            |     x    |

## Checklist

- [x] ~~I have added tests for my changes, if applicable~~.
- [x] ~~I have updated the project documentation, if applicable~~.
- [x] ~~If this is my first contribution, I have added myself to the [list of contributors](https://github.com/pyta-uoft/pyta/blob/master/README.md#contributors)~~.

## Questions and Comments

1) Just to confirm I understood correctly. When you said “There are two places in the codebase right now that import z3 at the top level”, I assumed you were referring to `condition_logic_checker.py` and `graph.py`. I did find a third file `z3_parser.py` but since it is entirely built around z3, I figured it made sense to leave the top-level import there. 

2) I was having Git problems because the base commit of this PR was no longer the HEAD of the master branch. I tried rebasing and then pushing, but it wouldn't work due mismatched history. I’ve now reset my branch and successfully rebased it onto the latest master. A force push fixed the issue (this only affected my branch).
Was there a better solution?

3) Any tips/criticism is highly appreciated. I'm always looking to improve!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### 🔧 Internal changes
 
+- Dynamically loaded only the reporter specified in the configuration
 - Added test case for `check_all` function ensuring proper behaviour when handling inputs in package notation.
 - Improved `get_valid_files_to_check` function by removing unreachable code.
 - Refactored `test_check.py` to use `pytest.mark.parametrize` annotation, improving test isolation and extracting inputs from test functions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Refactored `render_pep8_errors` to use a dict that maps error codes to error functions instead of repeated conditional statements
 - Added two test cases to `test_accumulation_table.py` to verify that `AccumulationTable` correctly extracts loop variables from nested tuple structures.
 - Refactored `condition_logic_checker.py` and `graph.py` by removing top-level `z3` imports and moving them inside of functions where needed.
-- Wrapped type-only `z3` imports in a `TYPE_CHECKING` guard in `condition_logic_checker.py` and `graph.py` to avoid runtime dependency issues.
+- Wrapped type-only `z3` imports in a `TYPE_CHECKING` guard in `condition_logic_checker.py` and `graph.py`.
 
 ## [2.10.1] - 2025-02-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Added two test cases to `test_accumulation_table.py` to verify that `AccumulationTable` correctly extracts loop variables from nested tuple structures.
 - Refactored `condition_logic_checker.py` and `graph.py` by removing top-level `z3` imports and moving them inside of functions where needed.
 - Wrapped type-only `z3` imports in a `TYPE_CHECKING` guard in `condition_logic_checker.py` and `graph.py`.
+- Added tests to improve coverage in `condition_logic_checker.py` and `graph.py`, verifying behaviour in case of failed `z3` import via import patching
 
 ## [2.10.1] - 2025-02-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Update how error messages are overridden such that section headers are no longer required within the config file
 - Added `presistent_server` which recives the watch property changes through websockets and updates the HTML report
 - Added optional `on_verify_fail` argument to `check_all` and `check_error`, allowing users to raise a `ValueError` and immediately stop execution when a file cannot be checked.
+- Added the optional `format` argument to the `AccumulationTable` class, allowing users to select between csv or table formatted outputs.
 - Added optional `z3_enabled` argument (default False) to `generate_cfg`, allowing users to enable `z3` functionalities and providing extra safeguard to prevent z3 (and z3 related) imports from being executed when z3 is not enabled.
 
 ### 💫 New checkers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,9 +44,11 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Wrapped type-only imports in if `TYPE_CHECKING` guards
 - Refactored `render_pep8_errors` to use a dict that maps error codes to error functions instead of repeated conditional statements
 - Added two test cases to `test_accumulation_table.py` to verify that `AccumulationTable` correctly extracts loop variables from nested tuple structures.
-- Refactored `condition_logic_checker.py` and `graph.py` by removing top-level `z3` imports and moving them inside of functions where needed.
-- Wrapped type-only `z3` imports in a `TYPE_CHECKING` guard in `condition_logic_checker.py` and `graph.py`.
-- Added tests to improve coverage in `condition_logic_checker.py` and `graph.py`, verifying behaviour in case of failed `z3` import via import patching
+- Refactored `condition_logic_checker.py`, `cfg_generator.py` and `graph.py` by removing top-level `z3` (and z3 related) imports and moving them inside of functions where needed.
+- Wrapped type-only `z3` (and z3 related) imports in a `TYPE_CHECKING` guard in `condition_logic_checker.py`, `graph.py` and `cfg_generator.py`.
+- Added tests to improve coverage in `condition_logic_checker.py`, `graph.py` and `cfg_generator.py`, verifying behaviour in case of failed `z3` (and z3 related) import via import patching.
+- Added `z3_enabled` optional initializer argument (default False) to `ControlFlowGraph` class in `graph.py` and `CFGVisitor` class in `visitor.py`.
+- Added `z3_enabled` optional function argument (default False) to `generate_cfg` (and by extension `_generate` helper) in `cfg_generator.py` providing extra safeguard to prevent z3 (and z3 related) imports from being executed when z3 is not enabled.
 
 ## [2.10.1] - 2025-02-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Update how error messages are overridden such that section headers are no longer required within the config file
 - Added `presistent_server` which recives the watch property changes through websockets and updates the HTML report
 - Added optional `on_verify_fail` argument to `check_all` and `check_error`, allowing users to raise a `ValueError` and immediately stop execution when a file cannot be checked.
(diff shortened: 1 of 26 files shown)


python_ta/cfg/graph.py
```diff
@@ -72,11 +52,15 @@ def add_arguments(self, args: Arguments) -> None:
         args.parent.cfg = self
         args.parent.cfg_block = self.start
 
-        if ExprRef is not Any:
+        try:
+            from ..z3.z3_parser import Z3Parser
```
david-yz-liu: Overall you've done what I asked you to do (great!), but looking at this again, I'd actually like to you go a step further. Please introduce a new initializer argument and instance attribute called `z3_enabled` (default `False`), which, if `True`, enables the z3 functionality in this class. Concretely, you can keep the try-except code, but wrap it in an if statement:

```python
if self.z3_enabled:
   ...
```

This will even more aggressively ensure that Z3 isn't imported if it isn't going to be used. It also allows you to just import `z3` once in the initializer; you can set it to a _global variable_ so that it becomes globally accessible, so that you don't have to repeat the import in each method.

```python
# outside the class
z3 = None

# in the class
class ControlFlowGraph:
    def __init__(...):
        global z3
        if z3_enabled:
            try:
                import z3 as imported_z3
                z3 = imported_z3
            except ImportError: ...
```

Note that `CFGGraph` is instantiated a few places (like `CFGVisitor`), so you might need to add additional arguments/attributes to that class as well. This might also impact the test suite.

david-yz-liu: A second comment: in general, try blocks should contain the minimal code that causes the expected error. So in this case, you should _only_ include the import statement in the try-except. The subsequent code below (`parser = Z3Parser()`, etc.) can do in an `else` block (these can be attached to try-except blocks, if you haven't seen that before).

python_ta/cfg/graph.py
```diff
@@ -72,11 +52,15 @@ def add_arguments(self, args: Arguments) -> None:
...
+
             # Parse types
             parser = Z3Parser()
             z3_vars = parser.parse_arguments(args)
             self.z3_vars.update(z3_vars)
+        except ImportError:
+            return
```
david-yz-liu: The return is good; however, please also use `logging.error` to display an error message to tell the user that even though they tried to enable Z3, the module couldn't be imported.

python_ta/cfg/graph.py
```diff
@@ -271,7 +255,10 @@ def update_edge_z3_constraints(self) -> None:
 
         Constraints with reassigned variables are not included in subsequent edges.
         """
-        if not z3_dependency_available:
+        try:
+            from z3 import Not
+
```
david-yz-liu: delete this line

python_ta/cfg/graph.py
```diff
@@ -310,7 +297,10 @@ def update_edge_feasibility(self) -> None:
         attribute of each edge. Edges that are unreachable with the given
         set of Z3 constraints will have is_feasible set to False
         """
-        if not z3_dependency_available:
+        try:
+            from z3 import Solver, unsat
+
```
david-yz-liu: delete this line

python_ta/cfg/graph.py
```diff
@@ -472,6 +462,14 @@ def parse_constraint(self, node: NodeNG) -> Optional[ExprRef]:
         """Parse an Astroid node to a Z3 constraint
         Return the resulting expression
         """
+        try:
```
david-yz-liu: Unlike the `Graph` class, here the `Z3Environment` class is so connected to Z3 that I don't think we need a try-except. Instead just include the imports, and if they fail it's fine for the error to be raised.

david-yz-liu (commented):
@Karl-Michaud good work; I left some comments on this PR.

You indeed modified the files I was hoping you would, so that's great. You don't need to worry about `z3_parser.py` for exactly the reason you mentioned.

Regarding your git question: it is normal to have your feature branch get out of date with the master branch. In this case, you should use a `git pull` to update your feature branch; rebasing is a more advanced technique, so I recommend avoiding it until you're comfortable with using `git pull` to do a merge. This is covered in the _Using Git Effectively_ page.

After implementing these changes, it would be great if you added a cProfile analysis of the timing of running `import python_ta` and `python_ta.check_all` on `master` vs. this branch. I asked @CulmoneY to do the same thing on his PRs, so you can talk to him about doing this. Hopefully we'll see a speedup in both of those commands!

CHANGELOG.md
```diff
@@ -44,12 +44,18 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+- Added tests to improve coverage in `condition_logic_checker.py`, `graph.py` and `cfg_generator.py`, verifying behaviour in case of failed `z3` (and z3 related) import via import patching.
+- Added `z3_enabled` optional initializer argument (default False) to `ControlFlowGraph` class in `graph.py` and `CFGVisitor` class in `visitor.py`.
+- Added `z3_enabled` optional function argument (default False) to `generate_cfg` (and by extension `_generate` helper) in `cfg_generator.py` providing extra safeguard to prevent z3 (and z3 related) imports from being executed when z3 is not enabled.
 
 ## [2.10.1] - 2025-02-19
 
 ### 🐛 Bug fixes
 
 - Fix import error when `graphviz` is not installed
+- Fixed CI test failure in `test_examples.py` caused by `patch_ast_transforms` function that captured the `z3` option value from the first linter instance (default False on CI). Now, it dynamically reads `z3` option from linter config to reflect correct runtime setting.
```
david-yz-liu: This entry is in the wrong section (should be unreleased, above).

Also, you can reword this entry to just focus on the functionality change, it's not just about CI test failures. The key part is something like "Ensure the patched version of [the pylint function] dynamically reads `z3` option from linter config to reflect correct runtime setting."

python_ta/cfg/graph.py
```diff
@@ -57,22 +41,41 @@ class ControlFlowGraph:
...
-    def __init__(self, cfg_id: int = 0) -> None:
+    def __init__(self, cfg_id: int = 0, z3_enabled: bool = False) -> None:
         self.block_count = 0
         self.cfg_id = cfg_id
         self.unreachable_blocks = set()
         self.start = self.create_block()
         self.end = self.create_block()
         self.z3_vars = {}
         self.precondition_constraints = []
+        self.z3_enabled = z3_enabled
```
david-yz-liu: You can wait to set this after the try-except. Something like

```python
if z3_enabled:
    try:
       ...

    except ImportError:
       ...
    else:
       self.z3_enabled = True
else:
    self.z3_enabled = False
```

python_ta/cfg/graph.py
```diff
@@ -271,9 +274,8 @@ def update_edge_z3_constraints(self) -> None:
 
         Constraints with reassigned variables are not included in subsequent edges.
         """
-        if not z3_dependency_available:
+        if not self.z3_enabled:
             return
-
```
david-yz-liu, line 276: revert this change

python_ta/cfg/graph.py
```diff
@@ -472,23 +474,28 @@ def parse_constraint(self, node: NodeNG) -> Optional[ExprRef]:
...
+
+        from ..z3.z3_parser import Z3ParseException, Z3Parser
+
         parser = Z3Parser(self.variables)
         try:
             return parser.parse(node)
         except (Z3Exception, Z3ParseException):
             return None
 
+    def _get_vars(self, expr: ExprRef) -> Set[str]:
```
david-yz-liu: I don't think this needs to be an instance method. Just adding the `z3` import should be good

python_ta/checkers/condition_logic_checker.py
```diff
@@ -101,12 +100,15 @@ def _check_condition(self, node: Union[nodes.If, nodes.While]) -> None:
...
             self.add_message("impossible-condition", node=node.test)
 
-    def _check_unsat(self, prev_constraints: z3.ExprRef, node_constraint: z3.ExprRef) -> bool:
+    def _check_unsat(self, prev_constraints: ExprRef, node_constraint: ExprRef) -> bool:
         """Check if the conjunction of the given constraints is unsatisfiable.
 
         - prev_constraints (z3.ExprRef): Constraints from previous nodes.
         - node_constraint (z3.ExprRef): The condition to check at the current node.
         """
+        # z3 is already imported by caller (cached), no need to check for importError again
```
david-yz-liu: capitalize "Import"

tests/test_cfg/test_cfg_generator_z3.py
```diff
@@ -0,0 +1,30 @@
...
+    # Create the graphviz files using my_file.py
+    cfg_generator.generate_cfg(mod=file_path, z3_enabled=True)
+
+    # Open the actual graphviz file for reading
+    gv_file_io = open(gv_file_path)
+
+    assert gv_file_path in gv_file_io.read()
+
+    # Teardown: close any open file IO and remove the graphviz generated files
+    gv_file_io.close()
```
david-yz-liu: It's important to ensure these three statements are executed even if the `assert` fails.

For the `close`, you can resolve this by using a `with` block enclosing the assertion.

For the `os.remove` calls, you can wrap the assert in a try block, and use a finally clause to ensure the `os.remove` calls get executed.

tests/test_custom_checkers/test_impossible_condition_checker.py
```diff
@@ -1,3 +1,5 @@
+from unittest.mock import patch
```
david-yz-liu: This import is unused and can be removed (along with the blank line below it)

tests/test_cfg/test_cfg_generator_z3.py
```diff
@@ -0,0 +1,30 @@
+"""
+Test suite for generating z3 control flow graphs using the PythonTA API.
+"""
+
+import os.path
```
david-yz-liu, line 5: This import and the `pytest`  import are unused

Karl-Michaud: Apologies. It was remnants of the test that had to be removed  due to CI errors.

david-yz-liu (approved):
Nice work, @Karl-Michaud!

PR #1194, opened 2025-06-26
Failing GitHub CI coverage tests [FOR @david-yz-liu]

Karl-Michaud:
## Failing GitHub CI test

I wrote two versions of the same failing test. These tests provide coverage for the import error code in the `_generate` helper in `cfg_generator.py`.

Here are my two best attempts:
1) attempt 1: Use patching to "mock" the error
2) attempt 2: Use monkeypatching to manually replace the module with a "dummy" module

I also tried:
- Using patch to replace the module with a "dummy" module (instead of None)
- Using monkeypatching to manually delete the module

### DIFF
--- a/tests/test_cfg/test_cfg_generator_z3.py
+++ b/tests/test_cfg/test_cfg_generator_z3.py
@@ -3,6 +3,9 @@ Test suite for generating z3 control flow graphs using the PythonTA API.
 """
 
 import os.path
+from unittest.mock import patch
+
+import pytest
 
 import python_ta.cfg.cfg_generator as cfg_generator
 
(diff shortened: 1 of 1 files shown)


CulmoneY:
Hey Karl, how are you running your tests locally? `pytest tests tests/test_cfg/test_cfg_generator_z3.py` produces the same error as CI for myself.

Karl-Michaud:
Hey @CulmoneY, thanks for taking a look at it.

I ran the test using `python3 -m pytest tests/test_cfg/test_cfg_generator_z3.py` and it worked fine. 

Screenshot 2025-06-27 at 1 43 36 PM

Furthermore, although `mod` is an empty string, `abs_path = _get_valid_file_path(mod)` would still return something on my computer (and no error would be produced).  Does it have to do with different python versions you think?

Since the test fails both on GitHub CI and locally on your computer, at least we are sure that my computer is the problem here.

I'll take a deeper look. I appreciate the debugging you did; it really helps to have a fresh set of eyes.

Also, I can definitely move the test to be inside `tests/test_cfg/test_cfg_generator.py`. I figured it made more sense in `tests/test_cfg/test_cfg_generator_z3.py`, since we set z3_enabled=True there. It doesn't affect anything though, since the z3 functionality isn't used either way due to the ImportError.

david-yz-liu (commented):
@Karl-Michaud thanks for sharing these. Calling `generate_cfg` without a `mod` argument means that the `'__main__'` module gets a graph generated. When pytest is running the tests, pytest itself is the main module. There could be something funny happening with how pytest is installed on CircleCI (or @CulmoneY's machine) that's causing this error.

In any case, please update the tests to pass a `mod` argument (you can just use the same one as the `test-cfg_z3_enabled`).

Karl-Michaud:
@david-yz-liu, I've added the changes. It works!

From what I understood, this was just a Pytest related error and shouldn’t affect the user experience.

david-yz-liu (approved):

PR #1200, opened 2025-07-15
Disabled Garbage Collection During Snapshot Tracing

CulmoneY:
## Proposed Changes

Currently, there is an issue with garbage collection throughout the lifecycle of the snapshot tracer. Currently, we use an objects id as a unqiue identifier. However, when an item gets garbage collected, this id can be reused. A simple and quite elegant solution is to simply disable garbage collection.

...

<details>

</details>

## Type of Change

| ♻️ _Refactoring_ (internal change to codebase, without changing functionality)          |        x |

## Checklist

## Questions and Comments


python_ta/reporters/TODO.md
```diff
@@ -20,7 +18,6 @@
 ## `HTMLReporter`
 
 - Fix the following `_COLOURING` dict entries, which need to be included so that HTMLReporter can call `_colour_messages_by_type` safely (currently useless):
-
```
CulmoneY: Weird. These changes were auto generated by the precommit hook. Maybe they updated it? This file has not been touched since 2021. 🤔

david-yz-liu (commented):
Hi @CulmoneY, you did good research here, but the Python garbage collector apparently doesn't track primitive objects. This means this is not a good solution to the issue, as I also do not want primitive objects to share ids across program execution steps. For example, consider the following code:

```python
from python_ta.debug import SnapshotTracer


def f(x):
    acc = 0
    with SnapshotTracer(webstepper=True):
        for i in range(x):
            acc += i
    return acc


if __name__ == '__main__':
    f(4)
```

The individual integers should all have different ids in the generated memory model diagrams.

tests/test_debug/snapshot_tracer_testing_snapshots/func_for_loop/snapshot-2.svg
CulmoneY, line 1: This change is a bug. I am working on this fix!

python_ta/debug/snapshot.py
```diff
@@ -56,6 +58,7 @@ def get_filtered_local_variables(
 
 
 def snapshot(
+    id_tracker: IDTracker,
```
david-yz-liu: Generally speaking, adding parameters to existing functions should be done quite carefully to preserve backwards compatibility.

In this case, (1) `id_tracker` should be added to the end of the list, and (2) it should be optional (default value `None`), as the only time it's needed is when the user wants to track ids across multiple calls to `snapshot`. This is only really necessary for the `SnapshotTracer` class, but not other uses of `snapshot`. Same applies to `snapshot_to_json`.

Also make sure to add documentation for this attribute.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### 🐛 Bug fixes
 
+- Disabled garbage collection in `SnapshotTracer` to prevent it from interfering with the memory model generation.
```
david-yz-liu: This entry needs to be updated

python_ta/debug/id_tracker.py
```diff
@@ -0,0 +1,35 @@
+from typing import Any, Optional
+
+
+class IDTracker:
```
david-yz-liu, line 4: Add documentation for this class and the initializer. In particular, explain how this class is intended to support both long-term and short-term id tracking, as this is the most important and subtle part of this class.

Add attribute type annotations in the body of the class, not the initializer. This looks like:

```python
class IDTracker:
    """..."""

    id_counter: int
    # etc.
```

python_ta/debug/id_tracker.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def get_id(self, obj) -> Optional[int]:
+        """Retrieve the unique ID for an object, or none if not tracked."""
+        return self.tracker.get(id(obj), None)
+
+    def is_tracked(self, obj) -> bool:
+        """Check if the object has already been tracked"""
+        return id(obj) in self.tracker
+
+    def add_snapshot_object(self, obj) -> None:
```
david-yz-liu: It isn't necessary to make this a separate function. Combine its behaviour with `set_id` above.

python_ta/debug/id_tracker.py
```diff
@@ -0,0 +1,35 @@
+from typing import Any, Optional
+
+
+class IDTracker:
+    def __init__(self) -> None:
+        self.id_counter: int = 0
+        self.tracker: dict[int, int] = {}
+        self.current_snapshot_objects: set[int] = set()
+
+    def set_id(self, obj) -> Optional[int]:
```
david-yz-liu: The return type is `None`. Also, make sure to include type annotations for all parameters here and throughout.

python_ta/debug/id_tracker.py
```diff
@@ -0,0 +1,35 @@
...
+        """Assign a unique ID to an object."""
+        if self.is_tracked(obj):
+            return
+        self.id_counter += 1
+        self.tracker[id(obj)] = self.id_counter
+
+    def get_id(self, obj) -> Optional[int]:
+        """Retrieve the unique ID for an object, or none if not tracked."""
+        return self.tracker.get(id(obj), None)
```
david-yz-liu: Just use regular indexing (i.e., with square brackets). It's fine to raise an error here; the calling code expects an `int`, not `Optional[int]`.

python_ta/debug/id_tracker.py
```diff
@@ -0,0 +1,35 @@
...
+        self.current_snapshot_objects: set[int] = set()
+
+    def set_id(self, obj) -> Optional[int]:
+        """Assign a unique ID to an object."""
+        if self.is_tracked(obj):
+            return
+        self.id_counter += 1
+        self.tracker[id(obj)] = self.id_counter
+
+    def get_id(self, obj) -> Optional[int]:
```
david-yz-liu: Overall, this class is a thin wrapper around a `dict`. So, we can make this more explicit by using `__getitem__` an `__setitem__` for the two main methods; then in the calling code, we can use regular indexing syntax (`id_tracker[obj]`, `id_tracker[obj] = ...`) to make this a bit more explicit

python_ta/debug/id_tracker.py
```diff
@@ -0,0 +1,35 @@
...
+        """Retrieve the unique ID for an object, or none if not tracked."""
+        return self.tracker.get(id(obj), None)
+
+    def is_tracked(self, obj) -> bool:
```
david-yz-liu: Similarly, use `__contains__` here so that you can use the syntax `obj in id_tracker`

python_ta/debug/snapshot.py
```diff
@@ -159,66 +160,64 @@ def process_value(val: Any) -> int:
...
-                    json.dumps(val)
-                    jsonable_val = val
-                except TypeError:
-                    jsonable_val = repr(val)
-                value_entry = {
-                    "type": type(val).__name__,
-                    "id": value_id_diagram,
-                    "value": jsonable_val,
-                }
-
```
david-yz-liu, line 218: Revert this change

david-yz-liu (commented):
@CulmoneY this approach is good; please also add tests for the `IDTracker` class.

python_ta/debug/id_tracker.py
CulmoneY: The only issue I see is that `set_id` does not behave like i`d_tracker[obj] = ...` Instead, it always assigns a new internally derived value based on a counter.  I will do this refactor for `get_id` however./

david-yz-liu: It's up to us (as designers of this class) to determine how we want `id_tracker[obj] = ...` to behave. In this case, what I'm saying is that I'm good with conveying to the user that what they should be doing is writing `id_tracker[obj] = ...` in order to store new objects. What happens internally can be kept private.

python_ta/debug/id_tracker.py
```diff
@@ -0,0 +1,46 @@
...
+        """Initialize internal counters and containers for tracking object IDs."""
+        self.id_counter = 0
+        self.tracker = {}
+        self.current_snapshot_objects = set()
+
+    def __getitem__(self, obj: Any) -> int:
+        """Retrieve the unique ID for an object, or none if not tracked."""
+        return self.tracker[id(obj)]
+
+    def __setitem__(self, obj: Any, value: Any) -> None:
```
david-yz-liu: Ah okay I understand better what you meant now (the `value` argument isn't used in this function, which is a significant difference from how `__setitem__` is normally used).

We can go back to the previous approach you had, but change the name to `add` (which mirrors the `set.add` method name). You can also modify this method so that it returns the `self.id_counter` value so that it can be used directly in `process_value` (without needing to then call `__getitem__`).

python_ta/debug/id_tracker.py
```diff
@@ -0,0 +1,46 @@
+from typing import Any
+
+
+class IDTracker:
+    """
+    A helper class to assign and track unique integer IDs for Python objects across multiple memory model snapshots.
```
david-yz-liu: Start the description on the same line as the triple quotes.

tests/test_debug/test_id_tracker.py
```diff
@@ -0,0 +1,57 @@
...
+    assert tracker[obj1] == 1
+    assert tracker[obj2] == 2
+
+
+def test_getitem_keyerror() -> None:
+    """Test that __getitem__ raises KeyError for untracked objects."""
+    tracker = IDTracker()
+    obj = object()
+    with pytest.raises(KeyError):
+        _ = tracker[obj]
```
david-yz-liu: you can just write `tracker[obj]` here, no need for the `_ =` part

PR #1204, opened 2025-07-16
Addition of new InfiniteLoopChecker

Karl-Michaud:
## Proposed Changes

> 🚨 See comment below for summary of new changes

1) Added a new custom checker `InfiniteLoopChecker` that detects while loops whose condition variables are never updated inside the loop body. The checker parses the condition of each while loop to identify variables. If no condition variable is used in the body, then an warning message is added to the reporter.

2) Added tests for the new checker (which demonstrates intended behaviour)

3) Updated the `Custom errors` documentation to include: `Unused Loop Condition Variable (W9501)`. Furthermore, create an example file `w9501_unused_loop_condition_variables.py`

## Type of Change

| ✨ _New feature_ (non-breaking change that adds functionality)                          |     X     |
| 🚦 _Test update_ (change that _only_ adds or modifies tests)                            |      X    |

## Checklist

## Questions and Comments


python_ta/checkers/infinite_loop_checker.py
```diff
@@ -0,0 +1,73 @@
...
+from pylint.checkers import BaseChecker
+from pylint.lint import PyLinter
+
+
+class InfiniteLoopChecker(BaseChecker):
+    name = "infinite-loop"
+    msgs = {
+        "W9501": (
+            "No condition variable in the while loop is used inside the loop body",
+            "loop-condition-variable-unused",
```
david-yz-liu: You can just call this check `infinite-loop`. Even though the implementation is only detecting one case currently, we can extend it to detect others in the future, so we shouldn't limit ourselves to a particular case in the error name/message/description. Please update the message and description as well.

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -0,0 +1,73 @@
...
+        ),
+    }
+
+    def __init__(self, linter: Optional[PyLinter] = None) -> None:
+        super().__init__(linter=linter)
+
+    def visit_while(self, node: nodes.While) -> None:
+        self._check_condition_all_var_used(node)
+
+    def leave_while(self, node: nodes.While) -> None:
```
david-yz-liu: this method doesn't need to be defined

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -0,0 +1,73 @@
...
+    def _check_condition_all_var_used(self, node: nodes.While) -> None:
+        """Helper function that checks whether variables used in a while loop's condition
+        are also used anywhere inside the loop body.
+
+        Note: This is a basic check. It only flags loops where NONE of the condition variables
+        appear in the body, which indicates an infinite loop.
+        """
+        # Get variable(s) used inside condition
+        cond_vars = set()
+        for child in node.test.nodes_of_class((nodes.Name, nodes.Attribute)):
```
david-yz-liu: Considering attributes isn't a bad idea in general, but it is more complex because attributes can be modified through method calls, so it isn't enough to check for whether the attribute is used directly in the loop body.

So, you should remove checks for attributes, and instead just focus on variable names (i.e., `Name` nodes). Moreover, you don't need to exclude `self`; if `self` appears in the loop condition, e.g. through attribute access or a method call, it's valid to also check whether `self` is used in the loop body.

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -0,0 +1,73 @@
...
+                    or isinstance(name_node, nodes.Subscript)
+                    and name_node.value.as_string() in cond_vars
+                ):
+                    # At least one condition variable is used in the loop body
+                    return
+        else:
+            self.add_message(
+                "loop-condition-variable-unused",
+                node=node.test,
+                line=node.test.lineno,
```
david-yz-liu: You don't need to provide `line` and `end_lineno` arguments (by default these will be set to the corresponding attributes for the `node` argument

tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -0,0 +1,177 @@
...
+
+        with self.assertAddsMessages(
+            pylint.testutils.MessageTest(
+                msg_id="loop-condition-variable-unused",
+                node=node.test,
+            ),
+            ignore_position=True,
+        ):
+            self.checker.visit_while(node)
+            self.checker.leave_while(node)
```
david-yz-liu: Similar to my above comment, you don't need to call `leave_while`

tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -0,0 +1,177 @@
...
+
+class TestInfiniteLoopChecker(pylint.testutils.CheckerTestCase):
+    CHECKER_CLASS = InfiniteLoopChecker
+
+    def test_var_not_updated(self) -> None:
+        """Test that the checker correctly flags a while loop when no condition variables are used in the loop body."""
+        src = """
+            i, k, l = 0, 20, 40
+            j = 10
+            while i < 100 and k < 21 or l < 40:
```
david-yz-liu: A technique that some tests use (but not all) is to add the comment `#@` to the end of the source code line where the node you want to extract begins. That would be the line with the `while` keyword in this case.

By doing so, you can then use `astroid.extract_node` to obtain exactly the node you want, without needing to call `nodes_of_class`.

Please update the tests here to use this approach.

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -0,0 +1,53 @@
...
+        cond_vars = set(child.name for child in node.test.nodes_of_class(nodes.Name))
+        if not cond_vars:
+            return
+        # Check to see if condition variable(s) used inside body
+        for child in node.body:
+            for name_node in child.nodes_of_class((nodes.Name, nodes.AssignName)):
+                if name_node.name in cond_vars:
+                    # At least one condition variable is used in the loop body
+                    return
+        else:
```
vibin9, line 39: The `else` block is not needed.

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -0,0 +1,53 @@
...
+            self.add_message(
+                "infinite-loop",
+                node=node.test,
+            )
+            return
```
vibin9: Return statement is not needed .

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -0,0 +1,53 @@
...
+        # Check to see if condition variable(s) used inside body
+        for child in node.body:
+            for name_node in child.nodes_of_class((nodes.Name, nodes.AssignName)):
+                if name_node.name in cond_vars:
+                    # At least one condition variable is used in the loop body
+                    return
+        else:
+            self.add_message(
+                "infinite-loop",
```
vibin9, line 41: Consider replacing with `InfiniteLoopChecker.name`

tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -0,0 +1,173 @@
...
+
+        node, detected_node = astroid.extract_node(src)
+
+        with self.assertAddsMessages(
+            pylint.testutils.MessageTest(
+                msg_id="infinite-loop",
+                node=detected_node.test,
+            ),
+            ignore_position=True,
+        ):
```
vibin9, line 70: Since you are reusing this testing code in a lot of places, you can factor it out into a helper:
```python
from contextlib import contextmanager

@contextmanager
def assert_infinite_loop_message():
    with self.assertAddsMessages(
        pylint.testutils.MessageTest(
            msg_id=InfiniteLoopChecker.name,
            node=detected_node.test,
        ),
        ignore_position=True,
    ):
        yield
```

docs/checkers/index.md
```diff
@@ -2742,6 +2742,23 @@ python_ta.check_all(config={
...
+
+This error is raised when a `while` loop never terminates, indicating an infinite loop.
+
+Example:
+
+```{literalinclude} /../examples/custom_checkers/w9501_infinite_loop.py
+
+```
+
+This checker is enabled by default and requires no special configuration.
```
david-yz-liu: You can remove this sentence (it's the default behaviour for all custom checkers, so we should not say anything to maintain consistency with the rest of the documentation)

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -0,0 +1,53 @@
...
+    msgs = {
+        "W9501": (
+            "Infinite while loop: loop does not terminate",
+            "infinite-loop",
+            """Used when a while loop is guaranteed to never terminate based on its current structure. This usually
+            indicates a logical error leading to an unintended infinite loop.""",
+        ),
+    }
+
+    def __init__(self, linter: Optional[PyLinter] = None) -> None:
```
david-yz-liu: You don't need this initializer

python_ta/checkers/infinite_loop_checker.py
david-yz-liu: You can delete the return statement here

tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -0,0 +1,173 @@
...
+            self.checker.visit_while(node)
+
+    def test_condition_variable_used_only_as_argument(self):
+        """Test that the checker does not flag while loops when the condition variable is passed to a function
+        or method."""
+        src = """
+         i = 0
```
david-yz-liu: The indentation here is off (a bit surprised black didn't change this, but oh well)

david-yz-liu (approved):
Thanks, @Karl-Michaud!

PR #1207, opened 2025-07-23
Exclude function names from condition variable analysis in infinite-loop checker

Karl-Michaud:
## Proposed Changes

1) Excluded function names from the condition variable analysis to avoid false positives like:

```python
while f(10):
    ...
```

Added:

```python
 if isinstance(node.test, nodes.Call):
      return
  # Get variable(s) used inside condition
  cond_vars = set(
      child.name for child in node.test.nodes_of_class(nodes.Name, skip_klass=nodes.Call)
  )
```

This skips any function call (or method call since the same logic would aply here).

2) Added a test to verify expected behaviour.

## Type of Change

| 🐛 _Bug fix_ (non-breaking change that fixes an issue)                                  |     X     |

## Checklist

- [x] ~~I have updated the project documentation, if applicable.
  - This is **required** for new features.~~

## Questions and Comments

### DIFF
--- a/tests/test_custom_checkers/test_infinite_loop_checker.py
+++ b/tests/test_custom_checkers/test_infinite_loop_checker.py
@@ -1,6 +1,5 @@
 import astroid
 import pylint.testutils
-import pytest
 
 from python_ta.checkers.infinite_loop_checker import InfiniteLoopChecker
 
@@ -179,8 +178,8 @@ class TestInfiniteLoopChecker(pylint.testutils.CheckerTestCase):
         src = """
         while foo(10): #@
             x += 1
-        while foo(10) < 21 and faa(13): #@
-            x += 1
+        while foo(10) < 21 and faa(x): #@
+            x += 1 # Note: x is not considered a condition variable since it is used inside a function call
         """
         node1, node2 = astroid.extract_node(src)
 
--- a/tests/test_custom_checkers/test_infinite_loop_checker.py
+++ b/tests/test_custom_checkers/test_infinite_loop_checker.py
@@ -1,5 +1,6 @@
 import astroid
 import pylint.testutils
+import pytest
 
 from python_ta.checkers.infinite_loop_checker import InfiniteLoopChecker
 
@@ -171,3 +172,18 @@ class TestInfiniteLoopChecker(pylint.testutils.CheckerTestCase):
 
         with self.assertNoMessages():
             self.checker.visit_while(node)
+
+    def test_while_func_call(self) -> None:
+        """Test verifies that function calls in while-loop conditions do not incorrectly trigger infinite-loop
+        warnings."""
+        src = """
+        while foo(10): #@
+            x += 1
+        while foo(10) < 21 and faa(13): #@
+            x += 1
+        """
+        node1, node2 = astroid.extract_node(src)
+
+        with self.assertNoMessages():
+            self.checker.visit_while(node1)
+            self.checker.visit_while(node2)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Fixed error in `contracts` where comments in docstring assertions are not removed while parsing
 - Improved error message in `patches/transforms.py` where CFGVisitor is run
 - Fixed a bug in AccumulationTable where loop variable names weren't being captured for all nested targets.
+- Fixed a bug in the `infinite-loop` checker where function and method names were incorrectly included in the set of condition variables.
 
 ### 📚 Documentation Updates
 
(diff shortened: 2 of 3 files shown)


tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -171,3 +171,18 @@ def test_while_true_enabled(self) -> None:
...
             self.checker.visit_while(node)
+
+    def test_while_func_call(self) -> None:
+        """Test verifies that function calls in while-loop conditions do not incorrectly trigger infinite-loop
+        warnings."""
+        src = """
+        while foo(10): #@
+            x += 1
+        while foo(10) < 21 and faa(x): #@
+            x += 1 # Note: x is not considered a condition variable since it is used inside a function call
```
david-yz-liu: This isn't exactly the behaviour I'm looking for. I don't want you to skip all variables that are passed as arguments to function calls; instead, I'd like you to just ignore the function names themselves.

For example, given 

```python
while foo(x):
    y += 1
```

we should still treat `x` as a condition variable and this loop should trigger the `infinite-loop` error.

In terms of implementation, don't ignore all variables inside `Call` nodes; instead, just ignore _function names_. This is a bit harder to do, but remember than given a `Name` node you can access its `parent` attribute and check whether that's a function call, and whether the `Name` is the expression being called (you'll need to access the appropriate `Call` attribute to check this).

Karl-Michaud: I completely misunderstood the task! Sorry for any inconvenience :)

Karl-Michaud:
Hi @david-yz-liu, here are the new changes:

# Changes

I didn't see a clean or reliable way to determine the parent node of each `Name` node. For example:

```python
foo(faa(lst[1][2]["yellow"]))
```
The parent of the `Name` node corresponding to `lst` would be a `Subscript` node, not a `Call` node.

Instead, I opted for a different approach: manually removing function names from the variable set.

```python
  # Get variable(s) used inside condition
  cond_vars = set(child.name for child in node.test.nodes_of_class(nodes.Name))
  # Remove function names
  for call in node.test.nodes_of_class(nodes.Call):
      if isinstance(call.func, nodes.Name) and call.func.name in cond_vars:
          cond_vars.remove(call.func.name)
```

# Analysis: why I believe this approach is still good!

Let n = number of `nodes.Name` in node.test

Let m = number of `nodes.Call` in node.test

1) Constructing the `cond_vars` set involves up to n insertions, so O(n). 
2) The loop over `nodes.Call` takes O(m) time. Each iteration is constant time (due to fast set lookups and removals).

Since there's at least one `Name` node per `Call`, we know m <= n. Therefore, the worst case runtime is still O(n).

However, conditions rarely contain many variables or function calls. The number of `Name` and `Call` nodes in a condition is usually small and bounded, so the amortized complexity is O(1).

In conclusion, this 2 loop approach has negligible impact on the checker's performance.

david-yz-liu:
> I didn't see a clean or reliable way to determine the parent node of each `Name` node. For example:
> 
> ```python
> foo(faa(lst[1][2]["yellow"]))
> ```
> 
> The parent of the `Name` node corresponding to `lst` would be a `Subscript` node, not a `Call` node.

Hi @Karl-Michaud, I think you may be overcomplicating things. Your task is only to exclude `Name` nodes corresponding to *the names of functions being called*. In the example you provided, `lst` is not a function name, so it should be treated as a "condition variable" and so not ignored. So, it's fine that its parent isn't the `Call` node. In fact, checking `isinstance(node.parent, Call)` is a great first step to see whether the name should be filtered out, but even if that is true, you should also make sure to check that the Name` node is the function being called.

Karl-Michaud:
> Your task is only to exclude `Name` nodes corresponding to _the names of functions being called_. In the example you provided, `lst` is not a function name, so it should be treated as a "condition variable" and so not ignored. 

It seems that I wasn't clear in my comment. The code works as expected and `lst` is in fact treated as a condition variable (and not ignored).

I was trying to explain how I didn't really see how checking the parent node would come in useful, when we can just iterate over all Call nodes, and manually remove the function name from the set.

Karl-Michaud:
> Note: It appears that Coveralls is reporting a lower coverage percentage for `python_ta/contracts/__init__.py`, even though there are no differences in the covered lines between this commit and the base commit. The logic and tests remain the same, and no related code was modified (only the code below was changed and a comment was modified in `test_while_fund_call_var`).


Hi @david-yz-liu,

I’ve completed the task as intended (hopefully!).

I want to apologize again for any inconvenience, and I really appreciate your patience and guidance :)

It admittedly took me longer than expected to fully understand what you meant about using parent nodes to distinguish function names... the solution turned out to be much simpler than I initially imagined.

```python
  cond_vars = set()
  for child in node.test.nodes_of_class(nodes.Name):
      if not isinstance(child.parent, nodes.Call) or child.parent.func is not child:
          cond_vars.add(child.name)
```

This gives us the expected behaviour in a single loop. For example:

```python
while faa(all(x)) and lst[1][2]["yellow"].get_address() or func(var, foo(all(z, 10))): 
    y += 1 # Should trigger an infinite loop since condition variables set: {'lst', 'var', 'z', 'x'}
```

david-yz-liu (approved):
Great work!

PR #1212, opened 2025-07-30
Addition of new check to InfiniteLoopChecker

Karl-Michaud:
> See comment below

## Proposed Changes

This PR implements a new `infinite-loop` check to `InfiniteLoopChecker`. As requested, this new check flags `while` loops that satisfy both:
- A constant loop condition (i.e.: always True)
- No exit statement (`return`, `break`, `raise`, `yield`, `sys.exit()`)

To detect constant conditions, I adapted the logic from Pylint’s `basic_checker._check_using_constant_test` method, which cannot be directly reused.

Exit statements are detected by walking the loop body and checking for:
- Direct `return`, `break`, `raise`, or `yield` statements
- Calls to `sys.exit()`, verified via inference to ensure the callee is the `sys` module

## Type of Change

| ✨ _New feature_ (non-breaking change that adds functionality)                          |     X     |

## Checklist


Karl-Michaud:
Hi @david-yz-liu,

I’ve made a lot of progress implementing the new check, but I’ve encountered an issue I wanted to flag before proceeding further.

---

## Problem

The Pylint `_check_using_constant_test` helper does **not** flag constant conditions like:

```python
while 1 < 2:
    # body
```

This is due to the following block of code:

```python
except_nodes = (
    nodes.Call,
    nodes.BinOp,
    nodes.BoolOp,  # BoolOp nodes are skipped
    nodes.UnaryOp,
    nodes.Subscript,
)
inferred = None
maybe_generator_call = None
emit = isinstance(test, (nodes.Const, *structs, *const_nodes))
if not isinstance(test, except_nodes):
    inferred = utils.safe_infer(test)
    # If we can't infer what the value is but the test is just a variable name
    if isinstance(inferred, util.UninferableBase) and isinstance(test, nodes.Name):
        emit, maybe_generator_call = InfiniteLoopChecker._name_holds_generator(test)
```

Here, `1 < 2` is represented as a `BoolOp` node, which is explicitly excluded by the check, causing constant expressions like this to be ignored.

---

##  Temporary Solution (Proof of Concept)

To resolve this, I’ve implemented a temporary workaround that extends the existing logic. Here's the modified version:

```python
except_nodes = (
    nodes.Call,
    nodes.BinOp,
    # nodes.BoolOp,  # BoolOp no longer excluded
    nodes.UnaryOp,
    nodes.Subscript,
)
inferred = None
maybe_generator_call = None
emit = isinstance(test, (nodes.Const, *structs, *const_nodes))
if not isinstance(test, except_nodes):
    inferred = utils.safe_infer(test)
    if isinstance(inferred, util.UninferableBase):
        if isinstance(test, nodes.Name):
            emit, maybe_generator_call = InfiniteLoopChecker._name_holds_generator(test)
    else:
        if (
            inferred is not None
            and isinstance(inferred, nodes.Const)
            and bool(inferred.value) is True
        ):
            cond_vars = set(child for child in node.test.nodes_of_class(nodes.Name))
            if not cond_vars:
                emit = True
```

This patch:
- Infers the result of the boolean condition,
- Verifies it evaluates to `True`,
- Ensures no `Name` nodes (i.e., variables / functions) are used — confirming it's a constant expression.

---

## Notes

- I’ve **temporarily included the patch** in the latest commit so it’s easier for you to review — this is **not** permanent.
- I’ve also added **tests** that (to the best of my knowledge) verify the patch works on the types of cases described in the task Slack message.

---

## Conclusion

> Since you previously brought up the issue of constant `while` loop conditions (e.g., `while 1 < 2`): you can go ahead and add an additional case for the infinite-loop checker which reports the error when both of the following conditions are met: (1) the `while` loop condition is constant; (2) the loop body has no `return`/`break`/`raise`/`yield` statements, nor calls to `sys.exit` (use astroid inference for this last check). To determine whether the `while` loop condition is constant, you should use the same logic as the Pylint using-constant-test checker, specifically this helper method: https://github.com/pylint-dev/pylint/blob/main/pylint/checkers/base/basic_checker.py#L303. Unfortunately, you can't use that method directly (consider: why not?) but you can copy the logic into your own helper method and provide a reference.

As per the spec, a condition like `while 1 < 2` should be recognized as constant.

Would you like me to proceed with integrating this logic permanently in `InfiniteLoopChecker`?

Should we disregard cases where we have constant bool operations?

david-yz-liu:
Hi @Karl-Michaud, you're absolutely correct that the `_check_using_constant_test` method from pylint is ignoring more complex expressions even when all subexpressions are constants (including `1 < 2`).

I'm fine addressing this, but I wouldn't do it in the way that you've suggested, as it is cleaner to separate out the custom code from the pylint-sourced code. Logically, we can keep the pylint code pretty much exactly the same as it was before, and instead add a completely separate check for when (1) there are no `Name` nodes in the while loop `test`, and (2) the `test` is inferred to be `True`. This doesn't even need to be in the same function as `_check_using_constant_test`, it can be in a completely separate helper.

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -45,6 +50,159 @@ def _check_condition_all_var_used(self, node: nodes.While) -> None:
...
+                        not isinstance(inferred, util.UninferableBase)
+                        and inferred is not None
+                        and isinstance(inferred, nodes.Module)
+                        and inferred.name == "sys"
+                    ):
+                        return
+        else:
+            self.add_message("infinite-loop", node=node.test, confidence=INFERENCE)
+
+    def _check_constant_comp_test(self, node: nodes.While) -> bool:
```
david-yz-liu: Hi @Karl-Michaud, since this PR is still marked as a draft, I'm just leaving a quick review here about the new part you added.

Overall I think `_check_constant_comp_test` is good, but here are two comments:

1. I think the name (and function) is unduly narrow. This won't just check for "comparison of constant notes", but really any Python expression that involves only constants (for example, `1 + 2`). And I think that's a good thing!
2. You're right abut the duplication of inference not being ideal. I think your suggestion of just having the inference done in the "main" function is a good one.

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -45,6 +50,162 @@ def _check_condition_all_var_used(self, node: nodes.While) -> None:
...
+        if not self._check_constant_loop_cond(
+            node, node.test
+        ) and not self._check_constant_form_condition(node):
+            return
+        inferred = utils.safe_infer(node.test)
+        if isinstance(inferred, util.UninferableBase) or inferred is None:
+            return
+        if (
+            isinstance(inferred, nodes.Const)
+            and bool(inferred.value) is False
```
david-yz-liu: Remembering the precedence of `and`/`or` is tricky, so I would put parentheses around each `and` clause:

```python
(isinstance(inferred, nodes.Const) and bool(inferred.value) is False)
or (... and ...)
or (... and ...)
```

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -45,6 +50,162 @@ def _check_condition_all_var_used(self, node: nodes.While) -> None:
...
+                        and isinstance(inferred, nodes.Module)
+                        and inferred.name == "sys"
+                    ):
+                        return
+        else:
+            self.add_message("infinite-loop", node=node.test, confidence=INFERENCE)
+
+    def _check_constant_form_condition(self, node: nodes.While) -> bool:
+        """Helper function that checks if while loop condition is of constant form (e.g.: `1 < 2`, `5 - 1 >= 2 + 2`)"""
+        cond_vars = set(child for child in node.test.nodes_of_class(nodes.Name))
```
david-yz-liu: Since you just need to know whether the collection is empty or not, you can directly call `any` on `node.test.nodes_of_class(nodes.Name)`. No need for the comprehension expression (`child for child in ...`); this is redundant.

tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -204,3 +195,347 @@ def test_while_fund_call_var(self) -> None:
...
+        """,
+            """
+        faa = lambda: True
+
+        while faa: #@
+            x += 1
+        """,
+        ],
+    )
+    def test_while_func_obj_condition(self, src: str) -> None:
```
david-yz-liu, line 342: This is good, and in fact a third test you can use is with an inline lambda:

```python
while (lambda: True):
    ...
```

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -17,6 +21,7 @@ class InfiniteLoopChecker(BaseChecker):
     }
 
     def visit_while(self, node: nodes.While) -> None:
+        self._check_condition_constant(node)
```
david-yz-liu: We shouldn't be having the same message emitted more than once (thanks for flagging this in the tests). You can change the `_check` helpers to return a `bool` that indicates whether an error was flagged, and only execute the next check if no error was flagged.

Karl-Michaud: Hi @david-yz-liu,

I modified the test suite/checker code to implement the specified behavior.

Each `_check` helper now returns True if it emits a message, and False otherwise. In visit_while, the logic is:

```python
def visit_while(self, node: nodes.While) -> None:
    checks = [self._check_condition_constant, self._check_condition_all_var_used]
    any(check(node) for check in checks)
```

Conveniently, `any` comes in handy again, since it short-circuits once any check returns True.

david-yz-liu (approved):
@Karl-Michaud nice work!

PR #1231, opened 2025-09-07
Add infinite-loop detection for `while` loops with immutable condition variables

Karl-Michaud:
## Proposed Changes

This PR uses **astroid inference** to determine the data type of variables in a `while` loop condition. If all variables are inferred to be immutable, the loop body is checked for reassignments only (i.e., `AssignName` nodes), allowing the detection of potentially infinite loops, such as `while i < 1: print(i)` when `i` is immutable.  

### Behaviour
- Loops with conditions that are always `False` or cannot be inferred are ignored.  
- If all condition variables are immutable and none are reassigned in the loop body, the loop is flagged as potentially infinite.  
- Any mutable variable or other operations on immutable variables prevent the loop from being flagged.  

### More about

1. **Mixed mutable and immutable variables:** Not flagged, since behaviour is undefined.  
2. **Reassigned immutable variables:** Loops are not flagged if any immutable variable is reassigned, as it could change the loop condition.  
3. **Have to infer loop condition:** If not checked, loops with False conditions may still be flagged.
4. **Inference limitations:** Variables with complex number value (immutable type), may not be inferable by `astroid` :(  .In such cases, the loop is not flagged.  

## Type of Change

| ✨ _New feature_ (non-breaking change that adds functionality)                          |    X     |
| 🚦 _Test update_ (change that _only_ adds or modifies tests)                            |    X     |

## Checklist

- [ ] I have performed a self-review of my changes.
- [ ] I have added tests for my changes, if applicable.
- [ ] I have updated the project documentation, if applicable.
- [ ] I have updated the project Changelog (this is required for all changes).
- [ ] If this is my first contribution, I have added myself to the [list of contributors](https://github.com/pyta-uoft/pyta/blob/master/README.md#contributors).

## Questions and Comments

### DIFF
--- a/python_ta/checkers/infinite_loop_checker.py
+++ b/python_ta/checkers/infinite_loop_checker.py
@@ -201,12 +201,11 @@ class InfiniteLoopChecker(BaseChecker):
         immutable_types = (
             int,
             float,
-            complex,
             bool,
+            complex,
             str,
             bytes,
             tuple,
-            frozenset,
             type(None),
         )
         immutable_vars, cond_vars = set(), set()
@@ -218,13 +217,19 @@ class InfiniteLoopChecker(BaseChecker):
                     continue
                 # Check if type of inferred value is immutable
                 if (
-                    (isinstance(inferred, nodes.Const) and type(inferred.value) in immutable_types)
-                    or isinstance(inferred, nodes.Tuple)
-                    or isinstance(inferred, nodes.Const.FrozenSet)
-                ):
+                    isinstance(inferred, nodes.Const) and type(inferred.value) in immutable_types
+                ) or isinstance(inferred, nodes.Tuple):
                     immutable_vars.add(child.name)
         if not immutable_vars or immutable_vars != cond_vars:
             # There are no vars with immutables values OR there are vars with mutable values
+            return False
+
+        # Infer the loop condition
+        inferred_test = utils.safe_infer(node.test)
+        if isinstance(inferred_test, util.UninferableBase) or inferred_test is None:
+            return False
+        if isinstance(inferred_test, nodes.Const) and inferred_test.value is False:
+            # Condition is always false, loop won't run. No need to check for infinite loop.
             return False
 
         for child in node.body:
--- a/python_ta/checkers/infinite_loop_checker.py
+++ b/python_ta/checkers/infinite_loop_checker.py
@@ -198,6 +198,14 @@ class InfiniteLoopChecker(BaseChecker):
         - All variables in the `while` condition are immutable (int, float, complex, bool,
           str, bytes, tuple, frozenset, or NoneType)
         - None of these variables are reassigned in the loop body"""
+        # Infer the loop condition
+        inferred_test = utils.safe_infer(node.test)
+        if isinstance(inferred_test, util.UninferableBase) or inferred_test is None:
+            return False
+        if isinstance(inferred_test, nodes.Const) and inferred_test.value is False:
+            # Condition is always false, loop won't run. No need to check for infinite loop.
+            return False
+
         immutable_types = (
             int,
             float,
@@ -222,14 +230,6 @@ class InfiniteLoopChecker(BaseChecker):
                     immutable_vars.add(child.name)
         if not immutable_vars or immutable_vars != cond_vars:
             # There are no vars with immutables values OR there are vars with mutable values
-            return False
-
-        # Infer the loop condition
-        inferred_test = utils.safe_infer(node.test)
-        if isinstance(inferred_test, util.UninferableBase) or inferred_test is None:
-            return False
-        if isinstance(inferred_test, nodes.Const) and inferred_test.value is False:
-            # Condition is always false, loop won't run. No need to check for infinite loop.
             return False
 
         for child in node.body:
--- a/python_ta/checkers/infinite_loop_checker.py
+++ b/python_ta/checkers/infinite_loop_checker.py
@@ -198,14 +198,6 @@ class InfiniteLoopChecker(BaseChecker):
         - All variables in the `while` condition are immutable (int, float, complex, bool,
           str, bytes, tuple, frozenset, or NoneType)
         - None of these variables are reassigned in the loop body"""
-        # Infer the loop condition
-        inferred_test = utils.safe_infer(node.test)
-        if isinstance(inferred_test, util.UninferableBase) or inferred_test is None:
-            return False
-        if isinstance(inferred_test, nodes.Const) and inferred_test.value is False:
-            # Condition is always false, loop won't run. No need to check for infinite loop.
-            return False
-
         immutable_types = (
             int,
             float,
@@ -230,6 +222,14 @@ class InfiniteLoopChecker(BaseChecker):
                     immutable_vars.add(child.name)
         if not immutable_vars or immutable_vars != cond_vars:
             # There are no vars with immutables values OR there are vars with mutable values
+            return False
+
+        # Infer the loop condition
+        inferred_test = utils.safe_infer(node.test)
+        if isinstance(inferred_test, util.UninferableBase) or inferred_test is None:
+            return False
+        if isinstance(inferred_test, nodes.Const) and inferred_test.value is False:
+            # Condition is always false, loop won't run. No need to check for infinite loop.
             return False
 
         for child in node.body:
--- a/python_ta/checkers/infinite_loop_checker.py
+++ b/python_ta/checkers/infinite_loop_checker.py
@@ -21,7 +21,11 @@ class InfiniteLoopChecker(BaseChecker):
     }
 
     def visit_while(self, node: nodes.While) -> None:
-        checks = [self._check_condition_constant, self._check_condition_all_var_used]
+        checks = [
+            self._check_condition_constant,
+            self._check_condition_all_var_used,
+            self._check_immutable_cond_var_reassigned,
+        ]
         any(check(node) for check in checks)
 
     def _check_condition_all_var_used(self, node: nodes.While) -> bool:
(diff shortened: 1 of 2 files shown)


python_ta/checkers/infinite_loop_checker.py
```diff
@@ -186,6 +190,61 @@ def _name_holds_generator(test_node: nodes.Name) -> tuple[bool, Optional[nodes.C
...
 
+    def _check_immutable_cond_var_reassigned(self, node: nodes.While) -> bool:
+        """Helper function that checks if a while-loop condition uses only immutable variables
+        and none of them are reassigned inside the loop body.
+
+        Flags loops that meet **both** of the following criteria:
+        - All variables in the `while` condition are immutable (int, float, complex, bool,
+          str, bytes, tuple, or NoneType)
+        - None of these variables are reassigned in the loop body"""
+        immutable_types = (
```
david-yz-liu: Pull this out into a top-level constant.

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -186,6 +190,61 @@ def _name_holds_generator(test_node: nodes.Name) -> tuple[bool, Optional[nodes.C
...
+                    continue
+                # Check if type of inferred value is immutable
+                if (
+                    isinstance(inferred, nodes.Const) and type(inferred.value) in immutable_types
+                ) or isinstance(inferred, nodes.Tuple):
+                    immutable_vars.add(child.name)
+                else:
+                    return False
+        if not immutable_vars:
+            # There are no vars with immutables values
```
david-yz-liu: immutables -> immutable

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -186,6 +190,61 @@ def _name_holds_generator(test_node: nodes.Name) -> tuple[bool, Optional[nodes.C
...
+            complex,
+            str,
+            bytes,
+            tuple,
+            type(None),
+        )
+        immutable_vars = set()
+        for child in node.test.nodes_of_class(nodes.Name):
+            if not isinstance(child.parent, nodes.Call) or child.parent.func is not child:
+                inferred = utils.safe_infer(child)
```
david-yz-liu: `.safe_infer` just returns one value; here, you should use a different astroid method (`.infer`), which returns a generator for all possible values

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -186,6 +190,61 @@ def _name_holds_generator(test_node: nodes.Name) -> tuple[bool, Optional[nodes.C
...
+                ) or isinstance(inferred, nodes.Tuple):
+                    immutable_vars.add(child.name)
+                else:
+                    return False
+        if not immutable_vars:
+            # There are no vars with immutables values
+            return False
+
+        # Infer the loop condition
+        inferred_test = utils.safe_infer(node.test)
```
david-yz-liu: This part is being repeated enough across the checks that it's probably a good idea to pull this out into a helper function

tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -407,6 +389,141 @@ def test_constant_fail(self, src: str) -> None:
         with self.assertNoMessages():
             self.checker._check_condition_constant(node)
 
+    @pytest.mark.parametrize(
+        "src",
```
david-yz-liu, line 416: This is good; please also add a `None` case. You can add a `complex` case as well with just the imaginary part (e.g. `3j`) and see if astroid correctly performs inference on that.

Karl-Michaud: Hi @david-yz-liu, I've found something interesting. Turns out:

# Observations

1) Inference succeeds in the following case:

```python
  i = 1 + 3j
  while i:
      print(i)
```

```python
  i =  3j
  while i:
      print(i)
```

2) Inference fails in the following cases:

```python
  i = 1 + 3j
  while i == 1 + 2j:
      print(i)
```

```python
  i = 3j
  if i == 1:
      print(i)
```


# Conclusion

Inference succeeds when the condition is a directly evaluable constant expression. It fails when the condition involves a comparison or more complex expression.

In other words, the inference engine can resolve truthiness of constants, but not the result of runtime comparisons between constants.

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -1,13 +1,24 @@
...
 import itertools
-from typing import Optional
+from typing import Optional, Union
 
 from astroid import BoundMethod, InferenceError, UnboundMethod, bases, nodes, util
 from pylint.checkers import BaseChecker, utils
 from pylint.interfaces import INFERENCE
 from pylint.lint import PyLinter
 
+immutable_types = (
```
david-yz-liu: use the `ALL_CAPS` naming style for this, since it's a constant. Also, you can make this a set

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -186,6 +200,70 @@ def _name_holds_generator(test_node: nodes.Name) -> tuple[bool, Optional[nodes.C
...
+        Flags loops that meet **both** of the following criteria:
+        - All variables in the `while` condition are immutable (int, float, complex, bool,
+          str, bytes, tuple, or NoneType)
+        - None of these variables are reassigned in the loop body"""
+        immutable_vars = set()
+        for child in node.test.nodes_of_class(nodes.Name):
+            if isinstance(child.parent, nodes.Call) and child.parent.func is child:
+                continue
+            try:
+                # Get all possible values
```
david-yz-liu: This comment isn't necessary

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -186,6 +200,70 @@ def _name_holds_generator(test_node: nodes.Name) -> tuple[bool, Optional[nodes.C
...
+            for inferred in inferred_values:
+                if not _is_immutable_node(inferred):
+                    # Return False when the node may evaluate to a mutable object
+                    return False
+            immutable_vars.add(child.name)
+        if not immutable_vars:
+            # There are no vars with immutable values
+            return False
+
+        # Infer the loop condition
```
david-yz-liu: I think you misunderstood my previous comment. I wasn't referring to the use of `.safe_infer()` here, I was referring to the entire block of logic that infers the value of `node.test` and checks the inferred value. This is repeated across the checks, and should be pulled out into a helper.

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -186,6 +200,70 @@ def _name_holds_generator(test_node: nodes.Name) -> tuple[bool, Optional[nodes.C
...
+                "infinite-loop",
+                node=node.test,
+                confidence=INFERENCE,
+            )
+            return True
+
+
+def _is_immutable_node(node: nodes.NodeNG) -> bool:
+    """Helper used to check whether node represents an immutable type."""
+    if (isinstance(node, nodes.Const) and type(node.value) in immutable_types) or isinstance(
```
david-yz-liu: In general, an if statement with `return True`/`return False` statements is redundant, and can be replaced with a single return statement: `return (isinstance(node, nodes.Const) and ...)`

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -186,6 +200,70 @@ def _name_holds_generator(test_node: nodes.Name) -> tuple[bool, Optional[nodes.C
...
+    else:
+        return False
+
+
+def get_safely_inferred(node: nodes.NodeNG) -> Union[nodes.NodeNG, None]:
+    """Helper used to safely infer a node with `astroid.safe_infer`. Return None if inference failed."""
+    inferred = utils.safe_infer(node)
+    if isinstance(inferred, util.UninferableBase) or inferred is None:
+        return None
+    return inferred
```
david-yz-liu: put this in an else block

tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -407,6 +389,151 @@ def test_constant_fail(self, src: str) -> None:
...
+                print(i)
+            """,
+            """
+            i = 0.1
+            while i < 1: #@
+                print(i)
+            """,
+            """
+            i = 1 + 3j
+            while i:
```
david-yz-liu: This test is missing the `#@`

tests/test_custom_checkers/test_infinite_loop_checker.py
david-yz-liu: That's great, thanks for the additional information 👍

tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -407,6 +389,151 @@ def test_constant_fail(self, src: str) -> None:
...
+                print(i)
+            """,
+            """
+            i = ""
+            while i != "ooooooo": #@
+                i = i + "o"
+            """,
+            """
+            i, j, k, l = 1, 0.9, True, "Goodbye"
+            while i + j > 10 or not k or l == "cool":
```
david-yz-liu: This test is missing the `#@`

tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -407,6 +389,151 @@ def test_constant_fail(self, src: str) -> None:
...
+                print(i, j, k, l)
+            """,
+            """
+            i, j, k, l = 1, 0.9, True, "Goodbye"
+            while i + j < 10 and not k or l != "cool":
```
david-yz-liu: This test is missing the `#@`

tests/test_custom_checkers/test_infinite_loop_checker.py
Karl-Michaud: Apologies for all those misses😅

python_ta/checkers/infinite_loop_checker.py
Karl-Michaud: Understood. I’ve reverted back to using `safe_infer`, this time via the helper.

david-yz-liu: @Karl-Michaud you are still misunderstanding what I'm saying. I referred to "the entire block of logic that infers the value of `node.test` and checks the inferred value". This is the following code:

```python
        inferred_test = get_safely_inferred(node.test)
        if inferred_test is None:
            return False
        if isinstance(inferred_test, nodes.Const) and inferred_test.value is False:
            # Condition is always false, loop won't run. No need to check for infinite loop.
            return False
```

The same basic logic is repeated in lines 80-88---in fact, the logic isn't identical, but it should be; this is an even stronger argument for making a single helper function and calling it in both places. I also see that this logic is _not_ repeated in `_check_condition_all_var_used`, but it probably should be.

Karl-Michaud: Ah, I see...
I should’ve asked for clarification earlier. I’ll make sure to be more mindful of that in the future. Third time’s the charm! 🙂

tests/test_custom_checkers/test_infinite_loop_checker.py
```diff
@@ -178,24 +176,6 @@ def test_while_func_call(self) -> None:
             self.checker.visit_while(node1)
             self.checker.visit_while(node2)
 
-    def test_while_fund_call_var(self) -> None:
```
david-yz-liu: Hmmm, the fact that you deleted this test is a sign something weird happened.

Right now the logic is considering inference failures (returning `Uninferable`, `None`) as preventing the infinite loop error from being flagged. However, I think that might actually be the wrong approach: if the inference fails, I think it's still okay to flag the possible infinite loop. This is because it's more likely than not that the person writing the code would expect the loop condition to be true rather than false.

So, please modify the implementation of `infer_condition` so that it also returns `True` when the inference fails.

python_ta/checkers/infinite_loop_checker.py
```diff
@@ -62,8 +77,8 @@ def _check_condition_constant(self, node: nodes.While) -> bool:
             node.test
         ) and not self._check_constant_form_condition(node):
             return False
-        inferred = utils.safe_infer(node.test)
-        if isinstance(inferred, util.UninferableBase) or inferred is None:
+        inferred = get_safely_inferred(node.test)
```
david-yz-liu: You should be able to replace the code here with the new helper function.

PR #1251, opened 2025-10-22
Optimizing performance for 'test_examples.py'

Karl-Michaud:
## Proposed Changes

> Note: See comments for full investigation

### Summary of changes

- Added checker code constants (`_PYCODESTYLE_CODES`, `_Z3_CODES`) to avoid hardcoding and improve maintainability
- Separated z3 tests into dedicated fixture and test function:
     - Added `z3_enabled` parameter to conditionally filter z3-related files based on whether z3 tests should run, guard module z3 import
     - Added new `test_z3_files_pyta` test and `pyta_z3_symbols` fixture.
     - Added `z3_enabled` and `pycodestyle_enabled` parameters to `_symbols_by_file_pyta()` to control which checkers run
- Changed from batch checking all files to per-file checking using enable/disable method
- Updated `pyta_pycodestyle_symbols()` fixture to use new parameterized `_symbols_by_file_pyta()` with `pycodestyle_enabled=True`

### Before

```shell
======================================================================
Comparing Test Methods in test_examples.py
Runs per test: 5
======================================================================

test_examples_files_pyta                     :  75.8615s (min: 63.8625s, max: 87.8605s)
test_pycodestyle_errors_pyta                 :   8.4406s (min: 8.3915s, max: 8.4897s)
test_c9104_module_name_violation             :   1.6716s (min: 1.6688s, max: 1.6744s)
test_cyclic_import                           :   1.3572s (min: 1.2409s, max: 1.4736s)

======================================================================
Test Methods Sorted by Average Time (slowest first):
======================================================================

1. test_examples_files_pyta                     : 75.8615s
2. test_pycodestyle_errors_pyta                 : 8.4406s
3. test_c9104_module_name_violation             : 1.6716s
4. test_cyclic_import                           : 1.3572s

======================================================================
```

### After

> I was actually hoping you could try modifying this to make one check_all call per file, and then for that specific file disable all the checks except the one based on the filename.

This approach was indeed way faster! The results speak for themselves, almost a 30s difference with the previous best time for `test_examples_files_pyta`.

```
======================================================================
Comparing Test Methods in test_examples.py
Runs per test: 10
======================================================================

test_examples_files_pyta                     :  30.9596s (min: 30.6273s, max: 32.2233s)
test_pycodestyle_errors_pyta                 :   9.0762s (min: 8.2886s, max: 12.3741s)
test_c9104_module_name_violation             :   1.6491s (min: 1.5433s, max: 1.7838s)
test_cyclic_import                           :   1.2155s (min: 1.1911s, max: 1.2551s)

======================================================================
Test Methods Sorted by Average Time (slowest first):
======================================================================

1. test_examples_files_pyta                     : 30.9596s
2. test_pycodestyle_errors_pyta                 : 9.0762s
3. test_c9104_module_name_violation             : 1.6491s
4. test_cyclic_import                           : 1.2155s

======================================================================
```

## Type of Change

## Checklist

## Questions and Comments

At this point, I believe that there isn’t much more that can be optimized. The only way to achieve another meaningful performance gain would be to optimize the `check_all` function itself.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,7 +18,6 @@ Z3_RELATED_TESTS = {
     r".*test_possibly_undefined_checker.*",
     r".*test_redundant_assignment_checker.*",
     r".*test_cfg_generator_z3.*",
-    r".*test_examples.*",
 }
 
 
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -60,9 +60,6 @@ def get_file_paths(paths: Union[str, list[str]], z3_enabled: bool = False) -> li
     if isinstance(paths, str):
         paths = [paths]
 
-    if z3_enabled:
-        import z3
-
     for path in paths:
         for root, _, files in os.walk(path, topdown=True):
             for filename in files:


david-yz-liu:
@Karl-Michaud the observation about Z3 is good. Regarding the `check_all` and `enable`/`disable` question, I'd like to see more information. Please push an example of the code you ran and the precise timing experiment you performed.

david-yz-liu:
Hi @Karl-Michaud, very nice update to your report. I appreciate you showing the code you used, as I think you may have misunderstood exactly what I was going for. Currently there's one main `check_all` call, and then some configuration passed in (which you used to disable/enable checks). I was actually hoping you could try modifying this to make _one `check_all` call per file_, and then for that specific file disable all the checks except the one based on the filename. Moreover, one additional thing you can do is add the `@only_required_for_messages` decorator to all of our custom checks, to make sure the disabling of checks really causes the visit functions to not be executed.

This would increase the number of `check_all` calls in total, of course, but hopefully reduce the total amount of checks. I'm not sure if this will result in a speedup (as you make a great point that each `check_all` call also has a cost), but this is the timing experiment I'd really like to see. 😄

david-yz-liu:
@Karl-Michaud for the last part, I think it's fine for you to add this in the same PR, as it will hopefully affect the timing results, and so is part of the "optimization" exploration.

python_ta/checkers/global_variables_checker.py
```diff
@@ -33,16 +33,19 @@ def __init__(self, linter=None) -> None:
...
         self.add_message("forbidden-global-variables", node=node, args=args)
 
+    @only_required_for_messages("forbidden-global-variables")
     def visit_assignname(self, node: nodes.AssignName) -> None:
         """Allow global constant variables (uppercase) and type aliases (type alias pattern), but issue messages for
         all other globals.
         """
         self._inspect_vars(node)
 
+    @only_required_for_messages("forbidden-global-variables")
```
david-yz-liu, line 48: This is good, but you can add the decorator to the other `visit_` methods below as well (since they aren't needed if we aren't checking for `forbidden-global-variables`)

tests/conftest.py
```diff
@@ -18,6 +18,7 @@
     r".*test_possibly_undefined_checker.*",
     r".*test_redundant_assignment_checker.*",
     r".*test_cfg_generator_z3.*",
+    r".*test_examples.*",
```
david-yz-liu: So some examples use Z3 but others don't. I think it's too restrictive to add all `test_examples` here.

Instead, you should just include the `test_z3_files_pyta` test that you created. However, the opposite is also true: when `z3` is enabled, we should be skipping `test_examples_files_pyta` and just running `test_z3_files_pyta` instead.

tests/test_examples.py
```diff
@@ -42,67 +45,86 @@
...
             for filename in files:
-                if (
-                    filename not in IGNORED_TESTS
-                    and filename.endswith(".py")
-                    and (z3_dependency_available or filename not in Z3_RELATED_TESTS)
-                ):
-                    full_path = os.path.join(root, filename)
-                    rel_path = os.path.relpath(full_path, path)
-                    test_files.append(os.path.join(path, rel_path))
+                if filename in IGNORED_TESTS or not filename.endswith(".py"):
```
david-yz-liu: The change in logic here isn't good. When `z3_enabled` is True, all example files should be checked, not just the ones in `Z3_RELATED_TESTS`. This is to ensure that enabling Z3 doesn't break existing checker functionality.

You can keep the code the same as before but swap out `z3_dependency_available` for `z3_enabled`.

tests/test_examples.py
```diff
@@ -42,67 +45,86 @@
...
     if isinstance(paths, str):
         paths = [paths]
 
-    # check if z3 dependency is available
-    z3_dependency_available = True
-    try:
-        import z3
-    except ImportError:
-        z3_dependency_available = False
+    if z3_enabled:
```
david-yz-liu: It's good to add a `z3_enabled` parameter explicitly, but in this case there shouldn't be a try-except block. When `z3_enabled` we can just run `import z3`, and if there's an error this should raised to the user, since the test isn't working as intended.

tests/test_examples.py
```diff
@@ -167,6 +201,25 @@ def test_examples_files_pyta(test_file: str, pyta_examples_symbols: dict[str, se
     ), f"Failed {test_file}. File does not add expected message {checker_name}: {file_symbols}."
 
 
+@pytest.mark.parametrize("test_file", get_file_paths([_CUSTOM_CHECKER_PATH], z3_enabled=True))
```
david-yz-liu: It's okay to keep `_CUSTOM_CHECKER_PATH` here though (since even with `Z3` enabled, this won't affect any of pylint's checkers

tests/test_examples.py
```diff
@@ -42,67 +45,86 @@
...
-    )
-
-    jsons_output = sys.stdout.getvalue()
-    sys.stdout = sys.__stdout__
-    pyta_list_output = json.loads(jsons_output)
+    files_to_check = get_file_paths(paths, z3_enabled=z3_enabled)
+    pyta_list_output = []
+    for file in files_to_check:
+        if pycodestyle_enabled:
+            enable_checks = _PYCODESTYLE_CODES + ["C9960"]
```
david-yz-liu: It isn't necessary to add `C9960` to these lists. It doesn't matter that C9960 is disabled by default; it should be enabled explicitly _only_ on the file `c9960_unmentioned_parameter.py`, which will be handled automatically by the code on line 115.

tests/test_examples.py
```diff
@@ -42,67 +45,86 @@
...
+        elif z3_enabled:
```
david-yz-liu: Similar to my previous Z3-related comment, this branch isn't necessary, nor is defining a separate `_Z3_CODES`. When the Z3 tests are running, we should still only be checking a single check (whatever is specified by the filename, determined by Line 115). `z3_enabled` can be used to set the `z3` config value, as you're currently doing, but should _not_ be determining which checks are run.

tests/test_examples.py
```diff
@@ -42,67 +45,86 @@
...
     included along with message symbols.
 
     :param paths: The paths to retrieve the files from.
     :param include_msg: whether to include message descriptions in the symbol set
+    :param z3_enabled: Option to enable z3 dependency.
+    :param pycodestyle_enabled: Option to enable pycodestyle-specific checks (E9989).
     :return: A dictionary mapping each file name to a set of PythonTA message symbols
     (and descriptions if include_msg is True).
     """
     sys.stdout = StringIO()
```
david-yz-liu: This line can be deleted, as you are using the `output` argument to `check_all`, which is great.

tests/test_examples.py
Karl-Michaud: Hi @david-yz-liu, just to confirm, did `_EXAMPLES_PATH`? `_CUSTOM_CHECKER_PATH` stays unchanged in both versions.

Karl-Michaud:
Hi @david-yz-liu ,.

Thanks for all the feedback.

Regarding Z3 test isolation, I chose to use the `@pytest.mark.skipif` ([documentation](https://docs.pytest.org/en/stable/how-to/skipping.html#skipif)) decorator (see the final paragraph in that section on string conditions). Since `pytest_ignore_collect` in `tests/conftest.py` filters based on file paths rather than individual test functions, the `@pytest.mark.skipif` decorator comes in handy by letting us selectively run only the tests we want.

tests/conftest.py
david-yz-liu: Using `@pytest.mark.skipif` below is good, but in that case I don't think we need to include `test_examples` here.

Karl-Michaud: I see, that makes sense. I assumed that including the file would trigger both `--exclude-z3` and no flag.

david-yz-liu: @Karl-Michaud I see you requested another review. When removing this line, there is an import error with `z3` when running the `pytest.mark.parametrize` for the z3 test. However, looking at the code more carefully, we don't need to import `z3` at all here. It's fine to let the `_get_file_paths` function be called even without `z3`; the resulting tests will be skipped, so PythonTA itself won't need to import z3.

david-yz-liu (approved):
Nice work, @Karl-Michaud. Looking at the full diff, it looks like the reason `z3` was imported earlier was to determine if we should enable the z3 checks. But you replaced this with an explicit parameter `z3_enabled`, so the "import test" was no longer necessary.

PR #1366, opened 2026-07-03
Add function tracing to SnapshotTracer

rachelzUT:
## Proposed Changes

Modified `SnapshotTracer` to trace into calls of functions defined in the same module that `SnapshotTracer` is called, as `SnapshotTracer` previously only traced lines of code within the context block and did not trace into any function calls within the block. Note that built-in functions and functions from external libraries/modules are not traced into. 

...

<details>

</details>

## Type of Change

| ✨ _New feature_ (non-breaking change that adds functionality)                          |     X  |

## Checklist

## Questions and Comments

- I modified the webstepper documentation `packages/python-ta/docs/debug/index.md` with the new feature. Before it was listed under "Current Limitations" which I just edited to reflect the change, but if it's no longer considered a 'limitation' since this fixes that limitation, I can move the description elsewhere.
- The same change to the MemoryViz Webstepper code (handled by the MemoryViz team) is required for this PR as well as for #1360, specifically implementing a `startLineNumber` which will ensure the correct line numbers and highlighting are shown in the code section of the Webstepper

### DIFF
--- a/packages/python-ta/docs/debug/index.md
+++ b/packages/python-ta/docs/debug/index.md
@@ -489,7 +489,7 @@ if __name__ == '__main__':
     func_multi_line()
 ```
 
-When this function runs, MemoryViz snapshots are captured and stored internally in the `SnapshotTracer`. These snapshots can be accessed via the `snapshots` property. If `webstepper=True`, the snapshots are rendered into an HTML visualization. For the expected output, refer to the snapshots in `tests/test_debug/snapshot_tracer_testing_snapshots/func_multi_line`.
+When this function runs, MemoryViz snapshots are captured and stored internally in the `SnapshotTracer`. These snapshots can be accessed via the `snapshots` property. If `webstepper=True`, the snapshots are rendered into an HTML visualization. The context manager will step into Python functions defined in the same module as the traced context; calls into external modules and built-ins are not traced. For the expected output, refer to the snapshots in `tests/test_debug/snapshot_tracer_testing_snapshots/func_multi_line`.
 
 ### API
 
@@ -504,5 +504,4 @@ When this function runs, MemoryViz snapshots are captured and stored internally 
 The `SnapshotTracer` has the following limitations:
 
 1. Due to differences in Python interpreters, this context manager only works with Python versions >= 3.10.
-2. The context manager only steps into Python functions defined in the same module as the traced context. Calls into external modules and builtins are not traced.
-3. `SnapshotTracer` uses [`sys.settrace`] to update variable states, and therefore is not compatible with other libraries (e.g., debuggers, code coverage tools).
+2. `SnapshotTracer` uses [`sys.settrace`] to update variable states, and therefore is not compatible with other libraries (e.g., debuggers, code coverage tools).
--- a/packages/python-ta/docs/debug/index.md
+++ b/packages/python-ta/docs/debug/index.md
@@ -486,5 +486,5 @@ When this function runs, MemoryViz snapshots are captured and stored internally 
 The `SnapshotTracer` has the following limitations:
 
 1. Due to differences in Python interpreters, this context manager only works with Python versions >= 3.10.
-2. The context manager does not step into any function calls. Calling functions within the traced function may lead to undefined behavior.
+2. The context manager only steps into Python functions defined in the same module as the traced context. Calls into external modules and builtins are not traced.
 3. `SnapshotTracer` uses [`sys.settrace`] to update variable states, and therefore is not compatible with other libraries (e.g., debuggers, code coverage tools).
--- a/packages/python-ta/src/python_ta/debug/snapshot_tracer.py
+++ b/packages/python-ta/src/python_ta/debug/snapshot_tracer.py
@@ -73,14 +73,17 @@ class SnapshotTracer:
         self._origin_file = None
         self._module_source_lines = None
 
-    def _trace_func(self, frame: types.FrameType, event: str, _arg: Any) -> Any:
-        """Take a snapshot of the variables in the functions specified in `self.include`, tracing into same-module function calls."""
+    def _global_trace_func(self, frame: types.FrameType, event: str, _arg: Any) -> Any:
+        """Global trace function that handles 'call' events to determine which functions to trace into."""
         if event == "call":
             if self._origin_file is None:
                 return None
             # Only trace functions in the same module as the calling function.
             called_file = os.path.normcase(os.path.abspath(frame.f_code.co_filename))
-            if called_file == self._origin_file and frame.f_code.co_name != "_trace_func":
+            if called_file == self._origin_file and frame.f_code.co_name not in (
+                "_trace_func",
+                "_global_trace_func",
+            ):
                 self._start_lineno = min(self._start_lineno, frame.f_code.co_firstlineno)
                 self._end_lineno = max(
                     self._end_lineno,
@@ -90,6 +93,8 @@ class SnapshotTracer:
                 return self._trace_func
             return None
 
+    def _trace_func(self, frame: types.FrameType, event: str, _arg: Any) -> Any:
+        """Local trace function set on each frame to take a snapshot of the variables in the functions specified in `self.include`."""
         if event == "line":
             self._start_lineno = min(self._start_lineno, frame.f_lineno)
             self._end_lineno = max(self._end_lineno, frame.f_lineno)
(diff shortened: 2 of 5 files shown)


packages/python-ta/src/python_ta/debug/snapshot_tracer.py
```diff
@@ -63,9 +67,24 @@ def __init__(
...
+            # Only trace functions in the same module as the calling function.
+            called_file = os.path.normcase(os.path.abspath(frame.f_globals.get("__file__", "")))
+            if called_file == self._origin_file and frame.f_code.co_name != "_trace_func":
+                self._traced_frames.append(
+                    {
+                        "co_firstlineno": frame.f_code.co_firstlineno,
+                        "source_lines": inspect.getsource(frame.f_code).splitlines(),
+                    }
+                )
+                return self._trace_func
```
david-yz-liu: Add a comment explaining the return value here

packages/python-ta/src/python_ta/debug/snapshot_tracer.py
```diff
@@ -85,7 +104,10 @@ def __enter__(self):
         """Set up the trace function to take snapshots at each line of code."""
         func_frame = inspect.getouterframes(inspect.currentframe())[1].frame
         func_frame.f_trace = self._trace_func
-        sys.settrace(lambda *_args: None)
+        self._origin_file = os.path.normcase(
+            os.path.abspath(func_frame.f_globals.get("__file__", ""))
+        )
+        sys.settrace(self._trace_func)
```
david-yz-liu: I'm not sure why this is necessary; is it not sufficient to have `_trace_func` return itself when a new function is called?

packages/python-ta/src/python_ta/debug/snapshot_tracer.py
```diff
@@ -85,7 +104,10 @@ def __enter__(self):
         """Set up the trace function to take snapshots at each line of code."""
         func_frame = inspect.getouterframes(inspect.currentframe())[1].frame
         func_frame.f_trace = self._trace_func
-        sys.settrace(lambda *_args: None)
+        self._origin_file = os.path.normcase(
+            os.path.abspath(func_frame.f_globals.get("__file__", ""))
```
david-yz-liu: `""` is not a good default. If the file cannot be found, let's be conservative and _do not trace into any function calls_.

packages/python-ta/src/python_ta/debug/snapshot_tracer.py
```diff
@@ -63,9 +67,24 @@ def __init__(
...
+    def _trace_func(self, frame: types.FrameType, event: str, _arg: Any) -> Any:
+        """Take a snapshot of the variables in the functions specified in `self.include`, tracing into same-module function calls."""
+        if event == "call":
+            # Only trace functions in the same module as the calling function.
+            called_file = os.path.normcase(os.path.abspath(frame.f_globals.get("__file__", "")))
+            if called_file == self._origin_file and frame.f_code.co_name != "_trace_func":
+                self._traced_frames.append(
+                    {
+                        "co_firstlineno": frame.f_code.co_firstlineno,
+                        "source_lines": inspect.getsource(frame.f_code).splitlines(),
```
david-yz-liu: Overall this approach is storing more than it needs.

The existing `_first_line` attribute can be reused but renamed to `_start_lineno`; its role should be changed to store the minimum line number that should be displayed. Similarly, you can add an `_end_lineno` attribute to store the maximum line number that should be displayed. Both can be updated directly here, instead of introducing a new `self._trace_frames` attribute.

In `_get_code` below, right now `_first_line` is being used to index into just the source lines of the original function frame. We can greatly simplify by (1) getting the full lines of the _module_ rather than just the single function, and then (2) directly indexing into that list using `_start_lineno` and `_end_lineno`, without doing any offsetting.

packages/python-ta/src/python_ta/debug/webstepper/webstepper_template.html.jinja
```diff
@@ -10,6 +10,7 @@
     <script>
       window.codeText = `{{ code_text }}`;
       window.memoryVizData = {{ memory_viz_data | tojson }};
+      window.startLineNumber = {{ start_line_number }};
```
david-yz-liu: Do not make changes to the webstepper in this PR. This PR should deal entirely with the logical changes necessary to do the tracing into function calls. This is why I said:

> To avoid the issue you ran into in the earlier webstepper PR, you could try testing this with only functions that are defined below the `SnapshotTracer` context block.

packages/python-ta/src/python_ta/debug/snapshot_tracer.py
```diff
@@ -63,9 +67,24 @@ def __init__(
...
         self.webstepper = webstepper
         self._first_line = float("inf")
+        self._origin_file = None
+        self._traced_frames = []
+
+    def _trace_func(self, frame: types.FrameType, event: str, _arg: Any) -> Any:
+        """Take a snapshot of the variables in the functions specified in `self.include`, tracing into same-module function calls."""
+        if event == "call":
+            # Only trace functions in the same module as the calling function.
+            called_file = os.path.normcase(os.path.abspath(frame.f_globals.get("__file__", "")))
```
david-yz-liu: Similar to my comment below, if no `__file__` attribute is found, do not trace into the call.

packages/python-ta/src/python_ta/debug/snapshot_tracer.py
rachelzUT: We still need to set `self._trace_func` as the argument to register it as the global trace function so that on any “call” events in the future (when a function is called inside the `with` block), Python uses the return value of `self._trace_func` as the local trace for the called function. If we only have `_trace_func` return itself, tracing wouldn't actually be activated for the called functions.

packages/python-ta/docs/debug/index.md
```diff
@@ -504,5 +504,5 @@ When this function runs, MemoryViz snapshots are captured and stored internally
 The `SnapshotTracer` has the following limitations:
 
 1. Due to differences in Python interpreters, this context manager only works with Python versions >= 3.10.
-2. The context manager does not step into any function calls. Calling functions within the traced function may lead to undefined behavior.
+2. The context manager only steps into Python functions defined in the same module as the traced context. Calls into external modules and builtins are not traced.
```
david-yz-liu: This wording change is good, but I think we should move this out of the "limitations" section and instead have this be a general description of the function tracing behaviour for `SnapshotTracer`.

packages/python-ta/src/python_ta/debug/snapshot_tracer.py
david-yz-liu: Okay yes, I recall you also explained this in your presentation last week.

In this case it seems to me like we should leave `_trace_func` unchanged and instead define a separate global tracing function that only handles function calls. This will help keep the code more modular and make clear the two separate purposes.

packages/python-ta/src/python_ta/debug/snapshot_tracer.py
```diff
@@ -62,13 +68,36 @@ def __init__(
...
-    def _trace_func(self, frame: types.FrameType, event: str, _arg: Any) -> None:
-        """Take a snapshot of the variables in the functions specified in `self.include`"""
-        if self._first_line == float("inf"):
-            self._first_line = frame.f_lineno
+        self._start_lineno = sys.maxsize
+        self._end_lineno = 0
+        self._origin_file = None
+        self._module_source_lines = None
+
+    def _global_trace_func(self, frame: types.FrameType, event: str, _arg: Any) -> Any:
```
david-yz-liu: Oh I think I missed this last time, but the return type here can be simplified (it's either a `Callable` or `None`)

packages/python-ta/src/python_ta/debug/snapshot_tracer.py
```diff
@@ -62,13 +68,36 @@ def __init__(
...
+        """Global trace function that handles 'call' events to determine which functions to trace into."""
+        if event == "call":
```
david-yz-liu: By splitting this off into a separate helper you can simplify the logic a bit. I would have one guard at the very start of the function for an early return (`if event != "call" or self._origin_file is None`).

packages/python-ta/src/python_ta/debug/snapshot_tracer.py
```diff
@@ -62,13 +68,36 @@ def __init__(
...
+            if self._origin_file is None:
+                return None
+            # Only trace functions in the same module as the calling function.
+            called_file = os.path.normcase(os.path.abspath(frame.f_code.co_filename))
+            if called_file == self._origin_file and frame.f_code.co_name not in (
+                "_trace_func",
```
david-yz-liu: Oh I missed this too, but please use `self._trace_func.__name__`, and similarly for the other one. This is better than string constants, as it'll be detected as a usage of the methods, which is good if we ever rename them.

david-yz-liu (approved):
Nice work, @rachelzUT!

PR #1369, opened 2026-07-26
Allow PyTA CLI to read file content from stdin

rachelzUT:
## Proposed Changes

Modified PyTA CLI to optionally read file contents through standard input. This is triggered by either passing the `--stdin` command-line argument or passing `-` as the filenames argument. 

...

<details>

</details>

## Type of Change

| ✨ _New feature_ (non-breaking change that adds functionality)                          |     X    |

## Checklist


packages/python-ta/src/python_ta/__main__.py
```diff
@@ -65,26 +74,49 @@ def main(
             sys.exit(0)
 
     checker = check_errors if errors_only else check_all
-    paths = [click.format_filename(fn) for fn in filenames]
+    use_stdin = stdin or (len(filenames) == 1 and filenames[0] == "-")
+
+    if use_stdin:
+        stdin_contents = sys.stdin.read()
```
david-yz-liu: Instead of manually calling `read`/`write`, let's use [`shutil.copyfileobj`](https://docs.python.org/3/library/shutil.html#shutil.copyfileobj). Note the comment in the documentation about calling `flush` on the file.

packages/python-ta/src/python_ta/__main__.py
```diff
@@ -65,26 +74,49 @@ def main(
...
+        ) as temp_file:
+            temp_file.write(stdin_contents)
+            temp_file_path = temp_file.name
 
+        try:
+            paths = [temp_file_path]
+            reporter = _invoke_checker(checker, paths, config, output_format)
+        finally:
+            # Clean up the temporary file
+            path.os.unlink(temp_file_path)
```
david-yz-liu: Instead of doing this, you should be able to call `_invoke_checker` within the `with tempfile.NamedTemporaryFile` block.

packages/python-ta/src/python_ta/__main__.py
```diff
@@ -65,26 +74,49 @@ def main(
...
 
     checker = check_errors if errors_only else check_all
-    paths = [click.format_filename(fn) for fn in filenames]
+    use_stdin = stdin or (len(filenames) == 1 and filenames[0] == "-")
+
+    if use_stdin:
+        stdin_contents = sys.stdin.read()
+        # Write the contents to a temporary file
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".py", delete=False, encoding="utf-8"
```
david-yz-liu: I assume here there's a random filename generated; does this appear to the user in the report? If so we might want a name prefix like `stdin`.

david-yz-liu (commented):
@rachelzUT nice work; I left a few comments, but please also update the docs CLI page with this change.

packages/python-ta/src/python_ta/__main__.py
```diff
@@ -65,26 +75,44 @@ def main(
...
+    if use_stdin:
+        # Write the contents of stdin to a temporary file
+        with tempfile.NamedTemporaryFile(
+            mode="w", prefix="stdin_", suffix=".py", delete=False, encoding="utf-8"
+        ) as temp_file:
+            shutil.copyfileobj(sys.stdin, temp_file)
+            temp_file.flush()
+            reporter = _invoke_checker(checker, [temp_file.name], config, output_format)
+        # Clean up the temporary file
+        path.os.unlink(temp_file.name)
```
david-yz-liu, line 92: We should be able to remove the `delete=False` from `NamedTemporaryFile` so that we don't need to delete manually. (I assume the reason for the deletion earlier was to keep the file around for `_invoke_checker`, but now that's been moved into the block.)

rachelzUT: @david-yz-liu I kept the manual deletion because removing` delete=False` (which would default to `delete=True`) would cause issues on Windows devices. Specifically, opening the temp file from another process would cause a `PermissionError`:
https://docs.python.org/3/library/tempfile.html#:~:text=Opening%20the%20temporary,a%20PermissionError.

david-yz-liu: Hi @rachelzUT, thanks for linking to the docs. You should be able to use `delete_on_close=False` then to prevent this behaviour.

rachelzUT: @david-yz-liu The `delete_on_close` parameter appears to only be compatible with Python version 3.12+. Would that be an issue? I can also add this detail to the documentation if necessary.

david-yz-liu: Ah that's a great point. We continue to support Python 3.11 until its end of life, so for now please keep your current approach but add a TODO comment saying to switch to `delete_on_close=True` after we drop support for Python 3.11.

david-yz-liu (approved):
Nice work, @rachelzUT!

